PR #1, opened 2020-10-08
Creating Datamodel Tree

PaulHuwe:
Issue RCAL-33

### DIFF
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -1,0 +1,109 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/roman_datamodel/core.schema"
+type: object
+properties:
+  meta:
+    type: object
+    properties:
+      filename:
+        title: Name of the file
+        type: string
+      model_type:
+        title: Type of data model
+        type: string
+      telescope:
+        title: Telescope used to acquire the data
+        type: string
+      exposure:
+        title: Exposure parameters
+        type: object
+        properties:
+          type:
+            title: Type of data in the exposure
+            type: string
+            enum:
+              # Wide Field Imager
+             [WFI_CORON, WFI_DARK, WFI_FLAT, WFI_FOCUS,  WFI_IMAGE, WFI_WFSS,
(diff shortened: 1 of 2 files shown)


dmggh:
I'm not sure what I should be seeing here - a directory with 2 files of size 0 ?

PaulHuwe:
Yeah, that's all it is.. Nadia asked me to put the structure in. If doing this alone is too simple, I can undo the PR and wait until I have files completed to do it.

dmggh:
This may be a minor point, but in jwst there are :


jwst/datamodels/schemas/core.schema.yaml

jwst/datamodels/schemas/referencefile.schema.yaml


There is no:

jwst/datamodels/schemas/reference_files


Was there a decision to have all of the reference file schemas in their own directory ? I think it's probably a good idea.

________________________________

nden:
@dmggh Yes, there was a decision to split the schemas for reference files in a separate directory. Also for jwst.
@PaulHuwe If you have the files ready add them to this PR. Add them even if they are not complete so we can look at them and comment.

ddavis-stsci:
Here are the proposed initial metadata from INS for the flat ref file,
'Detector'; 'WFI01'
'Instrument':  'WFI'
'Description': 'Flat reference file'
'Author': 'Detector Lab'
'Date': ' '
'Useafter': ' '
'Origin':  'STScI'
'Filetype': 'ASDF'
'Reftype': 'Flat'
'Telescope': 'Roman'
'Pedigree': ''
'History': 'A novel is needed'
and these should be in the initial three schemas.

PaulHuwe:
Initial core schema file with Nadia's suggestion's have been added. I will address Dave's comments in the references file.

PaulHuwe:
I did not include filetype, as we are only working with ASDF now. I can put that back in, if desired.

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -0,0 +1,109 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
```
nden, line 3: This line is not necessary any more.

nden (commented):
Looks good to me. Only need to remove 
`$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"`

nden:
I suggest we merge this now and include improvements later. This will allow periwinkle to use the datamodels subpackage for further development.

ddavis-stsci (approved):

ddavis-stsci:
Merging PR

PR #2, opened 2020-10-13
Create asdf extension for datamodels schemas

eslavich:
This PR adds an extension that registers Roman datamodels schema URIs with asdf.  The test doesn't currently pass, but will when @PaulHuwe's update to the referencefile.schema.yaml `id` field is merged.


PaulHuwe (approved):
Looks okay to me.

PR #3, opened 2020-10-14
Create RomanDataModel and open function

eslavich:
This PR adds the `RomanDataModel` base class and `romancal.datamodels.open`.  The logic in `open` that selects the model class is just stubbed out for now, since we haven't settled on how that's going to work (and anyway, we haven't implemented any real models yet).

### DIFF
--- a/romancal/datamodels/tests/test_models.py
+++ b/romancal/datamodels/tests/test_models.py
@@ -1,0 +1,15 @@
+import asdf
+
+from romancal import datamodels
+
+
+def test_model_schemas():
+    model_classes = [
+        m for m in datamodels.__dict__.values()
+        if isinstance(m, type) and issubclass(m, datamodels.RomanDataModel)
+    ]
+    for model_class in model_classes:
+        try:
+            asdf.schema.load_schema(model_class.schema_url)
+        except Exception:
+            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model class {model_class.__name__}"
--- a/romancal/datamodels/tests/test_models.py
+++ b/romancal/datamodels/tests/test_models.py
@@ -1,14 +1,15 @@
 import asdf
 
-from romancal import datamodels
+from romancal.datamodels import RomanDataModel
 
 
 def test_model_schemas():
-    model_classes = [
-        m for m in datamodels.__dict__.values()
-        if isinstance(m, type) and issubclass(m, datamodels.RomanDataModel)
-    ]
-    for model_class in model_classes:
+    def iter_subclasses(model_class):
+        yield model_class
+        for sub_class in model_class.__subclasses__():
+            yield from iter_subclasses(sub_class)
+
+    for model_class in iter_subclasses(RomanDataModel):
         try:
             asdf.schema.load_schema(model_class.schema_url)
         except Exception:
--- a/romancal/datamodels/__init__.py
+++ b/romancal/datamodels/__init__.py
@@ -1,0 +1,5 @@
+from .core import RomanDataModel
+from .open_impl import open
+
+
+__all__ = ["RomanDataModel", "open"]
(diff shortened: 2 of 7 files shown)


PaulHuwe (approved):
Looks okay to me.

nden (approved):
Looks good.

romancal/datamodels/tests/test_models.py
```diff
@@ -0,0 +1,15 @@
+import asdf
+
+from romancal import datamodels
+
+
+def test_model_schemas():
+    model_classes = [
+        m for m in datamodels.__dict__.values()
+        if isinstance(m, type) and issubclass(m, datamodels.RomanDataModel)
+    ]
```
jdavies-st, line 10: Another way to do this would be to use the `RomanDataModel.__subclasses__()` method, though this method only returns immediate subclasses, not subclasses of subclasses.  But hopefully there are none of those.  😅

An example in `jwst`:

```
In [1]: from jwst import datamodels

In [2]: datamodels.DataModel.__subclasses__()
Out[2]: 
[jwst.datamodels.reference.ReferenceFileModel,
 jwst.datamodels.amilg.AmiLgModel,
 jwst.datamodels.asn.AsnModel,
 jwst.datamodels.combinedspec.CombinedSpecModel,
 jwst.datamodels.container.ModelContainer,
 jwst.datamodels.contrast.ContrastModel,
 jwst.datamodels.cube.CubeModel,
 jwst.datamodels.image.ImageModel,
 jwst.datamodels.extract1dimage.Extract1dImageModel,
 jwst.datamodels.gls_rampfit.GLS_RampFitModel,
 jwst.datamodels.guider.GuiderRawModel,
 jwst.datamodels.guider.GuiderCalModel,
 jwst.datamodels.ifucube.IFUCubeModel,
 jwst.datamodels.ifuimage.IFUImageModel,
 jwst.datamodels.irs2.IRS2Model,
 jwst.datamodels.level1b.Level1bModel,
 jwst.datamodels.ramp.RampModel,
 jwst.datamodels.slit.SlitDataModel,
 jwst.datamodels.slit.SlitModel,
 jwst.datamodels.multiexposure.MultiExposureModel,
 jwst.datamodels.multislit.MultiSlitModel,
 jwst.datamodels.spec.SpecModel,
 jwst.datamodels.multispec.MultiSpecModel,
 jwst.datamodels.quad.QuadModel,
 jwst.datamodels.rampfitoutput.RampFitOutputModel,
 jwst.datamodels.steppars.StepParsModel,
 jwst.datamodels.trapsfilled.TrapsFilledModel]
```

jdavies-st (approved):
Looks good.  Just some comments below.

romancal/datamodels/tests/test_models.py
eslavich: Ooo I like that better.  We can recurse into the subclasses and look for more.

nden:
If `__subclases__` is used would this get the reference file models? For example, in JWST we have something like (by memory):

```
class ReferenceFileModel(DataModel)

class FlatModel(ReferenceFileModel)
```

eslavich:
> If `__subclases__` is used would this get the reference file models? For example, in JWST we have something like (by memory):

I implemented a method like this:

```python
    def iter_subclasses(model_class):
        yield model_class
        for sub_class in model_class.__subclasses__():
            yield from iter_subclasses(sub_class)
```

So it yields subclasses and then recurses into their subclasses until the hierarchy is exhausted.

jdavies-st:
On the `jwst` side, using `__subclasses__()` finds `_SimpleModel` which is an intermediate subclass between `ReferenceFileModel` and some distortion models.  We may want to avoid that here in `romancal` and instead use a mixin for some special methods or attributes that are shared across models, but where the class itself is not useable as a datamodel.

PR #4, opened 2020-10-16
Datamodelinit

PaulHuwe:
Adding flat file schema.
Updated core & referencefile schemas with requested changes and tweaks to match sample data provided by @tddesjardins .

### DIFF
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -1,6 +1,5 @@
 %YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
 id: "http://stsci.edu/schemas/roman_datamodel/core.schema"
 type: object
 properties:
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -77,32 +77,10 @@ properties:
             enum: [WFI01, WFI02, WFI03, WFI04, WFI05, WFI06, WFI07, WFI08, WFI09,
                    WFI10, WFI11, WFI12, WFI13, WFI14, WFI15, WFI16, WFI17, WFI18,
                    ANY, N/A]
-# Placeholder for filters/etc.
-#          filter:
-#            title: Name of the filter element used
-#            type: string
-#            enum: [ANY, CLEAR, F070LP, F070W, F090W, F1000W, F100LP, F1065C,
-#              F110W, F1130W, F1140C, F115W, F1280W, F140M, F140X, F1500W, F150W,
-#              F150W2, F1550C, F170LP, F1800W, F182M, F187N, F200W, F2100W, F210M,
-#              F212N, F2300C, F250M, F2550W, F2550WR, F277W, F290LP, F300M,
-#              F322W2, F335M, F356W, F360M, F380M, F410M, F430M, F444W,
-#              F460M, F480M, F560W, F770W, FLENS, FND, GR150C,
-#              GR150R, OPAQUE, P750L, WLP4, N/A, MULTIPLE]
-#            fits_keyword: FILTER
-#          pupil:
-#            title: Name of the pupil element used
-#            type: string
-#            enum: [ANY, CLEAR, CLEARP, F090W, F115W, F140M, F150W, F158M, F162M, F164N,
-#              F200W, F323N, F405N, F466N, F470N, FLAT, GDHS0, GDHS60, GR700XD,
-#              GRISMC, GRISMR, GRISMV2, GRISMV3, MASKBAR, MASKIPR, MASKRND, NRM, PINHOLES,
-#              WLM8, WLP8, N/A]
-#          pupil_imaging_lens:
-#            title: Pupil imaging lens in the optical path?
-#            type: boolean
-#            fits_keyword: PILIN
-#          grating:
-#            title: Name of the grating element used
-#            type: string
-#            enum: [G140M, G235M, G395M, G140H, G235H, G395H, PRISM, MIRROR, N/A, ANY, UNKNOWN, MULTIPLE]
-#            fits_keyword: GRATING
+          filter:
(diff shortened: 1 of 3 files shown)


romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -1,6 +1,5 @@
 %YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
```
eslavich, line 3: Opinions may vary, but I don't think we should remove this entirely.  We don't need the FITS validators defined in the jwst metaschema but I think we should still specify _some_ metaschema.  Since we're not using the ndarray validators here the base YAML metaschema is appropriate:

```yaml
schema: http://stsci.edu/schemas/yaml-schema/draft-01
```

or we could just use the ASDF metaschema everywhere so we don't have to think about whether or not the array features are necessary:

```yaml
$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
```

Incidentally, the quotes around URI values aren't necessary.   There aren't any characters in there that YAML parsers find confusing.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
```diff
@@ -0,0 +1,25 @@
...
+---
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/flat.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: ../core.schema
+- type: object
+  properties:
+    data:
+      title: The science data
+      fits_hdu: SCI
```
eslavich, line 11: Can the `fits_hdu` property here and in the other arrays can be removed?  That's a pseudo-validator that defines a mapping between an ndarray model attribute and a FITS HDU.  I didn't think Roman was using that feature of DataModel.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
```diff
@@ -0,0 +1,25 @@
+%YAML 1.1
+---
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/flat.schema"
```
eslavich, line 3: This schema uses ndarray validators like `ndim` and `datatype` so it needs the ASDF metaschema:

```yaml
$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
```

romancal/datamodels/schemas/reference_files/flat.schema.yaml
```diff
@@ -0,0 +1,25 @@
...
+allOf:
+- $ref: referencefile.schema
+- $ref: ../core.schema
```
eslavich, line 6: Another option is to `$ref` in the core schema from within referencefile.schema.  Advantage is not having to remember to include core in the individual schemas for each reference type.  Disadvantage is a deeper hierarchy of schemas, which can be harder to think about.  It's a matter of taste, but I thought I'd better point out that the option exists.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
```diff
@@ -1,7 +1,6 @@
 %YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
```
eslavich, line 3: Like core.schema, this file can use either of the YAML or ASDF metaschemas.

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -1,6 +1,5 @@
 %YAML 1.1
```
eslavich, line 1: We'll need to get the schema tester running on these.  We need to get the tests running on Travis, too.  I'll file an issue for both of those tasks.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/core.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/core.schema.yaml
PaulHuwe: Done.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
```diff
@@ -1,33 +1,35 @@
...
+        pedigree:
+          title: The pedigree of the reference file
+          type: string
+        description:
+          title: Description of the reference file
+          type: string
+        author:
+          title: Author of the reference file
+          type: string
+        history:
```
eslavich, line 30: I don't know what this attribute is for, but there is a history feature built into ASDF that you might consider using.  DataModel has some special handling around it, see:

https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/model_base.py#L997

eslavich (approved):
One more suggestion to consider, but looks good either way.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
PaulHuwe: Done.

eslavich (approved):

PR #33, opened 2020-11-03
Initial Reference File Testing

PaulHuwe:
Added tests for core, referencefile, and flat schemas. Updated the schemas for required keywords and modified time types. Some tests are commented out due to known asdf errors. (RCAL-25, RCAL-26, RCAL-50, RCAL-52)

### DIFF
--- a/romancal/datamodels/open_impl.py
+++ b/romancal/datamodels/open_impl.py
@@ -6,6 +6,7 @@ from pathlib import PurePath, Path
 import asdf
 
 from .core import RomanDataModel
+from .referencefile import ReferenceFileModel
 from .flat import FlatModel
 
 
@@ -86,8 +87,14 @@ def _select_model_class(asdf_file):
     # - Use RomanDataModel for all data but select the schema according to one of the above
     # - ???
 
-    # If we keep it this route, need to check for the existence of reftype first
-    if asdf_file["meta"].get("reftype") == "FLAT":
-        return FlatModel
+    # Check for the existence of reftype to indicate a reference file model
+    if asdf_file["meta"].get("reftype", None):
+        # Check if flat file model
+        if asdf_file["meta"].get("reftype", None) == "FLAT":
+            return FlatModel
+        # Return base reference file model
+        else:
+            return ReferenceFileModel
+    # Not a reference file model
     else:
         return RomanDataModel
--- a/romancal/datamodels/open_impl.py
+++ b/romancal/datamodels/open_impl.py
@@ -88,13 +88,15 @@ def _select_model_class(asdf_file):
     # - ???
 
     # Check for the existence of reftype to indicate a reference file model
-    if asdf_file["meta"].get("reftype", None):
+    reftype = asdf_file["meta"].get("reftype")
+    if reftype is not None:
         # Check if flat file model
-        if asdf_file["meta"].get("reftype", None) == "FLAT":
+        if reftype == "FLAT":
             return FlatModel
         # Return base reference file model
         else:
             return ReferenceFileModel
+
     # Not a reference file model
     else:
         return RomanDataModel
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -16,7 +16,7 @@ properties:
       telescope:
         title: Telescope used to acquire the data
         type: string
-        enum: ["ROMAN"]
+        enum: [ROMAN]
       exposure:
         title: Exposure parameters
         type: object
@@ -84,6 +84,5 @@ properties:
             type: string
             enum: [F062, F087, F106, F129, W146, F158, F184, GRISM, PRISM, CLEAR,
                    N/A, ANY, UNKNOWN]
-        required: [detector,optical_element,name]
-    required: [instrument,telescope]
-additionalProperties: true
+        required: [detector, optical_element, name]
+    required: [instrument, telescope]
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -79,10 +79,11 @@ properties:
             enum: [WFI01, WFI02, WFI03, WFI04, WFI05, WFI06, WFI07, WFI08, WFI09,
                    WFI10, WFI11, WFI12, WFI13, WFI14, WFI15, WFI16, WFI17, WFI18,
                    ANY, N/A]
-          filter:
+          optical_element:
             title: Name of the filter element used
             type: string
             enum: [F062, F087, F106, F129, W146, F158, F184, GRISM, PRISM, CLEAR,
                    N/A, ANY, UNKNOWN]
-
-
+        required: [detector,optical_element,name]
+    required: [instrument,telescope]
+additionalProperties: true
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -16,7 +16,7 @@ properties:
       telescope:
         title: Telescope used to acquire the data
         type: string
-        enum: [Roman]
+        enum: ["ROMAN"]
       exposure:
         title: Exposure parameters
         type: object
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -10,8 +10,16 @@ properties:
       filename:
         title: Name of the file
         type: string
+# NOTE: Commented out due to conflict with meta.date defined in stdatamodels
+# Ticket: https://github.com/spacetelescope/stdatamodels/issues/23
+#        date:
+#          title: Date this file was created (UTC)
+#          tag: tag:stsci.edu:asdf/time/time-1.1.0
       model_type:
         title: Type of data model
+        type: string
+      origin:
+        title: Organization responsible for creating file
         type: string
       telescope:
         title: Telescope used to acquire the data
@@ -26,27 +34,13 @@ properties:
             type: string
             enum:
               # Wide Field Imager
-             [WFI_CORON, WFI_DARK, WFI_FLAT, WFI_FOCUS,  WFI_IMAGE, WFI_WFSS,
-              WFI_LED, WFI_WFSC, WFI_TACONFIRM, WFI_TACQ, WFI_TSGRISM, WFI_TSIMAGE,
-              # WFI Spectrograph
-              WFI_PRISM, WFI_GRISM,
-              # Coronagraph
-              CGI_IMAGE,
-              # Misc
-              N/A, ANY]
+             [WFI_IMAGE, WFI_GRISM, WFI_PRISM, WFI_DARK, WFI_FLAT, WFI_TACQ, WFI_WFSC, WFI_GW]
           start_time:
             title: UTC exposure start time
             type: number
           end_time:
             title: UTC exposure end time
             type: number
-          readpatt:
-            title: Readout pattern
-            type: string
-            enum: [RAPID, ANY, N/A]
-          nints:
-            title: Number of integrations in exposure
-            type: integer
           ngroups:
             title: Number of groups in integration
             type: integer
@@ -58,9 +52,6 @@ properties:
             type: integer
           frame_time:
             title: "[s] Time between frames"
-            type: number
-          group_time:
-            title: "[s] Time between groups"
             type: number
           exposure_time:
             title: "[s] Effective exposure time"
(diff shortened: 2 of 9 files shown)


romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
...
+                }
+    }
+
+    # Set temporary asdf file
+    file_path = tmp_path / "test.asdf"
+
+    # Cycle through reference file models to test each
+    for model_class in iter_subclasses(DataModels.referencefile.ReferenceFileModel):
+        # Test base referencefile model
+        if (model_class == DataModels.referencefile.ReferenceFileModel):
```
eslavich, line 125: I think this `if` statement is a sign that we need separate tests for `ReferenceFileModel` and actual reference models like `FlatModel`.  It's also nice to parametrize the test so that we can easily see what models are failing, otherwise the test will just halt at the first failure.  I'd recommend we set up like this:

```python
def test_reference_file_model_base(tmp_path):
  # ReferenceFileModel tests


@pytest.mark.parametrize("model_class", list(iter_subclasses(ReferenceFileModel)))
def test_reference_file_model(tmp_path, model_class):
  # Common tests for all ReferenceFileModel subclasses


def test_flat_model(tmp_path):
  # FlatModel tests
```

We'd need to add a flag to iter_subclasses that prevents it from yielding the original parent class.

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
...
+            # Test required key/values pairs in meta dictionary
+            confirm_required_keywords(tmp_path, meta)
+
+            # Testing referencefile asdf file
+            with asdf.AsdfFile(meta) as af:
+                af.write_to(file_path)
+
+                # Load base referencefile schema
+                schema = asdf.schema.load_schema(
+                    "http://stsci.edu/schemas/roman_datamodel/reference_files/referencefile.schema")
```
eslavich, line 135: You could grab this out of `ReferenceFileModel.schema_url` to save from having to specify the string in multiple places.

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
...
+
+                # Ensure that base referencefile schema contains core schema
+                found = False
+                for schemaref in schema.get("allOf", [{}]):
+                    if schemaref.get("$ref","") == "../core.schema":
```
eslavich, line 140: Hmm... this seems like a good candidate for a assertion helper:

```python
def assert_referenced_schema(schema_uri, ref_uri):
  __tracebackhide__ = True
  schema = asdf.schema.load_schema(schema_uri)
  for subschema in schema.get("allOf", []):
    if asdf.generic_io.resolve_uri(schema_uri, subschema.get("$ref")) == ref_uri:
      return
  assert False, f"Schema '{schema_uri}' does not reference '{ref_uri}' in a top-level allOf combiner"
```
then we can use it here:
```python
assert_referenced_schema(ReferenceFileModel.schema_url, RomanDataModel.schema_url)
```
and also in the `FlatModel` test:
```python
assert_referenced_schema(FlatModel.schema_url, ReferenceFileModel.schema_url)
```

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
...
         except Exception:
-            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model class {model_class.__name__}"
+            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model " \
+                          f"class {model_class.__name__}"
+
+
+
+# Helper function to test required keywords
+# meta = dictionary of required key/value pairs to ensure each are required
+def confirm_required_keywords(tmp_path, meta):
```
eslavich, line 30: I think you may be able to do a lot of this with existing `DataModel` methods:

```python
dm = RomanDataModel(meta)
for required_key in dm.keys():
  original_value = dm[required_key]
  dm[required_key] = None
  with pytest.warns(ValidationWarning):
    dm.validate()
  dm[required_key] = original_value
```
- DataModel.keys returns an iterator of all paths in the model (like "meta.instrument.name")
- Setting the value of a key to `None` is equivalent to removing it
- There is no need to check the length of the captured warnings, since `pytest.warns(ValidationWarning)` already asserts that the warning occurs at least once

romancal/datamodels/open_impl.py
```diff
@@ -86,8 +87,14 @@ def _select_model_class(asdf_file):
     # - Use RomanDataModel for all data but select the schema according to one of the above
     # - ???
 
-    # If we keep it this route, need to check for the existence of reftype first
-    if asdf_file["meta"].get("reftype") == "FLAT":
-        return FlatModel
+    # Check for the existence of reftype to indicate a reference file model
+    if asdf_file["meta"].get("reftype", None):
```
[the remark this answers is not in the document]
PaulHuwe: Changed.

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -79,10 +79,11 @@ properties:
             enum: [WFI01, WFI02, WFI03, WFI04, WFI05, WFI06, WFI07, WFI08, WFI09,
                    WFI10, WFI11, WFI12, WFI13, WFI14, WFI15, WFI16, WFI17, WFI18,
                    ANY, N/A]
-          filter:
+          optical_element:
```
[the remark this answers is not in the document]
PaulHuwe: Thanks.

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -79,10 +79,11 @@ properties:
...
             title: Name of the filter element used
             type: string
             enum: [F062, F087, F106, F129, W146, F158, F184, GRISM, PRISM, CLEAR,
                    N/A, ANY, UNKNOWN]
-
-
+        required: [detector,optical_element,name]
+    required: [instrument,telescope]
+additionalProperties: true
```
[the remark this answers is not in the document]
PaulHuwe: Removed.

romancal/datamodels/schemas/reference_files/flat.schema.yaml
```diff
@@ -19,4 +19,4 @@ allOf:
       title: Error array
       default: 0.0
       datatype: float32
-
+additionalProperties: true
```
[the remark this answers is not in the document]
PaulHuwe: Removed.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
```diff
@@ -29,4 +31,6 @@ allOf:
           type: string
         useafter:
           title: Use after date of the reference file
-          type: string
+          tag: "tag:stsci.edu:asdf/time/time-1.1.0"
+      required: [date,author,description,pedigree,useafter,reftype]
+additionalProperties: true
```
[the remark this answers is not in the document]
PaulHuwe: Removed.

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
+import pytest
+
 import asdf
+import numpy as np
+from astropy.time import Time
+from copy import deepcopy
 
 from romancal.datamodels import RomanDataModel
+import romancal.datamodels as DataModels
```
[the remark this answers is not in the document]
PaulHuwe: Changed (and cascaded through the file).

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -79,10 +79,11 @@ properties:
...
                    ANY, N/A]
-          filter:
+          optical_element:
             title: Name of the filter element used
             type: string
             enum: [F062, F087, F106, F129, W146, F158, F184, GRISM, PRISM, CLEAR,
                    N/A, ANY, UNKNOWN]
-
-
+        required: [detector,optical_element,name]
```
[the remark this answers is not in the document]
PaulHuwe: Spaced.

romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -16,7 +16,7 @@ properties:
       telescope:
         title: Telescope used to acquire the data
         type: string
-        enum: [Roman]
+        enum: ["ROMAN"]
```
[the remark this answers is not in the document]
PaulHuwe: Unquoted.

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
```diff
@@ -29,4 +31,6 @@ allOf:
           type: string
         useafter:
           title: Use after date of the reference file
-          type: string
+          tag: "tag:stsci.edu:asdf/time/time-1.1.0"
```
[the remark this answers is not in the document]
PaulHuwe: Unquoted.

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,186 @@
...
+from stdatamodels.validate import ValidationWarning
 
+# Helper class to iterate over model subclasses
+def iter_subclasses(model_class):
```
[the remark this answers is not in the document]
PaulHuwe: Init updated.

romancal/datamodels/tests/test_models.py
PaulHuwe: Refactored.

romancal/datamodels/tests/test_models.py
PaulHuwe: Grabbed.

romancal/datamodels/tests/test_models.py
PaulHuwe: Assertion helped.

romancal/datamodels/tests/test_models.py
PaulHuwe: Replaced and records removed.

romancal/datamodels/tests/test_models.py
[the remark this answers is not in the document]
PaulHuwe: Updated the __init__.py file.

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,192 @@
...
-            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model class {model_class.__name__}"
+            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model " \
+                          f"class {model_class.__name__}"
+
+
+# Helper function to test required keywords
+# meta = dictionary of required key/value pairs to ensure each are required
+# model_class = the class of the model to be tested
+def confirm_required_keywords(tmp_path, meta, model_class):
+    file_path = tmp_path / "test.asdf"
```
nden, line 30: What is `tmp_path` here and what is this line supposed to do?

PaulHuwe: This needs to be removed, thanks for the catch. (It was for the code that Ed's suggestion replaced.)

romancal/datamodels/tests/test_models.py
```diff
@@ -1,16 +1,190 @@
...
         except Exception:
-            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model class {model_class.__name__}"
+            assert False, f"Unable to load schema URI '{model_class.schema_url}' for model " \
+                          f"class {model_class.__name__}"
+
+
+# Helper function to test required keywords
+# meta = dictionary of required key/value pairs to ensure each are required
+# model_class = the class of the model to be tested
+def confirm_required_keywords(tmp_path, meta, model_class):
```
eslavich, line 29: I had a chat with @nden and @jdavies-st about this and our general advice is not to include this kind of check in the romancal tests.  This helper is testing a) that the schema has certain required attributes and b) that stdatamodels is properly enforcing that requirement.  We have tests in stdatamodels itself that cover its ability to enforce a schema, so b) is redundant.

For a), the question is, what kind of bug or issue would this catch?  We would have to update the test for a new required attribute but forget to add it to the schema.  Or we'd see a failure if we added a new required attribute to the schema but forgot to maintain the test.  But in both cases these are issues introduced by maintaining two parallel lists of required attributes.  In other words, this test creates a new opportunity for bugs and then guards against them.  IMO there's little enough net value that it would be better not to have the test.

jdavies-st: Agree.

Also, I would advise against having try/except blocks in tests, as that exception handling will mask the actual error you're seeing in the test.  Rely on `pytest` assertion rewriting to give a useful message, and if you're expecting an exception use the `with pytest.raises()` context manager to find it.

PaulHuwe: I am guessing that I should keep the failed enum case as a known failed test?

PaulHuwe: If so, these changes are implemented and ready for review.

eslavich (approved):
LGTM!

PR #44, opened 2020-11-13
Removed stpipe docs

dmggh:
The docs for stpipe are common to JWST and Roman, so are not included in the Roman build.


PaulHuwe (approved):

ddavis-stsci (approved):
Looks good to me.

PR #48, opened 2020-11-19
Additional files required by readthedocs

dmggh:
Adding more files required for the readthedocs build, as shown from the build log. There is 1 more error (which I am still investigating) from that log, regarding a missing class in the datamodels module.


ddavis-stsci (approved):
LGTM

PR #49, opened 2020-11-20
Corrected path.

dmggh:
This change corrects an error in a path in the docs that appeared in the readthedocs build log.


PaulHuwe (approved):

ddavis-stsci (approved):
LGTM

nden:
I see the readthedcs build passes now. Because docs/index.rst is an empty file there's nothing on the page.
Also the correct readthedocs badge needs to be copied to romancal/README.rst.

PR #54, opened 2020-11-26
Adding DarkModel & Schema

PaulHuwe:
Initial implementation of the DataModel and schema for dark current files.

### DIFF
--- a/romancal/datamodels/dark.py
+++ b/romancal/datamodels/dark.py
@@ -1,0 +1,30 @@
+from .referencefile import ReferenceFileModel
+# Saved for DQ code implementation
+# from .dynamicdq import dynamic_mask
+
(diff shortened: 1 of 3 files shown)


nden:
This looks good for now but let's wait until Thu and see if we get additional directions on the file structure.

One thing I want to point out is that in the future we may want to exclude the `core` schema from the `reference_file` schema. The reason is that `core` will grow and represent a science file, including many attributes not needed by reference files. This change can wait though until there' s more clarity about the data.

eslavich (approved):
This looks good and seems to match the sample dark reference that Tyler sent me today.

PaulHuwe:
He added one thing that I want to double check before merging this.

PaulHuwe:
Checked for now. As we iterate on this, may tweak it further.

PR #61, opened 2020-12-03
Replaced JWST references with Roman

dmggh:
Replaced references to JWST with references to Roman in the readthedocs Roman Cal documentation.  Some areas will need embellishment or clarification as code development progresses.

### DIFF
--- a/docs/roman/introduction.rst
+++ b/docs/roman/introduction.rst
@@ -252,8 +252,8 @@ For example, to save the result from the dark current step of
     $ strun calroman_detector1.cfg roman00017001001_01101_00001_uncal.asdf
         --steps.dark_current.output_file='intermediate'
 
-A file, ``intermediate_dark_current.asdf``, will then be created. Note that the
-suffix of the step is always appended to any given name.
+An asdf file named ``intermediate_dark_current.asdf`` will then be created. Note 
+that the suffix of the step is always appended to any given name.
 
 You can also specify a particular file name for saving the end result of
 the entire pipeline using the ``--output_file`` argument also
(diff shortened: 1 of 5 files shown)


docs/roman/introduction.rst
```diff
@@ -262,11 +245,11 @@ individual steps, you have two options:
     This option will save the step results using the name specified.
 
 For example, to save the result from the dark current step of
-``calwebb_detector1`` in a file named based on ``intermediate``, use
+``calroman_detector1`` in a file named based on ``intermediate``, use
```
ddavis-stsci, line 248: Does this add the ".asdf" or is the file just called "intermediate"? If the asdf is not added can we just call the file intermediate.asdf to be clear?

ddavis-stsci (approved):

dmggh:
I redid the readthedocs build, which built successfully. But in the built docs it appears that some of the JWST->Roman changes I made to introduction.rst and other files do not appear in the built version. I do not understand why, so I will look into this.

dmggh:
I wiped the readthedocs environment and did a rebuild, which appears to solve the problem. Now the built document contains my latest updates - all of the previous JWST references have been correctly replaced. Unless someone has any objections, I think this completes what is needed for RCAL-18. The built docs are at : https://roman-cal-pipeline.readthedocs.io/en/latest/index.html

ddavis-stsci:
Can you also remove the CGI references in flat_selection.inc?
Thanks,
    Dave

dmggh:
Yes, I think that is already been noticed for RCAL-49.


________________________________

PR #62, opened 2020-12-03
Kmacdonald stsci patch 1

kmacdonald-stsci:
I created a github workflow to use github actions.  Please take a look at the new workflow in `.github/workflow`.  To make this work I used `requirements.txt`, which I created using `pip freeze`.  I think there are probably a lot of requirements in there that are unnecessary.


nden:
You can see what the requirements are in `setup.cfg`, under `install_requires`, `docs`, and `test`.

jdavies-st (commented):
I think in order for this to run the repo needs actions turned on in

Settings > Actions

Also, the new CI github action should be named something like `ci.yml` so that it's clear what it is for.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,68 @@
...
+name: Python package
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CRDS_SERVER_URL: "https://jwst-crds.stsci.edu"
```
pllim, line 13: Need to make sure downloads from server won't hit limits to the point you get charged real money for CI; if it does, work with management to see if charges are acceptable. If not, might want to consider running data intensive CI on internal Jenkins instance.

Also not sure if the server would start to blacklist you if you hit it too much. If it does, you need to work with ITSD to see if Actions IPs can be whitelisted.

romancal/datamodels/core.py
```diff
@@ -7,4 +7,3 @@ class RomanDataModel(DataModel):
     model class for data that does not identify its model type.
     """
     schema_url = "http://stsci.edu/schemas/roman_datamodel/core.schema"
```
pllim, line 9: Is HTTPS not possible?

pllim (commented):
Probably need to fix the failure first: https://github.com/kmacdonald-stsci/romancal/runs/1538081149?check_suite_focus=true

pllim:
> I think in order for this to run the repo needs actions turned on in

It will run when this is merged and this repo has `.github/workflows`. The first PR like this always only run on fork with the addition until it is merged.

nden:
@kmacdonald-stsci Please rebase on main to pick up the latest changes. This will fix the test error.
Also squash your commits and we can merge this.

PR #68, opened 2020-12-07
Update README file ro remove JWST refs and regression test examples, Rcal 65

ddavis-stsci:
Updated the changes file and removed JWST refs from README

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
-0.0.1 (unreleased)
+0.1.0 (unreleased)
 ===================
 
-The 0.0.0 release is the initial release for I&T testing.
+The 0.1.0 release is the initial release for I&T testing.
 
 datamodels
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -0,0 +1,11 @@
+0.0.1 (unreleased)
```
dmggh, line 1: Should that be 0.0.0 ?

ddavis-stsci:
I'm assuming that our first build will be 0.0.1 but I'm not sure. I was 
using 0.0.0 as a placeholder.
I should be the same in any case.

sosey (changes requested):

ddavis-stsci:
Closing this PR, the changes here were include in PR #70

ddavis-stsci:
Closing this PR, the changes here were include in PR #70

PR #71, opened 2020-12-10
adding requirements file for sdp delivery

nden:
This ended up more than I expected.
- Added dependency requirements for the upcoming release
- Fixed documentation build
- Added/removed documentation so that it reflects the state of the pipeline code.


ddavis-stsci (approved):
The only thing I see is that we still have "Keyword" in references_general.rst, but I don't think we need to change it now.

dmggh (approved):
Is error propagation for flat-fielding to be added later once the variance arrays of the SCI exposure exist ?

docs/roman/references_general/references_general.rst
```diff
@@ -0,0 +1,84 @@
...
+| :ref:`FLAT <flat_reffile>`                       | :ref:`flatfield <flatfield_step>`           |
++--------------------------------------------------+---------------------------------------------+
+
+.. _`Standard ASDF metadata`:
+
+Standard ASDF metadata
+======================
+
+Al Roman science and reference files are ASDF files.
+
```
dmggh, line 63: Is 'Keyword' (for asdf) equivalent to its FITS usage ?

nden:
@dmggh Yes, Once we have the flat field step implemented we'll start adding the error propagation.

dmggh (commented):
Is 'Keyword' (for asdf) equivalent to its FITS usage ?

docs/roman/references_general/references_general.rst
nden: `Keyword` is not generally used for ASDF but yes, it's the closest to FITS equivalent. 
In YAML the term is `field` which is not too popular. So after talking it over with others we agreed on using `attribute`

PR #74, opened 2020-12-10
Added REFTYPE keywords from Roman Documentation.

PaulHuwe:
Added the REFTYPEs from the following document to the Reference File Schemas.
https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=SCSB&title=Roman+Reference+Files

### DIFF
--- a/romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
@@ -12,7 +12,8 @@ allOf:
         reftype:
           title: Reference file type
           type: string
-          enum: [FLAT, BASE]
+          enum: [AREA, BASE, DARK, FLAT, GAIN, IPC, LINEARITY, PERSAT, PHOTOM, READNOISE, REFPIX,
+                 SATURATION, SUPERBIAS, TRAPDENSITY, TRAPPARS,]
         pedigree:
           title: The pedigree of the reference file
           type: string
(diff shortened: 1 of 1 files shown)


romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
```diff
@@ -12,7 +12,8 @@ allOf:
         reftype:
           title: Reference file type
           type: string
-          enum: [FLAT, BASE]
+          enum: [AREA, BASE, DARK, FLAT, GAIN, IPC, LINEARITY, PERSAT, PHOTOM, READNOISE, REFPIX,
+                 SATURATION, SUPERBIAS, TRAPDENSITY, TRAPPARS,]
```
ddavis-stsci, line 16: I don't think BASE is a valid ref type. That was just my shorthand for a placeholder. 
Should there be a comma at the end ? TRPPARS,]  --> TRAPPARS]

ddavis-stsci (changes requested):

romancal/datamodels/schemas/reference_files/referencefile.schema.yaml
PaulHuwe: Base was a hold over from JWST that i left in case it had some special meaning in the pipeline. I can remove it, if you like.

Python allows for trailing commas in lists (which makes things like copy & paste and appending easier), but I can remove it if you like.

ddavis-stsci:
I think BASE should go, if we need it for some reason we'll just add it 
back.

The comma is more of a style issue. While python doesn't care it always 
looks like someone forgot
to finish the list. The convention in JWST is to not have a trailing comma.
So I'd suggest we follow the JWST convention.

PaulHuwe:
Fixed and committed.

ddavis-stsci (approved):
Looks good!

PR #75, opened 2020-12-16
Refactored reference file datamodel classes into a subdirectory.

PaulHuwe:
Refactoring the reference datamodel objects into their own folder. I will do the same with future subclasses (image, ramp, etc.). This required some test tweaks.


eslavich (approved):
LGTM.  I like that it matches the structure of the schema path, and it'll be nice to have all the reference models collected in one place.

PR #78, opened 2020-12-17
Removing Travis CI script.

kmacdonald-stsci:
Removed the Travis CI script to migrate fully to github actions.


nden:
GA work now. 🎉 

Some of the documentation examples fail. Could you also fix these in the current PR.
We can't really fix them at the moment, because they are importing models that don't exist, but we can skip them by adding

```
.. doctest-skip::
```
immediately before each example. I thnk you will need to skip three examples in the docs in this way and then we'll get a clean run. two are in docs/roman/datamodels/models.rst and one in docs/roman/datamodels/metadata.rst.

ddavis-stsci (approved):

PR #80, opened 2021-01-09
First version of flat fielding code ( RCAL-60 )

dmggh:
This PR includes code assuming access to libraries that have not yet been installed for Roman (DQ, CRDS, handling of ERR arrays, etc).


ddavis-stsci (approved):

dmggh:
The JWST version of this code is used for multiple modes and instruments; there is at least one (spectroscopy) mode for which a reference array can be larger than the science array. For expediency I left this in for Roman.

PR #81, opened 2021-01-09
Initial version of DQ model and required files.

PaulHuwe:
Here is an initial stab at the DQ Array Schema & functions/definition file. The latter required the addition of a library utility file.

I have removed the following keywords (listed with the code they appear in and reason for removal):
MSA_FAILED_OPEN - msaflagopen/msaflag_open.py [Grumm recommendation]
OUTLIER - outlier_detection/outlier_detection.py [Grumm recommendation]
RC - straylight/straylight.py [MIRI MRS correction]
OPEN - straylight/straylight.py [MIRI MRS correction] & msaflagopen/msaflag_open.py [MSA]
ADJ_OPEN - straylight/straylight.py [MIRI MRS correction]
I replaced the preceding keywords with RESERVED_X to match other unassigned powers of 2.

The following keywords don't appear to be used by the CAL pipeline, and I am wondering if some of them can be dropped (some are obviously essential):
GOOD
DROPOUT
UNRELIABLE_ERROR
WARM
LOW_QE
TELEGRAPH
NONLINEAR
BAD_REF_PIXEL
UNRELIABLE_BIAS
UNRELIABLE_DARK
UNRELIABLE_RESET
OTHER_BAD_PIXEL
I would appreciate some feedback on if any of these can be removed as well.


romancal/datamodels/models/image/dqflags.py
```diff
@@ -0,0 +1,150 @@
...
+    Using Roman mnemonics:
+    TBD
+    """
+    bit_flags_dm = bit_flags
+    if isinstance(bit_flags, str):
+        dm_flags = {
+            key: str(val)
+            for key, val in pixel.items()
+        }
+        bit_flags_dm = multiple_replace(bit_flags, dm_flags)
```
eslavich, line 106: Do we need to worry about flag names that are substrings of other flag names?  If so, we ought to add `r'\b'` to the beginning and ending of each flag.

romancal/datamodels/models/image/dqflags.py
```diff
@@ -0,0 +1,150 @@
...
+# are equivalent to the pixel-specific flags.
+group = {'GOOD':       pixel['GOOD'],
+         'DO_NOT_USE': pixel['DO_NOT_USE'],
+         'SATURATED':  pixel['SATURATED'],
+         'JUMP_DET':   pixel['JUMP_DET'],
+         'DROPOUT':    pixel['DROPOUT'],
+}
+
+
+def interpret_bit_flags(bit_flags, flip_bits=None):
```
eslavich, line 71: How does that dq_def schema relate to this function?  If dq_def redefines relationships between flag names and values, then will this not work?

romancal/lib/basic_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    The replacements are done in a single-pass. This means that a previous
+    replacement will not be replaced by a subsequent match.
+
+    Parameters
+    ----------
+    string: str
+        The source string to have replacements done on it.
+
+    rep_dict: dict
+        The replacements were key is the input substring and
```
dmggh, line 20: typo: "were" -> "where"

ddavis-stsci (approved):
I don't see any of the keywords that should obviously be dropped, at least for now.

romancal/datamodels/models/image/dqflags.py
```diff
@@ -0,0 +1,150 @@
...
+    """Converts input bit flags to a single integer value (bit mask) or `None`.
+
+    Wraps `astropy.nddate.bitmask.interpret_bit_flags`, allowing the Roman
+    bit mnemonics to be used in place of integers.
+
+    Parameters
+    ----------
+    bit_flags : int, str, list, None
+        See `astropy.nddate.bitmask.interpret_bit_flags`.
```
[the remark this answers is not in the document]
PaulHuwe: Fixed.

romancal/lib/basic_utils.py
PaulHuwe: Fixed

romancal/datamodels/models/image/dqflags.py
eslavich: Never mind!  I didn't notice that the code in `multiple_replace` sorts the keys with the longest first.

romancal/datamodels/models/image/dqflags.py
PaulHuwe: It related to the dynamic dq code.

PR #88, opened 2021-01-22
Minor corrections for RCAL-60

dmggh:
Fixed import error in  flatfield/tests/test_flatfield.py ; fixed flake8 complaints in flatfield/flat_field.py.


ddavis-stsci (approved):
LGTM

nden:
Tests are failing. There's no` __init__` in `flatfiled`.

dmggh:
The jwst file is below. What should __version__ be for Roman ? 


from jwst.flatfield.flat_field_step import FlatFieldStep

__version__ = '0.7.7'

nden:
I'm not sure what version you are asking about. I think the test are failing because there is no `romacal/flatfield/__init__.py` file which technically prevents imports.
This is the error
```
from romancal.flatfield import FlatFieldStep
E   ImportError: cannot import name 'FlatFieldStep' from 'romancal.flatfield' (unknown location)
```
`flatfield` needs to be a subpackage.

ddavis-stsci:
And we can start with version 0.1.0 for the flat field code.

dmggh:
For jwst. jwst/flatfield/__init__.py is the following 2 lines, the second is a version :

from jwst.flatfield.flat_field_step import FlatFieldStep
__version__ = '0.7.7'

I assume that for Roman, the analogous file romancal/flatfield/__init__.py should be

from romancal.flatfield.flat_field_step import FlatFieldStep
__version__ = '0.1.0'

nden:
@dmggh Are you looking at some really old version of the jwst file? 
We stopped versioning individual steps a while back. I don't think we should version the individual steps in romancal either.

dmggh:
Yes, sorry about that. By analogy with the CURRENT JWST  __init__.py, I think the Roman version should be :

from .flat_field_step import FlatFieldStep
__all__ = ['FlatFieldStep']

nden:
Agreed

PR #91, opened 2021-01-22
Add __init__.py

dmggh:
Tests require __init__.py


ddavis-stsci (approved):
Looks like we just need the updated stpipe next.

PR #96, opened 2021-01-27
Added variable objects to image schema.

PaulHuwe:
Added data variable objects to the image schema. Commented out variance and pathloss correction schemas until they are implemented. 

One question for reviewers: do we still need the drizzle correction arrays?


ddavis-stsci (approved):
You can drop the pathloss corrections. Those are for spectra and are to corrects based on the source position in the slit.

PR #108, opened 2021-02-09
First version of driver for ramp fitting step.

dmggh:
First commit for porting ramp fitting to Roman; for RVAL-72. This first version of the module ramp_fitting_step.py handles only single-integration datasets as will be the case for Roman. This version has been modified to recast the components of the input Roman Ramp Data Model  to a JWST Ramp Data Model, which is passed to the lower level functions in ramp_fit.py and utils.py; the results are recast back into a Roman Data Model.  It seems that the JWST versions of ramp_fit.py and utils.py can be used for Roman, but tests (once stpipe is in place) may prove otherwise.

### DIFF
--- a/romancal/ramp_fitting/ramp_fit_step.py
+++ b/romancal/ramp_fitting/ramp_fit_step.py
@@ -1,0 +1,126 @@
+#! /usr/bin/env python
+#
+import logging
+import numpy as np
+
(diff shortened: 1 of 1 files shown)


ddavis-stsci (approved):
LGTM

PR #109, opened 2021-02-09
Add developer CRDS setup instructions to readme

eslavich:
One of the tests I added recently requires access to a CRDS cache, but I forgot to add instructions on setting that up.


PaulHuwe (approved):
Works!

PR #110, opened 2021-02-09
Ramp Model Schemas & Models

PaulHuwe:
Implemented ramp models & schemas. 
Moved group table from level1.schema to group.schema. 
Fixed JWST->Roman wording on core.schema.


romancal/datamodels/ramp.py
```diff
@@ -0,0 +1,41 @@
+from .core import RampDataModel
+#from ..lib.basic_utils import deprecate_class
+
+class RampModel(RomanDataModel):
+    """
+    A data model for 4D ramps.
```
dmggh, line 6: Shouldn't this be for 3D ramps, as there will only be 1 integration ?

PaulHuwe: Yup.. and I need to fix it everywhere.

PaulHuwe: Fixed.

romancal/datamodels/ramp.py
```diff
@@ -0,0 +1,39 @@
+from .core import RomanDataModel
+
+class RampModel(RomanDataModel):
+    """
+    A data model for 4D ramps.
```
ddavis-stsci, line 5: 4D -> 3D

romancal/datamodels/ramp.py
```diff
@@ -0,0 +1,39 @@
...
+         Error array
+
+    zeroframe : numpy float32 array
+         Zeroframe array
+
+    group : numpy table
+         group parameters table
+
+    int_times : numpy table
+         table of times for each integration
```
ddavis-stsci, line 28: With one integration does this need to be a table?

romancal/datamodels/rampfitoutput.py
```diff
@@ -0,0 +1,42 @@
+from .core import RomanDataModel
+
+class RampFitOutputModel(RomanDataModel):
+    """
+    A data model for the optional output of the ramp fitting step.
+
+    In the parameter definitions below, `n_int` is the number of
+    integrations, `max_seg` is the maximum number of segments that
```
ddavis-stsci, line 8: remove `n_int` here and below.

romancal/datamodels/ramp.py
PaulHuwe: Good catch, this was actually removed - as it was for others. Deleted.

romancal/datamodels/ramp.py
PaulHuwe: Fixed.

romancal/datamodels/rampfitoutput.py
PaulHuwe: Fixed this as well as mentions of nread.

ddavis-stsci (approved):
LGTM

romancal/datamodels/__init__.py
```diff
@@ -22,4 +22,6 @@
 
            "ImageModel",
 
+           "RampModel", "GLS_RampFitModel", "RampFitOutputModel",
```
nden, line 25: These need to be imported from the respective modules in order to be able to define them here.
I'll fix it in #115

PR #114, opened 2021-02-10
Create readnoise schema.

dmggh:
Create a schema for the readnoise reference file, at romancal/datamodels/schemas/reference_files/readnoise.schema.yaml.
This satisfies RCAL-77 .

### DIFF
--- a/romancal/datamodels/schemas/reference_files/readnoise.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/readnoise.schema.yaml
@@ -1,0 +1,15 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/readnoise.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_gainfact.schema
+- $ref: bunit.schema
+- type: object
+  properties:
+    data:
+      title: Read noise
+      default: 0.0
+      ndim: 2
+      datatype: float32


ddavis-stsci (approved):

romancal/datamodels/schemas/reference_files/readnoise.schema.yaml
```diff
@@ -0,0 +1,15 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/readnoise.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_gainfact.schema
+- $ref: bunit.schema
```
nden, line 8: We are using ASDF files, no need to implement `bunit` which is FITS specific.

romancal/datamodels/schemas/reference_files/readnoise.schema.yaml
```diff
@@ -0,0 +1,15 @@
...
+- type: object
+  properties:
+    data:
+      title: Read noise
+      default: 0.0
+      ndim: 2
+      datatype: float32
```
nden, line 15: We should stop using FITS description of ndarrays in schemas and use a reference to ndarray.

`data: $ref: ../core/asdf/ndarray-1.0.0`

nden (changes requested):

romancal/datamodels/schemas/reference_files/readnoise.schema.yaml
dmggh: So  should that single line replace the 'data:' line and the 4 lines that follow it ?

eslavich: I recommend using `tag:stsci.edu:asdf/core/ndarray-1.0.0` here instead of `$ref`, which saves a redundant validation pass over the ndarray schema.

The `ndim` and `datatype` validators are ASDF features that let you restrict the dimensions and datatype of the array.  They're also used by stdatamodels, along with `default`, to create empty arrays of the correct type and default value on a new model instance.  Without them, this won't work:

```python
In [4]: model = ReadnoiseModel((4096, 4096))

In [5]: model.data
Out[5]:
array([[0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       ...,
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.],
       [0., 0., 0., ..., 0., 0., 0.]], dtype=float32)
```

dmggh: @eslavich : based on your comments here and for for PR#131, do you recommend that the final 3 lines of the file be:
data: tag:stsci.edu:asdf/core/ndarray-1.0.0
    ndim: 2
    datatype: float32

dmggh: ... with the final 2 lines properly indented.

eslavich: It should look like this (with properties in whatever order pleases you):
```yaml
data:
  title: Read noise
  tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
  datatype: float32
  default: 0.0
  ndim: 2
```

dmggh:
Closing, as this is superseded by PR# 137.

PR #115, opened 2021-02-10
CI improvements

nden:
- use tox
- add docs build
- fix badges
- add coverage report
- fix flat field tests
- remove subarray from flat field step (Roman does not have subarrays) @dmggh please review


romancal/stpipe/tests/test_integration.py
```diff
@@ -25,4 +25,4 @@ def test_entry_point():
     all_steps = entry_points.get_steps()
     roman_steps = [s for s in all_steps if s.package_name == "romancal"]
     tuples = {(s.class_name, s.class_alias, s.is_pipeline) for s in roman_steps}
-    assert tuples == set(get_steps())
+    assert len(tuples) == len(set(entry_points.get_steps()))
```
nden, line 28: @eslavich I wasn't sure how you wanted to test this. `get_steps` returns `StepInfo` objects and comparing to `tuples` fails. Let me know if you had something else in mind.

eslavich: Did the original assertion fail?  `integration.get_steps()` should return [tuples](https://github.com/spacetelescope/romancal/blob/main/romancal/stpipe/integration.py#L23)?

.github/workflows/roman_ci.yml
```diff
@@ -1,68 +1,109 @@
...
+
+          - name: Latest dependency versions w/coverage
+            os: ubuntu-latest
+            python: 3.8
+            toxenv: py38-cov
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Checkout code
```
jdavies-st, line 85: Over in `jwst`, before checking out the code, we do some package installation for the docs:

https://github.com/spacetelescope/jwst/blob/fec12a0e8eda7f930c33b8f0efaa440b5533a53f/.github/workflows/ci.yml#L105-L108

The conditional makes sure this only happens for tox env names that have "docs" in them.  This was cribbed from the way astropy did it until they switched to having a RTD integration as a check instead of doing it from the GA workflow.

romancal/stpipe/tests/test_integration.py
nden: yes, but I'll double check

eslavich: If the entry point wasn't installed, that would cause a failure.  It doesn't automatically install when pulling changes down even when the package is installed with `-e` -- it has to be pip install'ed again.

ddavis-stsci (approved):

PR #120, opened 2021-02-11
Create gain factor keyword schema.

dmggh:
Gain factor keyword is needed for the read noise schema ( RCAL-77 and PR #114 ). This is supported by PR #118.

Issue #79, opened 2021-01-06
Add tag validator to schema array properties


stscijgbot-rstdms:
_Issue [RCAL-70](https://jira.stsci.edu/browse/RCAL-70) was created on JIRA by [Edward Slavich](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eslavich):_

We should consider adding a tag validator to array properties in the model schemas:
 `tag: tag:stsci.edu:asdf/core/ndarray-1.0.0`
This is the validator that actually asserts that the object is of ndarray type, whereas others like "datatype" and "shape" only check certain properties. Not that it's likely that a non-ndarray object would coincidentally have a shape and datatype, but using the tag validator is more direct and better documentation.

stscijgbot-rstdms:
_Comment by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe) on [JIRA](https://jira.stsci.edu/browse/RCAL-70?focusedCommentId=531232#comment-531232):_

This work was spread out over several, content-related PRs: 31, 39, 40, 41, 42, 43, 44, 45, 47

RAD PR URLs:

spacetelescope/rad#31

spacetelescope/rad#39

spacetelescope/rad#40

spacetelescope/rad#41

spacetelescope/rad#42

spacetelescope/rad#43

spacetelescope/rad#44

spacetelescope/rad#45

spacetelescope/rad#47

PR #121, opened 2021-02-11
Gain Schema & Datamodel

PaulHuwe:
Gain and keyword schema and model. 
Init also updated to include dark model.

### DIFF
--- a/romancal/datamodels/reference_files/gain.py
+++ b/romancal/datamodels/reference_files/gain.py
@@ -1,0 +1,12 @@
+from .referencefile import ReferenceFileModel
+
+class GainModel(ReferenceFileModel):
+    """
+    A data model for 2D gain.
+
+    Parameters
+    __________
+    data : numpy float32 array
+         The gain
+    """
+    schema_url = "http://stsci.edu/schemas/roman_datamodel/reference_files/gain.schema"
--- a/romancal/datamodels/schemas/reference_files/gain.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/gain.schema.yaml
@@ -1,0 +1,14 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/gain.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_gainfact.schema
+- type: object
+  properties:
+    data:
+      title: The gain
+      default: 0.0
+      ndim: 2
+      datatype: float32
--- a/romancal/datamodels/schemas/reference_files/gain.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/gain.schema.yaml
@@ -12,3 +12,4 @@ allOf:
       default: 0.0
       ndim: 2
       datatype: float32
+      tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
--- a/romancal/datamodels/schemas/reference_files/gain.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/gain.schema.yaml
@@ -8,7 +8,7 @@ allOf:
 - type: object
   properties:
     data:
-      title: The gain
+      title: The detector gain map
       default: 0.0
       ndim: 2
       datatype: float32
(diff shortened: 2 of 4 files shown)


romancal/datamodels/schemas/reference_files/keyword_gainfact.schema.yaml
```diff
@@ -0,0 +1,15 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/keyword_gainfact.schema"
+type: object
+properties:
+  meta:
+    type: object
+    properties:
+      exposure:
```
dmggh, line 10: I see a potential conflict here - I put in a not-yet-approved PR #120 earlier this evening for this file, but my version lacks lines 8-10.

romancal/datamodels/schemas/reference_files/gain.schema.yaml
```diff
@@ -0,0 +1,14 @@
...
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/gain.schema"
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_gainfact.schema
+- type: object
+  properties:
+    data:
+      title: The gain
```
ddavis-stsci, line 11: A minor nit, can we change "The gain" --> "detector gain map"  or something similar.

ddavis-stsci (approved):
One minor point other wise it looks good to go.

romancal/datamodels/schemas/reference_files/gain.schema.yaml
PaulHuwe: Fixed.

romancal/datamodels/schemas/reference_files/keyword_gainfact.schema.yaml
PaulHuwe: Did you change how the object is referred to in the code? 8-10 affect the location (meta.exposure.gain_factor).

dmggh: Good point. I am revising the schema to match the code.

romancal/datamodels/reference_files/gain.py
```diff
@@ -0,0 +1,12 @@
+from .referencefile import ReferenceFileModel
+
+class GainModel(ReferenceFileModel):
+    """
+    A data model for 2D gain.
+
+    Parameters
+    __________
+    data : numpy float32 array
+         The gain
```
nden, line 10: This should also work with file names. And it does because the implementation is in stadatamodels but the `Parameters` signature is confusing. Also, it's a larger conversation for the team but if reference files are allowed to be initialized with a data array should the constructor also accept the `meta` attributes?

romancal/datamodels/schemas/reference_files/gain.schema.yaml
```diff
@@ -0,0 +1,14 @@
...
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_gainfact.schema
+- type: object
+  properties:
+    data:
+      title: The detector gain map
+      default: 0.0
+      ndim: 2
+      datatype: float32
```
nden, line 14: If this is a data array why not use 

`$ref: ../asdf/core/ndarray-1.0.0`

@eslavich Is this the correct way to reference ndarray?

eslavich: I'd recommend this to save a redundant validation pass:

```yaml
tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
```

There's an open issue to add this to the existing schemas: https://github.com/spacetelescope/romancal/issues/79

PaulHuwe: Tagged.

romancal/datamodels/reference_files/gain.py
PaulHuwe: I am a little confused - data in the gain schema is the values of the gain map. I don't understand how that is interchangeable with filenames?

nden: In case a `GainModel` is initialized with a file, something that the pipeline does all the time, the call will be:

```
model = GainModel(filename)
```
However, for testing purposes and also when generating reference files, it may be useful to initialize the model with a data array, which is the case I thought is currently covered in `Parameters`:
```
model = GainModel(data)
model.meta.pedigree = "Ground"
...
```
All I was saying is that while both use cases are covered already by stdatamodels, the Parameters section does not reflect this.

PaulHuwe:
Merging this now to resolve dependencies. I will address Nadia's comment in a new PR.

PR #122, opened 2021-02-12
make flatfield run

nden:
This cleans up the flatfield step to allow it to run with stpipe.
The CRDS servers are available only on the internal network so a unit test is not included.
This is a step towards setting up the Jenkins tests where such a test will be included.


ddavis-stsci (approved):
LG

dmggh (approved):

PR #128, opened 2021-02-16
RCAL-95: set default output_ext to .asdf

nden:
This PR sets the default extension for Roman products to `.asdf`. 
~It depends on spacetelescope/stpipe#17~

### DIFF
--- a/romancal/stpipe/core.py
+++ b/romancal/stpipe/core.py
@@ -8,7 +8,7 @@ class RomanStep(Step):
     Base class for Roman calibration pipeline steps.
     """
     spec = """
-    output_ext =  string(default='.asdf')
+    output_ext =  string(default='.asdf')    # Default type of output
     """
 
     @classmethod
--- a/romancal/stpipe/core.py
+++ b/romancal/stpipe/core.py
@@ -7,6 +7,9 @@ class RomanStep(Step):
     """
     Base class for Roman calibration pipeline steps.
     """
+    spec = """
+    output_ext =  string(default='.asdf')
+    """
 
     @classmethod
     def _datamodels_open(cls, init, **kwargs):
@@ -70,7 +73,7 @@ class RomanStep(Step):
         # JWST maintains a list of relevant suffixes that is monitored
         # by tests to be up-to-date.  Roman will likely need to do
         # something similar.
-        return name, " "
+        return name, "_"
 
 
 # RomanPipeline needs to inherit from Pipeline, but also
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,21 @@ stpipe
 ------
 
 - Create stpipe module which provides Roman-specific Step and Pipeline
-  subclasses. [#103].
+  subclasses. [#103, #128]
+
+flatfield
+---------
+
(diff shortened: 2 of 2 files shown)


PaulHuwe (approved):

romancal/stpipe/core.py
```diff
@@ -8,6 +8,8 @@ class RomanStep(Step):
     Base class for Roman calibration pipeline steps.
     """
 
+    output_ext =  '.asdf'
```
eslavich, line 11: I think we should add a spec string here:
```python
output_ext         = string(default='.asdf')     # Default type of output
```
This will automatically override the `stpipe.Step` definition of the parameter and allow us to merge this PR without waiting for an stpipe release.

romancal/stpipe/core.py
```diff
@@ -70,7 +72,7 @@ def remove_suffix(self, name):
         # JWST maintains a list of relevant suffixes that is monitored
         # by tests to be up-to-date.  Roman will likely need to do
         # something similar.
-        return name, " "
+        return name, "_"
```
eslavich, line 75: Oops, I think that was my bad, thanks...

romancal/stpipe/core.py
nden: This means adding configobj. Is this where we are going with romancal?

eslavich: Eventually we'll want to convert this to a traitlet, but that's a ways off.  The class attribute gets overwritten by `stpipe.Step` when the step is instantiated:

```python
In [1]: from romancal.stpipe import RomanStep

In [2]: s = RomanStep()
2021-02-16 10:21:39,225 - stpipe.RomanStep - INFO - RomanStep instance created.

In [3]: s.output_ext is None
Out[3]: True
```

romancal/extern/configobj/configobj.py
```diff
@@ -0,0 +1,2476 @@
+# configobj.py
```
eslavich, line 1: We don't need to vendorize configobj in this package because it's already included in stpipe at `stpipe.extern.configobj`.  I think I may have misunderstood your question at standup -- the spec string is required here, but if we need configobj we can import it from stpipe.

eslavich: (`jwst.extern.configobj` should have been removed, an oversight on my problem which I'll open a PR to fix in a moment)

nden: Ah, I did not realize it's in stpipe but it totally makes sense.

nden: `configobj` removed

romancal/stpipe/core.py
```diff
@@ -7,6 +7,9 @@ class RomanStep(Step):
     """
     Base class for Roman calibration pipeline steps.
     """
+    spec = """
+    output_ext =  string(default='.asdf')
```
eslavich, line 11: The trailing comment is actually important -- it's used as the description of the parameter in the strun help output.  I don't think it's inherited from Step now that we've overwritten output_ext, so we should probably keep it here as well.

eslavich (approved):
LGTM

PR #130, opened 2021-02-16
Rcal-83 Update optical element ENG -> F213

ddavis-stsci:
The engineering filter is to be officially removed from the set of elements and replaced with F213.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ fork and clone the `romancal` repo:
 
     cd <where you want to put the repo>
     git clone https://github.com/spacetelescope/romancal
-    cd roman
+    cd romancal
 
 *Note: `python setup.py install` and `python setup.py develop` commands do not work.*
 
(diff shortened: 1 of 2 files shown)


eslavich:
We'll need to update the CRDS validations as well, I'll create a CCD ticket

PaulHuwe (approved):

dmggh (approved):

eslavich:
@nden since this has been merged, should we make the update in crds before build 0.1 as well?

ddavis-stsci:
Ughh! I forgot about CRDS, and was trying to get it in before the code freeze at the end of the week.I think I can back that out if it is a problem.

eslavich:
We updated the value in crds and the change will be included in the release we make for romancal, so this shouldn't be a problem.

nden:
Thanks Ed!

ddavis-stsci:
That's great! Thanks.

PR #131, opened 2021-02-17
Create readnoise schema.

dmggh:
Create a schema for the readnoise reference file, at romancal/datamodels/schemas/reference_files/readnoise.schema.yaml.
This satisfies RCAL-77 . This PR addresses reviewer comments from PR#114.

PR #134, opened 2021-02-17
Gain factor keyword needed.

dmggh:
Gain factor keyword  (RCAL-91) is needed for the read noise schema ( RCAL-77 and PR #131 ). This PR addresses reviewer comments on PR #120.


ddavis-stsci (approved):
TGTM

PR #147, opened 2021-02-23
Initial implementation of mask datamodel and schema.

PaulHuwe:


romancal/datamodels/schemas/reference_files/mask.schema.yaml
```diff
@@ -0,0 +1,14 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/mask.schema"
+allOf:
+- $ref: referencefile.schema
+- type: object
+  properties:
+    dq:
```
eslavich, line 9: We've started adding the `tag:` validator to arrays in other schemas (well, one other schema so far):

https://github.com/spacetelescope/romancal/blob/main/romancal/datamodels/schemas/reference_files/readnoise.schema.yaml#L12

Should we do that here as well?

PaulHuwe: Tag added. I also updated RCAL-99 to add ndarray tags to all existing schemas.

ddavis-stsci (approved):
Looks Good with the tag

PR #179, opened 2021-03-09
Correct Model Docstrings and Make Models PEP 8 Compliant

PaulHuwe:
Changes:
- Made datamodels PEP 8 compliant. Removed remaining zeroframe and pathloss from models. 
- Added variance schema for image and rampoutput files. 
- Removed a seemingly incorrectly named file. @dmggh can you confirm that keyword_gainfact.schema is extraneous (given that we use keyword_gainfact.schema.yaml)?

Tests:
- Datamodels tests did not require updating. These changes passed those tests.

Documentation:
- This ticket is primarily an update to (as yet not fully supported) documentation.

Ticket:
- RCAL-99


dmggh:
Yes, keyword_gainfact.schema is extraneous and can be removed.

ddavis-stsci:
Why can we remove gain factor? in https://innerspace.stsci.edu/display/ROMAN/Use+Cases+for+Level+2+Science+Data+Processing
it is listed in the steps, 2k on the "Standard Path".

romancal/datamodels/dynamicdq.py
```diff
@@ -25,7 +27,8 @@ def dynamic_mask(input_model):
             try:
                 standard_bitvalue = dqflags.pixel[dqname]
             except KeyError:
-                log.warning('Keyword %s does not correspond to an existing DQ mnemonic, so will be ignored' % (dqname))
+                log.warning('Keyword %s does not correspond to an existing DQ mnemonic, '
+                            'so will be ignored' % dqname)
```
ddavis-stsci, line 31: Should we change keyword to something else? DQ flag?

ddavis-stsci (changes requested):
We need to resolve the gain factor.

ddavis-stsci (approved):
TGTM

romancal/datamodels/dynamicdq.py
PaulHuwe: dqflags is already used, so I think qdname should stay as is.

PR #180, opened 2021-03-12
Add cal_step to schema, needed for all steps

dmggh:
To satisfy RCAL-132. Adding cal_step to the core schema. cal_step is needed for all steps in the pipeline.

### DIFF
--- a/romancal/datamodels/schemas/core.schema.yaml
+++ b/romancal/datamodels/schemas/core.schema.yaml
@@ -17,6 +17,52 @@ properties:
           position_angle:
             title: "[deg] Position angle of aperture used"
             type: number
+      cal_step:
+        title: Calibration step information
+        type: object
+        properties:
+          assign_wcs:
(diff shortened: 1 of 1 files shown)


romancal/datamodels/schemas/core.schema.yaml
```diff
@@ -17,6 +17,52 @@ properties:
...
+          refpix:
+            title: Reference Pixel Correction
+            type: string
+          saturation:
+            title: Saturation Checking
+            type: string
+          superbias:
+            title: Superbias Subtraction
+            type: string
       calibration_software_version:
```
ddavis-stsci, line 66: Should we add the IPC step?

ddavis-stsci (changes requested):

dmggh:
Ok, I will add it. I had omitted it as that step was set to be skipped for JWST.

ddavis-stsci (approved):
TGTM

PR #183, opened 2021-03-13
Adding tests for dark current step.

dmggh:
This PR adds unit tests for the dark current subtraction step.  In dark_sub.py are updates which includes a workaround for a bug in the update function in the datamodel library,  and an added warning suppression for the case of an empty slice.

### DIFF
--- a/romancal/dark_current/dark_sub.py
+++ b/romancal/dark_current/dark_sub.py
@@ -5,6 +5,7 @@
 import numpy as np
 import logging
 from .. import datamodels
+import warnings
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


romancal/dark_current/dark_sub.py
```diff
@@ -140,11 +141,16 @@ def average_dark_frames(input_dark, ngroups, nframes, groupgap):
...
     # Create a model for the averaged dark data
     dny = input_dark.data.shape[1]
     dnx = input_dark.data.shape[2]
     avg_dark = datamodels.DarkModel((ngroups, dny, dnx))
+
+    # 03/12/21 this block is a workaround for datamodel's update bug
+    input_dark.meta.instrument.detector = 'WFI01'
+    avg_dark.meta.instrument.name = 'WFI'
+    avg_dark.meta.instrument.detector = 'WFI01'
+    avg_dark.meta.update(input_dark.meta._instance)
```
nden, line 153: What is the bug in datamodels and is there a ticket filed?

nden:
Another general comment - `readpatt` is not a valid attribute of the Roman datamodel. Unlike JWST it is always the same.

romancal/dark_current/dark_sub.py
dmggh: Perhaps it is not technically a bug, but according to @PaulHuwe , the dictionary function 'update' has some issues with recursion, so the code modification here is his brute force implementation.

dmggh:
Ah, yes - right about readpatt. I've changed the name of that variable to indicate that that information will later be retrieved from a MA table.

romancal/dark_current/dark_sub.py
dmggh: To link to this PR, I have created this stdatamodel GH PR regarding the update needed in the datamodel  : "Improve datamodel dictionary 'update' handling of recursion" - (https://github.com/spacetelescope/stdatamodels/issues/66. ) I have been unable to link these 2 PRs, because of what may be a permission problem.

ddavis-stsci (approved):
Looks good. 
I'm not sure we want to link the issues. I think that means we cannot close this one until the linked issue is resolved. We have a comment in the code which should work for now.

PR #192, opened 2021-03-25
RCAL-38: Regression tests infrastructure

### LINKED ISSUE #18: Port test infrastructure
_Issue [RCAL-38](https://jira.stsci.edu/browse/RCAL-38) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Setup test classes for the Roman Cal pipeline.

Use the jwst/tests as a template and copy/modify

base_classes.py
helpers.py
test_infrastructure.py
test_version.py

This should be broken up into a few tickets.

nden:
Closes #18

Resolves RCAL-38

**Description**

This PR ports the jwst RT infrastructure to romancal and adds one regression test.
The test will fail because the dates in the truth file and the output are different.
An option will be added to asdf diff to allow exclusion of a list of attributes.

In the future much of this infrastructure should be moved to a common place, most likely ci_watson.

Checklist
- [x] Tests

- [ ] Documentation (working on it

- [ ] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/romancal/lib/tests/helpers.py
+++ b/romancal/lib/tests/helpers.py
@@ -1,0 +1,42 @@
+import os
+import glob
+import io
+
+
+S3_BUCKET_NAME = "test-s3-data"
+
+
+class MockS3Client:
+    def __init__(self, s3_test_data_path):
+        self.s3_test_data_path = s3_test_data_path
+
+    def get_object(self, bucket_name, key):
+        assert self.object_exists(bucket_name, key)
+
+        with open(self._get_path(key), "rb") as f:
+            return io.BytesIO(f.read())
+
+    def object_exists(self, bucket_name, key):
+        if bucket_name != S3_BUCKET_NAME:
+            return False
+
+        return os.path.isfile(self._get_path(key))
+
+    def prefix_exists(self, bucket_name, key_prefix):
+        return any(self.iterate_keys(bucket_name, key_prefix))
+
+    def iterate_keys(self, bucket_name, key_prefix):
+        if bucket_name != S3_BUCKET_NAME:
+            return
+
+        for k in self._list_keys():
+            if k.startswith(key_prefix):
+                yield k
+
+    def _get_path(self, key):
+        return os.path.join(self.s3_test_data_path, key)
+
+    def _list_keys(self):
+        paths = glob.glob(self.s3_test_data_path + "/**", recursive=True)
+        paths = [p for p in paths if os.path.isfile(p)]
+        return [p.replace(self.s3_test_data_path, "") for p in paths]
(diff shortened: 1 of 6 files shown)


ddavis-stsci (approved):
LGTM

romancal/lib/tests/helpers.py
```diff
@@ -0,0 +1,42 @@
+import os
+import glob
+import io
+
+
+S3_BUCKET_NAME = "test-s3-data"
+
+
+class MockS3Client:
```
jdavies-st, line 9: This whole module is a good candidate for `stcal`.

romancal/regtest/test_wfi_flat_field.py
```diff
@@ -1,27 +1,33 @@
...
-# @pytest.fixture(scope='module')
-# def run_flat_field(rtdata_module):
-#     """ Run flat field on Level 2 imaging data."""
-#
-#     rtdata = rtdata_module
-#     rtdata.get_data("WFI/image/level2.asdf")
-#     step = FlatFieldStep()
-#     step.run(rtdata.input)
+@pytest.fixture(scope='module')
+def run_flat_field(rtdata_module):
```
jdavies-st, line 11: I would scope all fixtures in this module to `function` and use the `rtdata` fixture instead of `rtdata_module`.  `rtdata_module` is really designed for daisy-chaining one pipeline stage to another and retaining the filenames along the way, and then making sure the data from stage 1 ends up in the same tmpdir as stage 2 pipeline is expecting to find it.

If tests in a module are independent, then just use the simpler `rtdata` fixture.

jdavies-st (approved):
Looks good.  Just some minor comments below. And clearly some of this infrastructure could go into `ci-watson` in the near future.

romancal/lib/tests/helpers.py
nden: @jdavies-st Is `stcal` or `ci_watson` a better place for it?

jdavies-st: There's already an s3 client library that Ed wrote that both `jwst` and `crds` use:

https://github.com/spacetelescope/stsci-aws-utils

Maybe the mocker could go there?  @eslavich ?

nden:
Can we merge this @perrygreenfield @ddavis-stsci ?

perrygreenfield:
I haven't looked at it yet, but if you can wait until around lunchtime (I have a goddard meeting this morning) I will. Otherwise go ahead.

ddavis-stsci:
I don't see any problems with it so I'd go ahead.

perrygreenfield (approved):
I don't have much to add to the existing comments.

PR #200, opened 2021-04-07
Photom schema file (RCAL-109)

dmggh:
Closes #

Resolves RCAL-109

**Description**

This PR addresses RCAL-109. wfi_img_photom.schema.yaml is the WFI image Roman analog of the JWST file nrcimg_photom.schema.yaml.  As I understand , there is no wfss mode for Roman, so there is no equivalent of the JWST file nrcwfss_photom.schema.yaml.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
@@ -13,6 +13,8 @@ allOf:
       - name: optical_element
         datatype: [ascii, 12]
       - name: photmjsr
+        title: surface brightness, in MJy/steradian 
         datatype: float32
       - name: uncertainty
+        title: uncertainty of surface brightness, in MJy/steradian 
         datatype: float32
--- a/romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
+++ b/romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
@@ -1,0 +1,18 @@
+%YAML 1.1
+---
+$schema: http://stsci.edu/schemas/asdf/asdf-schema-1.0.0
+id: "http://stsci.edu/schemas/roman_datamodel/reference_files/wfi_img_photom.schema"
+title: WFI imaging photometric flux conversion data model
+allOf:
+- $ref: referencefile.schema
+- type: object
+  properties:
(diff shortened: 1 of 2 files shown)


romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
```diff
@@ -0,0 +1,18 @@
...
+allOf:
+- $ref: referencefile.schema
+- type: object
+  properties:
+    phot_table:
+      title: Photometric flux conversion factors table
+      datatype:
+      - name: optical_element
+        datatype: [ascii, 12]
+      - name: photmjsr
```
ddavis-stsci, line 15: It looks like this is MJy/steradian? Can we add some text to clarify that?

romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
```diff
@@ -0,0 +1,18 @@
...
+        datatype: float32
+      - name: uncertainty
+        datatype: float32
```
ddavis-stsci, line 18: If this is flux per some area do we need an area for each pixel? I assume that the JWST value is an average? or is all this information in the area ref file?

romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
dmggh: I added 2 titles.

romancal/datamodels/schemas/reference_files/wfi_img_photom.schema.yaml
dmggh: From the photom reference file, the JWST scalar conversion factor is copied to 'photmjsr', which is the nominal surface brightness in MJy/steradian. The area of each pixel (in steradians) is found by multiplying PIXAR_SR (the average area of the pixels, in steradians), by that pixel's area from the pixel area reference file, which is the dimensionless relative area of the pixel.

dmggh:
Are there any comments on this update ?

ddavis-stsci (approved):
Looks good, make sure you update the change log.

PR #212, opened 2021-05-04
RCAL-100 Integrate Tag approach into Roman pipelines

### LINKED ISSUE #211: Integrate Tag approach into Roman pipelines
The JWST pipelines generally do not use tags. This has some drawbacks as tags provide some extra functionality than the monolithic schema approach used for JWST. For example, the use of tags makes it easier to specify required keywords, and to customize how the ASDF data files are turned into datamodels. It uses the converter approach which is now standard for the Python ASDF library. Finally, it can discard much of the FITS cruft that is embedded in the JWST datamodels code that is not necessary for Roman.

perrygreenfield:
Closes #211

Resolves RCAL-100

**Description**

This PR addresses allows the Roman pipelines to make use of the new roman_datamodels package, which is based on using ASDF tags for the data file. For the most part, previous groups of info (e.g., exposure, observation, instrument, etc) present in the metadata the JWST data models now will appear as a typed node, although most of these nodes will have exactly the same object behavior. On the other hand, the ASDF converter approach used allows customized behavior to allow the ROMANCAL code to be consistent with JWST pipeline code. The best example of this is the divergence between how optical components are identified; for JWST it is by the attribute 'filters' but for Roman the attribute is 'optical_elements'. The converters allow aliasing optical_elements to filters so that we may used JWST code that expects the attribute filters. 

This new approach has been implemented for the flat field step.

It requires an updated version of CRDS to deal with the new attribute chains. That version has been installed in the dev version of the server.

Checklist
- [ ] Tests

- [ ] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


romancal/flatfield/flat_field_step.py
```diff
@@ -15,18 +16,20 @@ class FlatFieldStep(RomanStep):
 
     def process(self, input):
 
-        input_model = datamodels.open(input)
-
+        input_model = rdm.open(input)
```
eslavich, line 19: It might be wise to start encouraging the use of `Step.open_model` to make these step classes more reusable between missions

romancal/flatfield/flat_field_step.py
```diff
@@ -15,18 +16,20 @@ class FlatFieldStep(RomanStep):
...
+        self._parent = None
```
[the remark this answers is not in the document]
perrygreenfield: Beats me, so I took it out.

romancal/flatfield/flat_field_step.py
```diff
@@ -15,18 +16,20 @@ class FlatFieldStep(RomanStep):
...
+        self._datamodels_open = rdm.open
```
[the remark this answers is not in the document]
perrygreenfield: Also blown away.

romancal/flatfield/flat_field_step.py
```diff
@@ -15,18 +16,20 @@ class FlatFieldStep(RomanStep):
...
         # Get reference file paths
         reference_file_names = {}
         reffile = self.get_reference_file(input_model, "flat")
+        input_model2 = input_model.copy()
```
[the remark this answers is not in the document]
perrygreenfield: Ditto

romancal/flatfield/flat_field_step.py
```diff
@@ -1,7 +1,8 @@
 from ..stpipe import RomanStep
-from .. import datamodels
+# from .. import datamodels
```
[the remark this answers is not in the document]
perrygreenfield: Detritus. Vacuumed.

romancal/flatfield/flat_field_step.py
perrygreenfield: Good point. Done.  (but not yet pushed since there is a regtest issue I need to look at).

romancal/stpipe/core.py
```diff
@@ -4,6 +4,7 @@
 from stpipe import Step, Pipeline
 
 from .. import datamodels
+import roman_datamodels as rdatamodels
```
nden, line 7: we should agree on one alias `rdm` or `rdatamodels`.

nden (commented):
flake8 complaints
```
./romancal/stpipe/core.py:6:1: F401 '..datamodels' imported but unused
./romancal/flatfield/flat_field_step.py:7:1: F811 redefinition of unused 'flat_field' from line 6
./romancal/flatfield/flat_field_step.py:26:9: F841 local variable 'input_model2' is assigned to but never used
```

romancal/stpipe/core.py
PaulHuwe: If I may vote, I strongly prefer rdm over rdatamodels (in fact, I'd prefer romandatamodels over the latter.. rdm is good for me either way).

PaulHuwe: I vote rdm

eslavich (approved):
LGTM!

PR #228, opened 2021-06-09
Update jump unit tests for Roman data (RCAL-114)

dmggh:
Closes #

Resolves RCAL-114

**Description**

This PR addresses RCAL-114; uses the jump code (which is now in stcal) to run 2 of the 3 unit test modules to test the jump code in stcal. The 3rd test module (test_jump_step.py) will be covered in a separate PR.

Checklist
- [X] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/romancal/jump/tests/test_detect_jumps.py
+++ b/romancal/jump/tests/test_detect_jumps.py
@@ -1,0 +1,752 @@
+import numpy as np
+from numpy.testing import assert_array_equal
+import pytest
+
+from romancal.jump.jump import detect_jumps
+from romancal.jump import constants
+
+dqflags = {
+    "GOOD": 0,
+    "DO_NOT_USE": 1,
+    "SATURATED": 2,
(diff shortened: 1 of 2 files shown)


ddavis-stsci (approved):
You'll need to update your imports,
from romancal.jump.jump import detect_jumps
E   ModuleNotFoundError: No module named 'romancal.jump.jump'

Otherwise looks good.

nden (changes requested):
@dmggh Please review spacetelescope/stcal#37 

In general unit tests which call functions in stcal with array inputs should be added to stcal and tests which use data models as input should be added to romancal. If there are tests in this draft PR which are covering code missed in spacetelescope/stcal#37  and use arrays as input add them to stcal. Otherwise jump detection in romancal needs tests using an input datamodel.

PR #262, opened 2021-07-26
Rcal 137 dq init

PaulHuwe:
Closes #

Resolves RCAL-137 & RCAL-92

**Description**

This PR addresses adds support for the DQ flagging step utilizing roman_datamodels. Tests have been updated for this code.

Some important Notes:
1. Build errors are due to CRDS files being out of date (needs to be done as part of switching romancal datamodels completely over to RDM)
2. RTD files will be in an another ticket.

This PR depends on 2 others:
a. https://github.com/spacetelescope/rad/pull/63
b. https://github.com/spacetelescope/roman_datamodels/pull/18

Checklist
- [ x  ] Tests

- [ x ] Documentation - inline documentation completed. RTD to be addressed in another ticket.

- [ x ] Change log

- [ x ] Milestone - 0.3

- [ x ] Label(s) - DQ, RCAL, team-azure

Issue #258, opened 2021-07-22
Implement Jump step via stpipe


stscijgbot-rstdms:
_Issue [RCAL-178](https://jira.stsci.edu/browse/RCAL-178) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Implement the Jump step code running via stpipe.

stscijgbot-rstdms:
_Comment by [David Grumm](https://jira.stsci.edu/secure/ViewProfile.jspa?name=grumm) on [JIRA](https://jira.stsci.edu/browse/RCAL-178?focusedCommentId=542045#comment-542045):_

Work on this ticket is blocked, awaiting updates in the datamodels library. Before I began work on this ticket, I  wanted to first verify that I could still run the flat field step (as I had first done several months ago), to make sure that the underlying libraries for running steps still worked. I could not run the flat step, as the updated datamodels library was not (and still is not) available,. As a result, I never reached the point of submitting a PR for RCAL-178.

dmggh:
Using the latest L2 file and mask file, I am able to run the flat field step, which indicates that the data models library was not a factor. I will file a PR, and proceed with work on RCAL-178

stscijgbot-rstdms:
_Comment by [David Grumm](https://jira.stsci.edu/secure/ViewProfile.jspa?name=grumm) on [JIRA](https://jira.stsci.edu/browse/RCAL-178?focusedCommentId=544243#comment-544243):_

Work on this ticket is blocked, as the roman_datamodels library's code needs some work for the tagged objects and metadata. (Similarly for RCAL-114: " Jump unit tests." )

PR #267, opened 2021-08-06
Implement jump step via stpipe ( RCAL-178 )

dmggh:
Closes #

Resolves RCAL-178

**Description**

This PR addresses RCAL-178 ("Implementing jump step in Roman via stpipe )

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ X ] Milestone

- [ ] Label(s)


romancal/jump/jump_step.py
```diff
@@ -5,8 +5,9 @@
 import roman_datamodels as rdm
 import numpy as np
 import time
+
 from ..stpipe import RomanStep
-from .. roman_datamodels import dqflags
+from .. datamodels import dqflags
```
nden, line 10: We need to agree on a place for dqflags, perhaps `romancal/lib`.

romancal/jump/jump_step.py
```diff
@@ -38,7 +39,7 @@ class JumpStep(RomanStep):
     def process(self, input):
 
         # Open input as a Roman DataModel (single integration; 3D arrays)
-        with rdm.RampModel(input) as input_model:
+        with rdm.datamodels.RampModel(input) as input_model:
```
nden, line 42: I don't think this works. You need to use 
```
from roman_datamodels import datamodels as rdd

with rdd.RampModel(input) as input_model:
    ...
```

nden: And open the reference files in a similar fashion.

nden (changes requested):

nden:
There are other issues with the code.


- https://github.com/spacetelescope/romancal/blob/d145251a9074f6255f22558c9a17c49916556f98/romancal/jump/jump_step.py#L54
While this creates a 4D array, there are two issues with it. It's not a float32 array and the new array is readonly and it won't work with the code in stcal. You can check this using `data.dtype` and `data.flags`.
So a better way to increase the dimension may be 
```
data = r_data[np.newaxis, :].astype(np.float32)
```
- https://github.com/spacetelescope/romancal/blob/d145251a9074f6255f22558c9a17c49916556f98/romancal/jump/jump_step.py#L112
You can pass the dqflags module here and remove `dqflags_d`.

- There's no test. As a start you can use the ramp file created by dq_init to make sure it runs through jump.

romancal/jump/jump_step.py
nden: `jwst/datamodels` are still in jwst, though there's a request to make them a standalone package. If/when this is done I suspect the module will be moved.
`romancal/datamodels` needs to be removed since the datamodels are a standalone package from the start. That's why `dqflags` needs to go in some other "common" directory. 
#262 moved `dqflags` to `lib`, import the module from there.

dmggh:
Changes for dqflags will be made in a later PR as they also involve changes to stcal.

romancal/jump/jump_step.py
```diff
@@ -85,21 +86,22 @@ def process(self, input):
             # Get the gain and readnoise reference files
             gain_filename = self.get_reference_file(input_model, 'gain')
             self.log.info('Using GAIN reference file: %s', gain_filename)
-            gain_model = rdm.GainModel(gain_filename)
+            gain_model = rdd.GainRefModel(gain_filename)
```
PaulHuwe, line 89: Note: this will not actually convert the file to a GainModel. Accordingly, this can't be relied upon to make non-gain input to work (if that is the desired effect).

romancal/jump/jump_step.py
```diff
@@ -85,21 +86,22 @@ def process(self, input):
...
             gain_2d = gain_model.data
 
             readnoise_filename = self.get_reference_file(input_model, 'readnoise')
             self.log.info('Using READNOISE reference file: %s',
                           readnoise_filename)
-            readnoise_model = rdm.ReadnoiseModel(readnoise_filename)
+            readnoise_model = rdd.ReadnoiseRefModel(readnoise_filename)
```
PaulHuwe, line 95: Note: this will not actually convert the file to a ReadnoiseModel. Accordingly, this can't be relied upon to make non-readnoise input to work (if that is the desired effect).

dmggh: How do I create models from files in these 3 cases ?

PaulHuwe: Well, there is an issue open for making the models opening other models properly in RDM: https://github.com/spacetelescope/roman_datamodels/issues/22
So the question is.. should we leave this code as is, with the assumption that that issue will fix things as intended [my preference], or do we need to manually cast until that issue is dealt with? @ddavis-stsci & @nden will have to make the call on that. 
If manually casting, check out lines 44-64 in dq_init_step.py.

nden: @PaulHuwe I'm not sure what you are referring to. When I open the readnoise file I get the model as I expect.

```
from roman_datamodels import datamodels as rdd 

rn = rdd.ReadnoiseRefModel('readnoise.asdf')  
rn                                   

 <roman_datamodels.datamodels.ReadnoiseRefModel at 0x7ff2a03a80d0>

rn.data
array([[3.546249 , 3.56219 ...
...
```

Same for the others.

PaulHuwe: Correct, the object is the correct class, but the underlying links/properties/etc. (to schemas, URI, taggedobjectnode, etc.) don't get changed over, so you get a hybrid object that doesn't completely work as expected. This is what I wrote in the DQ PR about it:

Simply opening the input_model as a RampModel wouldn't work right, no matter how many ways I tried it.. properties deep within the model never "changed over" (like the taggedobjectnode, uri, etc. - something Perry clued me in on), and it never added any metadata keywords required by ramp_model but not by the input (which would then make it fail on validation).

nden: This is not the context here. We have a file already written as a RampModel (the output of dq_init) and it opens fine as a RampModel . The same for the reference files. They are serialized as the respective models. Am I missing something?

PaulHuwe: Right, that was all I was getting at (that it needs to already be a RampModel to be properly cast as one with this command.. same with Readnoise/etc.).

PaulHuwe (approved):
LGTM

romancal/jump/jump_step.py
dmggh: With a fresh environment,  I don't seem to be getting the correct roman_datamodels :

In [1]: from roman_datamodels import datamodels as rdd

In [2]: rn_file =  "roman_wfi01_readnoise.asdf"

In [3]: rn = rdd.ReadnoiseRefModel(rn_file)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-19a848156a63> in <module>
----> 1 rn = rdd.ReadnoiseRefModel(rn_file)

AttributeError: module 'roman_datamodels.datamodels' has no attribute 'ReadnoiseRefModel'

In [6]: dir(rdd)
Out[6]: 
['DATAMODEL_EXTENSIONS',
 'DataModel',
 'FlatRefModel',
 'ImageModel',
 'PurePath',
 'Time',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__spec__',
 'asdf',
 'copy',
 'datetime',
 'model_registry',
 'np',
 'open',
 'os',
 'stnode',
 'sys',
 'validate']

PaulHuwe: Either the RDM in your environment is out of date, or the RDM line in RomanCAL's setup.cfg is locking you to an older version.

dmggh: Ok, I have a recent RMD version (0.5.0), and the romancal.setup.cfg has :

 < snip >
install_requires =
 < snip >
    roman_datamodels>=0.5.0
 < snip >

 . . . and now it fails for a different reason.  I am using :
/grp/roman/TEST_DATA/BUILD_0.1/DMS/roman_wfi01_readnoise.asdf 

In [1]: from roman_datamodels import datamodels as rdd

In [2]: rn_file = "roman_wfi01_readnoise.asdf"

In [3]: rn = rdd.ReadnoiseRefModel( rn_file )
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-3-df8d4452f5d4> in <module>
----> 1 rn = rdd.ReadnoiseRefModel(rn_file)

/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/roman_datamodels/datamodels.py in __init__(self, init, **kwargs)
     41                 raise ValueError(
     42                     f'ASDF file is not of the type expected. Expected {self.__class__.___name__}')
---> 43             self._instance = asdffile.tree['roman']
     44         elif isinstance(init, asdf.AsdfFile):
     45             asdffile = init

KeyError: 'roman'

In [4]: import roman_datamodels as rdm

In [5]: rdm.__version__
Out[17]: '0.5.0'

 Is there a problem with the asdf file ?

PaulHuwe: Quite likely, can you put the asdf files in a place for us to view?

nden: I copied updated files for you to central storage.

dmggh: Thanks @nden . Opening those readnoise and gain ref files now works as advertised :
In [39]: rn_file =  "readnoise.asdf"

In [40]: rn = rdd.ReadnoiseRefModel( rn_file )

In [41]: rn
Out[41]: <roman_datamodels.datamodels.ReadnoiseRefModel at 0x7fd119a592b0>

In [42]: rn.data
Out[42]: 
array([[3.546249 , 3.56219  , 3.5265708, ..., 3.457198 , 3.4957032,
        3.4854312],
       [3.4703345, 3.5205722, 3.509957 , ..., 3.5198073, 3.533454 ,
        3.5328283],
       [3.4962976, 3.5525622, 3.5132198, ..., 3.543181 , 3.5233061,
        3.5077085],
       ...,
       [3.5021415, 3.5244925, 3.5394077, ..., 3.4505248, 3.5169036,
        3.5146904],
       [3.4943545, 3.4304388, 3.5460162, ..., 3.4127467, 3.509256 ,
        3.5638077],
       [3.4554973, 3.5248353, 3.4717844, ..., 3.5402453, 3.426839 ,
        3.585414 ]], dtype=float32)

In [43]: rn.meta
Out[43]: {'author': 'Space Telescope Science Institute', 'date': <Time object: scale='utc' format='isot' value=2020-12-03T19:58:37.889>, 'description': 'Read noise reference file.', 'instrument': {'detector': 'WFI01', 'name': 'WFI'}, 'origin': 'STScI', 'pedigree': 'DUMMY', 'reftype': 'READNOISE', 'telescope': 'ROMAN', 'useafter': <Time object: scale='utc' format='isot' value=2020-01-01T00:00:00.000>}

 . . . and similarly for the gain file 'gain.asdf'

dmggh: Using my local copies of stpipe/integration.py , jump/jump_step.py, romancal/step.py, and __init__.py in jump, I find that I  can not strun the jump step. I can :
  
   strun romancal.step.FlatFieldStep -h
   . . .  which gives the expected step help (this is also the case for the DQInit, and DarkCurrent steps), but

(rcal_B) sal:jump grumm$ strun   romancal.step.JumpStep -h
----------------------------------------------------------------------
ERROR PARSING CONFIGURATION:
    'romancal.step.JumpStep' is not a path to a config file or a
    Python Step class
----------------------------------------------------------------------
usage: strun [--logcfg LOGCFG] [--verbose] [--debug] [--save-parameters SAVE_PARAMETERS] [--disable-crds-steppars] cfg_file_or_class

Run an stpipe Step or Pipeline

positional arguments:
  cfg_file_or_class     The configuration file or Python class to run

optional arguments:
  --logcfg LOGCFG       The logging configuration file to load
  --verbose, -v         Turn on all logging messages
  --debug               When an exception occurs, invoke the Python debugger, pdb
  --save-parameters SAVE_PARAMETERS
                        Save step parameters to specified file.
  --disable-crds-steppars
                        Disable retrieval of step parameter references files from CRDS
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/cmdline.py", line 40, in _get_config_and_class
    step_class = utilities.import_class(utilities.resolve_step_class_alias(identifier), Step)
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/utilities.py", line 63, in import_class
    step_class = getattr(imported, class_name)
AttributeError: module 'romancal.step' has no attribute 'JumpStep'

During handling of the above exception, another exception occurred:

  File "/opt/anaconda3/envs/rcal_B/bin/strun", line 28, in <module>
    step = Step.from_cmdline(sys.argv[1:])
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/step.py", line 173, in from_cmdline
    return cmdline.step_from_cmdline(args)
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/cmdline.py", line 331, in step_from_cmdline
    just_the_step_from_cmdline(args, cls)
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/cmdline.py", line 195, in just_the_step_from_cmdline
    _get_config_and_class(known.cfg_file_or_class[0])
  File "/opt/anaconda3/envs/rcal_B/lib/python3.9/site-packages/stpipe/cmdline.py", line 42, in _get_config_and_class
    raise ValueError(
ValueError: 'romancal.step.JumpStep' is not a path to a config file or a Python Step class

   So it seems that Jump step needs to be included in another file (besides those I mentioned above) that contains the other 3 steps. @PaulHuwe : having recently added the DQInit step, do you have any suggestions ?

PaulHuwe: Your code does contain the step.. I downloaded your code, and I get the following:

```
>strun romancal.step.JumpStep  -h
usage: strun [-h] [--logcfg LOGCFG] [--verbose] [--debug] [--save-parameters SAVE_PARAMETERS]
             [--disable-crds-steppars] [--pre_hooks] [--post_hooks] [--output_file]
             [--output_dir] [--output_ext] [--output_use_model] [--output_use_index]
             [--save_results] [--skip] [--suffix] [--search_output_file] [--input_dir]
             [--rejection_threshold] [--three_group_rejection_threshold]
             [--four_group_rejection_threshold] [--maximum_cores] [--flag_4_neighbors]
             [--max_jump_to_flag_neighbors] [--min_jump_to_flag_neighbors] [--override_gain]
             [--override_readnoise]
             cfg_file_or_class [args ...]

JumpStep: Performs CR/jump detection. The 2-point difference method is applied.

positional arguments:
  cfg_file_or_class     The configuration file or Python class to run
  args                  arguments to pass to step

optional arguments:
  -h, --help            show this help message and exit
  --logcfg LOGCFG       The logging configuration file to load
<snip>
```

Maybe double check that you are running the strun in the same conda environment you have done and installed your dev work in?

dmggh: It is definitely not the same Conda environment. But when I do run in the same  conda environment I get the same error.

ddavis-stsci:
This looks like a duplicate of #258

PR #283, opened 2021-08-24
Rcal 187 soc 636.1 crds flat test

### LINKED ISSUE #273: Test SOC-636.1
_Issue [RCAL-187](https://jira.stsci.edu/browse/RCAL-187) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Write a test for SOC-636: _The DMS shall generate WFI Level 2 Calibrated Science Data Products using the most current calibration reference data available._

The following description is based on [this page](https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=ROMAN&title=Build+0.3+requirements+and+test+planning)

 

Work with INS and [David Kauffman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dkauffman) to make sure there are two FLAT files in CRDS for the same instrument configuration with two different USEAFTER dates.

The test should show there are two flat files in the current context with two different USEAFTER dates.

The pipeline should make a call to CRDS and show that the flat with the latest USEAFTER is chosen.

PaulHuwe:
Closes #273 

Resolves RCAL-187

**Description**

This PR adds the required flat file CRDS temporal selection test required by SOC-636.1. The added regression test opens a file in artifactory, calls CRDS to select the matching flat file, tests CRDS matched to the correct file (by date), runs the flat step to produce output, and finally tests that final output against a truth file in artifactory. 

Checklist
- [x] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/romancal/flatfield/tests/test_flatfield.py
+++ b/romancal/flatfield/tests/test_flatfield.py
@@ -70,8 +70,6 @@ def test_flatfield_step_interface(instrument, exptype):
 def test_crds_temporal_match(instrument, exptype):
     """Test that the basic inferface works for data requiring a FLAT reffile"""
 
-    shape = (20, 20)
-
     wfi_image = testutil.mk_level2_image(arrays=True)
     wfi_image.meta.instrument.name = instrument
     wfi_image.meta.instrument.detector = 'WFI01'
@@ -85,9 +83,9 @@ def test_crds_temporal_match(instrument, exptype):
 
     step = FlatFieldStep()
     reffile = step.get_reference_file(wfi_image_model, "flat")
-    assert ("/".join(reffile.rsplit("/", 3)[1:])  == "roman/wfi/roman_wfi_flat_0002.asdf")
+    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0057.asdf")
 
-    wfi_image_model.meta.observation.start_time = Time('2021-01-01T11:11:11.110')
-    wfi_image_model.meta.observation.end_time = Time('2021-01-01T11:33:11.110')
-    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0002.asdf")
-    assert 1==0
+    wfi_image_model.meta.observation.start_time = Time('2021-08-11T11:11:11.110')
+    wfi_image_model.meta.observation.end_time = Time('2021-08-11T11:33:11.110')
+    reffile = step.get_reference_file(wfi_image_model, "flat")
+    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0039.asdf")
--- a/romancal/flatfield/tests/test_flatfield.py
+++ b/romancal/flatfield/tests/test_flatfield.py
@@ -7,7 +7,7 @@ from roman_datamodels import stnode
 from roman_datamodels.datamodels import ImageModel, FlatRefModel
 from roman_datamodels.testing import utils as testutil
 from romancal.flatfield import FlatFieldStep
-
+from astropy.time import Time
 
 @pytest.mark.parametrize(
     "instrument, exptype",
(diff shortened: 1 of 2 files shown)


PaulHuwe:
This is in draft PR until files are moved into artifactory and CRDS test server.

romancal/flatfield/tests/test_flatfield.py
```diff
@@ -54,3 +54,38 @@ def test_flatfield_step_interface(instrument, exptype):
...
+
+    wfi_image.meta.observation.start_time = Time('2020-01-01T11:11:11.110')
+    wfi_image.meta.observation.end_time = Time('2020-01-01T11:33:11.110')
+
+    wfi_image.meta.exposure.type = exptype
+    wfi_image_model = ImageModel(wfi_image)
+
+    step = FlatFieldStep()
+    reffile = step.get_reference_file(wfi_image_model, "flat")
+    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0057.asdf")
```
eslavich, line 86: This is going to be brittle as time goes on and replacement references are submitted to the test server.  Maybe we ought to add an argument to `step.get_reference_file` that allows us to pin the context here?

romancal/flatfield/tests/test_flatfield.py
```diff
@@ -54,3 +54,38 @@ def test_flatfield_step_interface(instrument, exptype):
     assert (result.data == wfi_image.data).all()
     assert result.var_flat.shape == shape
     assert result.meta.cal_step.flat_field == 'COMPLETE'
+
+
+
+@pytest.mark.parametrize(
+    "instrument, exptype",
+    [
+        ("WFI", "WFI_IMAGE"),
```
eslavich, line 63: It won't be easy to add other parameters since the detector, optical element, start/end times, and resulting reference paths are all hard coded in the test below.

romancal/flatfield/tests/test_flatfield.py
```diff
@@ -54,3 +54,38 @@ def test_flatfield_step_interface(instrument, exptype):
...
+    wfi_image_model = ImageModel(wfi_image)
+
+    step = FlatFieldStep()
+    reffile = step.get_reference_file(wfi_image_model, "flat")
+    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0057.asdf")
+
+    wfi_image_model.meta.observation.start_time = Time('2021-08-11T11:11:11.110')
+    wfi_image_model.meta.observation.end_time = Time('2021-08-11T11:33:11.110')
+    reffile = step.get_reference_file(wfi_image_model, "flat")
+    assert ("/".join(reffile.rsplit("/", 3)[1:]) == "roman/wfi/roman_wfi_flat_0039.asdf")
```
eslavich, line 91: This path depends on how the cache is configured -- the per-instrument directory structure here is less common than the "flat" structure that does not include the instrument name in the path.  I recommend comparing the filename only, something like:

```python
assert Path(reffile).name == "roman_wfi_flat_0039.asdf"
```

romancal/flatfield/tests/test_flatfield.py
PaulHuwe: Yeah, for this test, I feel like we are stuck with some brittleness, given that we are showing that different dates pull different flats. I a thinking to show that explicitly with something like:
`assert ("/".join(ref_file_path.rsplit("/", 3)[1:])) != ("/".join(ref_file_path_b.rsplit("/", 3)[1:]))`
where ref_file_path_b is the ref_file_path for the second file. I that with this, I could remove the flatfilename checks. What do you think?

romancal/flatfield/tests/test_flatfield.py
PaulHuwe: I can move these to the parameters.. I was just preserving the pattern Perry did above.

romancal/flatfield/tests/test_flatfield.py
PaulHuwe: Okay, if we keep the explicit filename check, I can simplify this.. I only specified it to easily show that the file we matched to was roman & wfi.. but, I that could be overkill.

PR #315, opened 2021-09-24
RCAL-220 Added regresson test for Jump Detection

ddavis-stsci:
Closes # 

Resolves RCAL-220

**Description**

This PR Add simple regression test for Jump detection.

Checklist
- [x ] Tests

- [ x] Documentation

- [ x] Change log

- [ x] Milestone

- [ x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 Jump
 ----------
 
- - Added simple regression test [#]
+ - Added simple regression test [#315]
 
  - Updated code for Jump step using stcal [#309]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Jump
 ----------
 
+ - Added simple regression test [#]
+
  - Updated code for Jump step using stcal [#309]
 
 0.4.2 (2021-09-13)
--- a/romancal/regtest/test_jump_det.py
+++ b/romancal/regtest/test_jump_det.py
@@ -19,9 +19,3 @@ def test_jump_detection_step(rtdata, ignore_asdf_paths):
     rtdata.get_truth(f"truth/WFI/image/{output}")
     assert (compare_asdf(rtdata.output, rtdata.truth,
             **ignore_asdf_paths) is None)
-
-    output = "l1_0006_science_raw_dqinitstep_jumpstep.asdf"
-    rtdata.output = output
-    rtdata.get_truth(f"truth/WFI/image/{output}")
-    assert (compare_asdf(rtdata.output, rtdata.truth,
-            **ignore_asdf_paths) is None)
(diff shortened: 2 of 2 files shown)


ddavis-stsci:
I hope the 20min run time is only over my VPN...

pytest --bigdata ~/src/Roman/romancal/romancal/regtest/ --basetemp=/Users/ddavis/src/Roman/develop/test/tmp_t/ -k test_jump_det
================================================== test session starts ===================================================
platform darwin -- Python 3.9.6, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /Users/ddavis/src/Roman/romancal, configfile: setup.cfg
plugins: cov-2.12.1, arraydiff-0.3, doctestplus-0.10.1, remotedata-0.3.2, xdist-2.3.0, ci-watson-0.5, filter-subpackage-0.1.1, openfiles-0.5.0, asdf-2.9.0.dev14+gee34d21, forked-1.3.0, astropy-header-0.1.2, hypothesis-6.21.3
collected 6 items / 5 deselected / 1 selected

../../romancal/romancal/regtest/test_jump_det.py .                                                                 [100%]

==================================================== warnings summary ====================================================
romancal/regtest/test_jump_det.py::test_jump_detection_step
  /Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.9/site-packages/stcal/jump/twopoint_difference.py:142: RuntimeWarning: invalid value encountered in sqrt
    sigma = np.sqrt(np.abs(median_diffs) + read_noise_2 / nframes)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
================================ 1 passed, 5 deselected, 1 warning in 1214.80s (0:20:14)

dmggh (commented):
Should there be ref file overrides specified ?

ddavis-stsci:
I don't think we need the overrides for the basic test. I'd like them to be included eventually but since the testing is already about 20min (I think because of the downloading large files on my network) I wanted to keep the time to test the routine reasonable.

dmggh:
I'm confused. Without the overrides, won't ref files be retrieved from CRDS that may differ from earlier reg tests, potentially resulting in different jump output ?  And aren't lines 17-21 identical to lines 23-27  in test_jump_det.py ?

ddavis-stsci:
> I'm confused. Without the overrides, won't ref files be retrieved from CRDS that may differ from earlier reg tests, potentially resulting in different jump output ? And aren't lines 17-21 identical to lines 23-27 in test_jump_det.py ?

I think that is what we want for the regression tests. For JWST the regression tests use the call to CRDS to select the current reference files and test the system end to end. I recall updating many truth files after reference file updates. Looking in test_nircam_image.py I don't see any overrides for CRDS. 

I think for the unit tests we want the tests to be independent of CRDS. Let me know if you think JWST does it differently and I'll look more closely and check with James or Howard.

And yes I've learned how to stutter in my editor, I've removed the duplicates.

dmggh:
Yes, it is important to include CRDS retreival in the reg tests. I too have also updated many truth files for JWST reg tests after ref file updates. Just checking.

dmggh (approved):
Looks good to me.

PR #339, opened 2021-11-02
fix for cov

nden:
**Description**
Attempt to fix testing with coverage.


nden:
I think this now fixes the last problem in the tox file. I'm going to merge in order to rebase and test it.

PR #348, opened 2021-11-09
Rcal 218 saturation

### LINKED ISSUE #312: Impliment Saturation correction using STCAL, RDM & stpipe
_Issue [RCAL-218](https://jira.stsci.edu/browse/RCAL-218) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Implement saturation correction using `stcal`, `roman_datamodels` and `romancal.stpipe`

PaulHuwe:
Closes #312 

Resolves RCAL-218

**Description**

This PR addresses implementation of saturation correction using stcal, roman_datamodels and romancal.stpipe

Checklist
- [x ] Tests

- [x ] Documentation. (code comments, full docs in a separate PR) 

- [x ] Change log

- [x ] Milestone

- [x ] Label(s)

### DIFF
--- a/romancal/regtest/test_wfi_saturation.py
+++ b/romancal/regtest/test_wfi_saturation.py
@@ -1,0 +1,75 @@
+import os
+import pytest
+
+from romancal.stpipe import RomanStep
+from romancal.step import SaturationStep
+import roman_datamodels as rdm
+from .regtestdata import compare_asdf
+
+@pytest.mark.bigdata
+def test_saturation_image_step(rtdata, ignore_asdf_paths):
+    # Testing retrieval of best ref file for image data,
+    # and creation of a ramp file with CRDS selected saturation file applied.
+
+    rtdata.get_data("WFI/image/ramp.asdf")
+    rtdata.input = "ramp.asdf"
+
+    # Test CRDS
+    step = SaturationStep()
+    model = rdm.open(rtdata.input)
+    ref_file_path = step.get_reference_file(model, "saturation")
+    ref_file_name = os.path.split(ref_file_path)[-1]
+
+    assert "roman_wfi_saturation" in ref_file_name
+
+    # Test DQInitStep
+    output = "ramp_saturationstep.asdf"
+    rtdata.output = output
+
+    args = ["romancal.step.SaturationStep", rtdata.input]
+    RomanStep.from_cmdline(args)
+
+    ramp_out = rdm.open(rtdata.output)
+    assert ("roman.pixeldq" in ramp_out.to_flat_dict())
+
+    rtdata.get_truth(f"truth/WFI/image/{output}")
+    assert (compare_asdf(rtdata.output, rtdata.truth, **ignore_asdf_paths) is None)
+
+#
+# @pytest.mark.bigdata
+# def test_saturation_grism_step(rtdata, ignore_asdf_paths):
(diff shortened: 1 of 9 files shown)


romancal/saturation/tests/test_saturation.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Also set an existing DQ flag in input science data
+    data.pixeldq[5, 5] = dqflags.pixel['DO_NOT_USE']
+
+    # Run the pipeline
+    output = flag_saturation(data, satmap)
+
+    # Make sure output GROUPDQ does not get flagged as saturated
+    # Make sure PIXELDQ is set to NO_SAT_CHECK and original flag
+    assert np.all(output.groupdq[:, 5, 5] != dqflags.group['SATURATED'])
```
dmggh, line 225: Perhaps change this first assert to check that the 'SATURATED' bit is not set, as 'DO_NOT_USE' could conceivably also be set.

romancal/saturation/tests/test_saturation.py
```diff
@@ -0,0 +1,274 @@
...
+    data.data[4, 5, 5] = 62000
+
+    # Set saturation value for pixel to NaN
+    satmap.data[5, 5] = np.nan
+
+    # Run the pipeline
+    output = flag_saturation(data, satmap)
+
+    # Check that output GROUPDQ is not flagged as saturated
+    assert np.all(output.groupdq[:, 5, 5] != dqflags.group['SATURATED'])
```
dmggh, line 256: Just curious - what would the groupdq be set to (if anything) in this case ? (I suppose it doesn't matter since pdq is set to 'NO_SAT_CHECK' )

romancal/regtest/test_wfi_saturation.py
```diff
@@ -0,0 +1,66 @@
...
+    rtdata.get_data("WFI/image/ramp.asdf")
+    rtdata.input = "ramp.asdf"
+
+    # Test CRDS
+    step = SaturationStep()
+    model = rdm.open(rtdata.input)
+    ref_file_path = step.get_reference_file(model, "saturation")
+    ref_file_name = os.path.split(ref_file_path)[-1]
+
+    assert "roman_wfi_saturation" in ref_file_name
```
dmggh, line 23: I don't quite understand this : doesn't this merely verify that *a* "roman_wfi_saturation" file (perhaps the only one currently in CRDS ? ) was retrieved ? If so an additional comment would help.

dmggh (changes requested):

ddavis-stsci (approved):
Other than what Dave G found it looks good.

romancal/saturation/tests/test_saturation.py
PaulHuwe: Added explicit check for saturation bit being NOT set.

romancal/saturation/tests/test_saturation.py
PaulHuwe: Nothing? The values set here are below the threshold, so they are all valid.

romancal/regtest/test_wfi_saturation.py
PaulHuwe: Added comments.

dmggh (approved):

PR #360, opened 2021-11-19
First version of implementation of linearity correction for RCAL-217

dmggh:
Closes #

Resolves RCAL-217

**Description**

This PR addresses RCAL-217; "Implement Linearity correction using STCAL, RDM & stpipe". This has only been tested using an override reference file. I will submit separate PRs for the documentation and tests later. 

Checklist
- [ ] Tests

- [ ] Documentation

- [X ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/romancal/linearity/linearity_step.py
+++ b/romancal/linearity/linearity_step.py
@@ -1,0 +1,64 @@
+"""
+Apply linearity correction to a science image
+"""
+
+from roman_datamodels import datamodels as rdd
+import numpy as np
+
+from romancal.stpipe import RomanStep
+from romancal.lib import dqflags
+from stcal.linearity.linearity import linearity_correction
+
+__all__ = ["LinearityStep"]
+
+
+class LinearityStep(RomanStep):
+    """
+    LinearityStep: This step performs a correction for non-linear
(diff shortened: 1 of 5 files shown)


nden:
I tested this locally, it works. I suggest adding adding a change log and merging it. Then add tests and documentation in a separate PR.

romancal/linearity/linearity_step.py
```diff
@@ -0,0 +1,64 @@
...
+    LinearityStep: This step performs a correction for non-linear
+    detector response, using the "classic" polynomial method.
+    """
+
+    reference_file_types = ['linearity']
+
+    def process(self, input):
+
+        # Open the input data model
+        with rdd.RampModel(input) as input_model:
```
ddavis-stsci, line 26: Can you change the input model open to 
       with rdm.open(input) as input_model:
This will allow us to pass the model and not try and open a file.

ddavis-stsci, line 26: Can you change the input model open to 
       with rdm.open(input) as input_model:
This will allow us to pass the model and not try and open a file.

ddavis-stsci (approved):

PR #366, opened 2021-11-24
Skip Flatfield For Non-Image Exposure Types

### LINKED ISSUE #363: Add a check for exposure type for the FlatField Step
_Issue [RCAL-253](https://jira.stsci.edu/browse/RCAL-253) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

The flat field step code should include a check for the exposure.type attribute and only perform the flat fielding if exposure.type="WFI_IMAGE". For the Prism and Grism data the code should set cal_step.flat_field="SKIPPED" and exit without running the flat fielding code.

PaulHuwe:
Closes #363 

Resolves RCAL-253 Add a check for exposure type for the FlatField Step

**Description**

This PR adds a check for WFI_Image in flat field step, and skips the step (with appropriate calstep update) if a different exposure type. Test of this is also added.

Checklist
- [x ] Tests

- [ ] Documentation

- [x ] Change log

- [x ] Milestone

- [ ] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ assign_wcs
 flat
 ----
 
-- Added check for WFI_Image in flat field step (skip if a different exposure type). Added test. [#366]
+- Added check in flat field step to skip spectroscopic observations. Added test. [#366]
 
 jump
 ----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,11 @@ assign_wcs
 
 - Added ``assign_wcs`` step to romancal. [#361]
 
+flat
+----
+
+- Added check for WFI_Image in flat field step (skip if a different exposure type). Added test. [#366]
+
 jump
 ----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,7 +25,7 @@ jump
 
 - Updates for unit tests to use stcal [#322]
 
-- Fix to jump_step to save the update pixel and group dq arrays [#319]
+- Fix to jump_step to save the update pixel and group dq arrays. [#319]
 
 - Updated code for ``jump`` step using ``stcal``. [#309]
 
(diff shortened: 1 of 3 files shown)


romancal/flatfield/flat_field.py
```diff
@@ -54,6 +54,11 @@ def do_flat_field(output_model, flat_model):
                     'shape as the science data')
         log.warning('Step will be skipped')
         output_model.meta.cal_step.flat_field = 'SKIPPED'
+    elif output_model.meta.exposure.type != "WFI_IMAGE":
+        # Check to see if attempt to flatten non-Image data
+        log.warning('Exposure type is '+output_model.meta.exposure.type+', which is not WFI_IMAGE')
```
ddavis-stsci, line 59: I'd suggest log.info instead of log warning since skipping the step is not an error.

ddavis-stsci (approved):
Looks good

romancal/flatfield/flat_field.py
nden: I agree. I would make it one line log message, e.g.
`log.info("Skipping flat field for spectral exposure."`

nden (approved):

dmggh (approved):
Not that it matters in a practical sense, but checking if flat fielding works might be more convincing if the flat fielding values are not all 1.0 (unlike at least 1 test here).

romancal/flatfield/flat_field.py
PaulHuwe: Good catch! Changed.

CHANGES.rst
```diff
@@ -11,6 +11,11 @@ assign_wcs
 
 - Added ``assign_wcs`` step to romancal. [#361]
 
+flat
+----
+
+- Added check for WFI_Image in flat field step (skip if a different exposure type). Added test. [#366]
```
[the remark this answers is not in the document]
PaulHuwe: Changed.

PR #381, opened 2021-12-08
Initial attempt at Image regtests.

PaulHuwe:
Closes #

Resolves RCAL-242

**Description**

This PR addresses tests to cover SOC-604. The tests cover:
In Wide Field Imaging mode, the DMS shall generate Level 2 science data products consisting of exposures corrected for instrument artifacts, including the following information for each calibrated exposure:

(a) Effective depth for each pixel
(b) Effective exposure time for each pixel
(c) World Coordinate System information.

Checklist
- [x ] Tests

- [ ] Documentation

- [x ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/romancal/regtest/test_wfi_pipeline.py
+++ b/romancal/regtest/test_wfi_pipeline.py
@@ -6,9 +6,17 @@ from romancal.pipeline.exposure_pipeline import ExposurePipeline
 import roman_datamodels as rdm
 
 from romancal.assign_wcs.assign_wcs_step import AssignWcsStep
-
+import numpy as np
+from numpy.testing import assert_allclose
 from .regtestdata import compare_asdf
 import copy
+import asdf
+
+def passfail(bool_expr):
+    if bool_expr:
+        return "Pass"
+    else:
+        return "Fail"
 
 @pytest.mark.bigdata
 def test_level2_image_processing_pipeline(rtdata, ignore_asdf_paths):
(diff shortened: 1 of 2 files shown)


romancal/regtest/test_wfi_pipeline.py
```diff
@@ -0,0 +1,144 @@
...
+        model.meta.wcsinfo.dec_ref += 10.0
+    else:
+        model.meta.wcsinfo.dec_ref -= 10.0
+
+    # Create new wcs object for the new pointing
+    model = AssignWcsStep.call(model)
+
+    pipeline.log.info('DMS89 MSG: Testing that the different pointings create differing wcs.......' +
+                  str(model.meta.wcs.footprint != orig_wcs.footprint))
+    assert model.meta.wcs.footprint != orig_wcs.footprint
```
nden, line 144: `footprint` is a function, not a property, so you need to call it.
It returns a  numpy array, so perhaps a test here can be checking that the difference in the footprint is ~10 deg, something like

```
footprint1 = model.meta.wcs.footprint()
footprint2 = orig_wcs.footprint()

assert_allclose(np.abs(footprint1 - footprint2), 10)
```
You will need to specify an appropriate `atol`  difference to the assert call.

romancal/regtest/test_wfi_pipeline.py
```diff
@@ -0,0 +1,144 @@
...
+                  str("exposure_time" in model.meta.exposure))
+    assert "exposure_time" in model.meta.exposure
+
+    # DMS89 WCS tests
+    pipeline.log.info('DMS89 MSG: Testing that the wcs bounding'
+                  'box was generated.......' +
+                  str(((len(model.meta.wcs.bounding_box) == 2) and
+                        (type(model.meta.wcs.bounding_box[0][0]) == float))))
+    assert ((len(model.meta.wcs.bounding_box) == 2) and
+            (type(model.meta.wcs.bounding_box[0][0]) == float))
```
nden, line 120: If checking for the type is not in the requirements I would skip it, as there's no reason this should be float, though it usually is. If we need to check the type, I would check it is a numerical type or a scalar.

nden (commented):
Comments inline

romancal/regtest/test_wfi_pipeline.py
PaulHuwe: Removed.

romancal/regtest/test_wfi_pipeline.py
PaulHuwe: I changed it to a test of RA & DEC of a center pixel (2048, 2048).

romancal/regtest/test_wfi_pipeline.py
```diff
@@ -0,0 +1,157 @@
...
+
+@pytest.mark.bigdata
+@pytest.mark.soctests
+def test_level2_image_processing_pipeline(rtdata, ignore_asdf_paths):
+    rtdata.get_data("WFI/image/l1_0001.asdf")
+    rtdata.input = "l1_0001.asdf"
+
+    # Test Pipeline
+    output = "l1_0001_cal.asdf"
+    rtdata.output = output
```
ddavis-stsci, line 26: Is there a reason you're not using an input file from INS?

ddavis-stsci (approved):
Looks good.

romancal/regtest/test_wfi_pipeline.py
```diff
@@ -0,0 +1,157 @@
...
+    args = ["--disable-crds-steppars",
+            "--steps.jump.rejection_threshold=180.0",
+            "--steps.jump.three_group_rejection_threshold=185.0",
+            "roman_elp", rtdata.input,
```
ddavis-stsci, line 30: If you add 
--steps.jump.four_group_rejection_threshold=190.
This might speed it up more.

romancal/regtest/test_wfi_pipeline.py
PaulHuwe: This is from INS.. I just renamed it to make it simpler. I can revert it to the original, if you wish.

ddavis-stsci: I think that would be a good idea. I'm just thinking that when we have to update the tests when INS updates the input files that not having to remember (or reconstruct) name changes will make maintenance more straightforward.

nden (approved):
LGTM
I didn't run the test.

romancal/regtest/test_wfi_pipeline.py
PaulHuwe: Files renamed back to INS names both in the code and in artifactory.

romancal/regtest/test_wfi_pipeline.py
PaulHuwe: Added, thanks.

PR #387, opened 2021-12-13
remove hard-coded config for number of test jobs

jneu:
the number of pytest jobs is hard coded at 30; changing it to 'auto' to properly scale with the build server's resources


ddavis-stsci (approved):
LGTM

PR #392, opened 2022-01-06
RCAL-186 Add regression tests for Dark Current Step

ddavis-stsci:
Closes #392

Resolves RCAL-186

**Description**

This PR adds an initial set of dark current regression tests. The test for the option --dark_output is currently xfailed and the ticket RCAL-285 is filed to address that failure. Once this is fixed the test should be reinstated. 

Checklist
- [x] Tests

- [ ] Documentation - N/A

- [x] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/romancal/regtest/test_dark_current.py
+++ b/romancal/regtest/test_dark_current.py
@@ -1,0 +1,83 @@
+"""Regression test for the Dark current subtraction step."""
+import pytest
+
+from romancal.stpipe import RomanStep
+from .regtestdata import compare_asdf
+
+
+@pytest.mark.bigdata
+def test_dark_current_subtraction_step(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Note: This
+        should include tests for overrides etc. """
+    rtdata.get_data(
+        "WFI/image/r0000101001001001001_01101_0001_WFI01_linearity.asdf")
+    rtdata.input = "r0000101001001001001_01101_0001_WFI01_linearity.asdf"
+
+    args = ["romancal.step.DarkCurrentStep", rtdata.input]
+    RomanStep.from_cmdline(args)
+    output =\
+        "r0000101001001001001_01101_0001_WFI01_linearity_darkcurrentstep.asdf"
+    rtdata.output = output
+    rtdata.get_truth(f"truth/WFI/image/{output}")
+    assert (compare_asdf(rtdata.output, rtdata.truth,
+            **ignore_asdf_paths) is None)
+
+
+@pytest.mark.bigdata
+def test_dark_current_outfile_step(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Here the
+        test is for renaming the output file. """
+    rtdata.get_data(
+        "WFI/image/r0000101001001001001_01101_0001_WFI01_linearity.asdf")
+    rtdata.input = "r0000101001001001001_01101_0001_WFI01_linearity.asdf"
+
+    args = ["romancal.step.DarkCurrentStep", rtdata.input,
+            '--output_file=Test_dark']
+    RomanStep.from_cmdline(args)
+    output = "Test_dark_darkcurrentstep.asdf"
+    rtdata.output = output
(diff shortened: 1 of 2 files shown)


romancal/regtest/test_dark_current.py
```diff
@@ -0,0 +1,82 @@
...
+import pytest
+
+from romancal.stpipe import RomanStep
+from .regtestdata import compare_asdf
+
+
+@pytest.mark.bigdata
+def test_dark_current_subtraction_step(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Note: This
+        should include tests for overrides etc. """
```
PaulHuwe, line 11: Are the overrides in the reference / output files, or something in the code?

romancal/regtest/test_dark_current.py
```diff
@@ -0,0 +1,82 @@
...
+    rtdata.output = output
+    rtdata.get_truth(f"truth/WFI/image/{output}")
+    assert (compare_asdf(rtdata.output, rtdata.truth,
+            **ignore_asdf_paths) is None)
+
+
+@pytest.mark.bigdata
+def test_dark_current_outfile_step(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Here the
+        test if for renaming the output file. """
```
PaulHuwe, line 29: if -> is

romancal/regtest/test_dark_current.py
```diff
@@ -0,0 +1,82 @@
...
+    rtdata.output = output
+    rtdata.get_truth(f"truth/WFI/image/{output}")
+    assert (compare_asdf(rtdata.output, rtdata.truth,
+            **ignore_asdf_paths) is None)
+
+
+@pytest.mark.bigdata
+def test_dark_current_outfile_suffix(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Here the
+        test if for renaming the output file. """
```
PaulHuwe, line 47: if -> is, though I think this comment should be change to indicate that it is the suffix renaming that is being tested.

romancal/regtest/test_dark_current.py
```diff
@@ -0,0 +1,82 @@
...
+from .regtestdata import compare_asdf
+
+
+@pytest.mark.bigdata
+def test_dark_current_subtraction_step(rtdata, ignore_asdf_paths):
+    """ Function to run and compare Dark Current subtraction files. Note: This
+        should include tests for overrides etc. """
+    rtdata.get_data(
+        "WFI/image/r0019101001023012008_01101_0173_WFI01_jumpstep.asdf")
+    rtdata.input = "r0019101001023012008_01101_0173_WFI01_jumpstep.asdf"
```
PaulHuwe, line 14: Isn't dark run before jump? Shouldn't the input be from a step before dark?

ddavis-stsci: Good catch, I was also using an older input file from INS and I've updated to the latest.

romancal/regtest/test_dark_current.py
ddavis-stsci: strun allows you to override many of the defaults and we should test them. You can see some of the options with
strun --verbose  roman_elp -h

and some of the overrides,
153:  --steps.dq_init.override_mask
180:  --steps.saturation.override_saturation
207:  --steps.linearity.override_linearity
236:  --steps.dark_current.override_dark
275:  --steps.jump.override_gain
277:  --steps.jump.override_readnoise
309:  --steps.rampfit.override_readnoise
311:  --steps.rampfit.override_gain
363:  --steps.flatfield.override_flat

PaulHuwe: Okay, then the note should probably be deleted from here, since this test doesn't test overrides.. that test is the XFAIL at the bottom.

ddavis-stsci: Deleted.

PaulHuwe (approved):
Approved. I ran all the regression tests, and they all passed (or known skip & xfail).

PR #416, opened 2022-01-29
fix string interpolation in Jenkins file

nden:
**Description**

This PR addresses a potential security issue in the Jenkins file with
groovy string interpolation of variables. Groovy treats "double quoted"
strings in a slightly different way from "single quoted" strings.
A warning was raised in the Jenkins run of the regression tests.


ddavis-stsci (approved):
LGTM

PR #428, opened 2022-02-11
docs for refpix describing reference pixels in data

cshanahan1:
This PR adds a section for the reference pixel correction docs to eventually live, and adds a subsection describing the location of reference pixels in the Level 1, ramp model, and level 2 files.


docs/roman/refpix/description.rst
```diff
@@ -0,0 +1,41 @@
...
+A RampModel, which is created during the ``dq_init`` step, represents a dataset
+at any intermediate step between Level 1 and the final Level 2 image. Like the
+Level 1 file, RampModels also contain an array with both the science and border
+reference pixels, and another with the Amp 33 reference pixels. In addition to
+these arrays, there are four more arrays that contain the original border
+reference pixels (top, bottom, left, and right), and an additional four for
+their DQ arrays. The border pixels are copied during the ``dq_init``, so they
+reflect the original state of the border pixels before any calibration.
+The border pixels that are still attached to the science data in the RampModel
+are needed during the Linearity and Saturation correction, and will later be
```
nden, line 32: I would rephrase this as
`The border pixels that are still attached to the science data in the RampModel will later be discarded when the Level 2 image is created.`
They are not needed by Linearity and Saturation, we keep them so that the model is consistent.

nden (changes requested):

nden (approved):

PR #435, opened 2022-02-22
RCAL-308 Update the output of dq_init and ramp_fitting for reference pixel correction

cshanahan1:
The dq_init step copies the original 3D border reference pixels and their dq arrays into individual attributes. The ramp fitting step trims the border pixel arrays from the science data/dq/err arrays.

### DIFF
--- a/romancal/dq_init/dq_init_step.py
+++ b/romancal/dq_init/dq_init_step.py
@@ -83,6 +83,21 @@ class DQInitStep(RomanStep):
                 **reference_file_models,
             )
 
+            # copy original border reference file arrays (data and dq)
+            # to their own attributes. they will also remain attached to
+            # the science data until they are trimmed at ramp_fit
+            # these arrays include the overlap regions in the corners
+
+            output_model.border_ref_pix_right = output_model.data[:, :, -4:]
+            output_model.border_ref_pix_left = output_model.data[:, :, :4]
+            output_model.border_ref_pix_top = output_model.data[:, :4, :]
+            output_model.border_ref_pix_bottom = output_model.data[:, -4:, :]
+
+            output_model.dq_border_ref_pix_right = output_model.pixeldq[:, -4:]
+            output_model.dq_border_ref_pix_left = output_model.pixeldq[:, :4]
+            output_model.dq_border_ref_pix_top = output_model.pixeldq[:4, :]
+            output_model.dq_border_ref_pix_bottom = output_model.pixeldq[-4:, :]
+
         else:
             # Skip DQ step if no mask files
             reference_file_models['mask'] = None
@@ -91,6 +106,7 @@ class DQInitStep(RomanStep):
 
             output_model = init_model
             output_model.meta.cal_step.dq_init = 'SKIPPED'
+
 
         # Close the input and reference files
         input_model.close()
--- a/romancal/ramp_fitting/ramp_fit_step.py
+++ b/romancal/ramp_fitting/ramp_fit_step.py
@@ -55,6 +55,7 @@ def create_optional_results_model(input_model, opt_info):
     out_node = rds.RampFitOutput(inst)
     opt_model = rdd.RampFitOutputModel(out_node)
     opt_model.meta.filename = input_model.meta.filename
+
     return opt_model
 
 
@@ -68,6 +69,8 @@ def create_image_model(input_model, image_info):
         Input ``RampModel`` for which the output ``ImageModel`` is created.
     image_info : tuple
         The ramp fitting arrays needed for the ``ImageModel``.
+    refpix_info : tuple
+        The reference pixel arrays.
 
     Returns
     -------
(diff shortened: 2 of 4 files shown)


romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -91,6 +92,12 @@ def create_image_model(input_model, image_info):
             }
     out_node = rds.WfiImage(inst)
     im = rdd.ImageModel(out_node)
+
+    # trim off border reference pixels from science data, dq, err
+    im.data = im.data[:, 4:-4, 4:-4]
+    im.dq = im.dq[:, 4:-4, 4:-4]
+    im.err = im.err[:, 4:-4, 4:-4]
```
ddavis-stsci, line 99: Don't we also need to trim var_poisson, var_rnoise?

ddavis-stsci: Gack, never mind, these will change when the reference files are updated.

ddavis-stsci (approved):

PaulHuwe (changes requested):

romancal/dq_init/dq_init_step.py
```diff
@@ -83,6 +83,21 @@ def process(self, input):
...
 
+            # copy original border reference file arrays (data and dq)
+            # to their own attributes. they will also remain attached to
+            # the science data until they are trimmed at ramp_fit
+            # these arrays include the overlap regions in the corners
+
+            output_model.border_ref_pix_right = output_model.data[:, -4:]
+            output_model.border_ref_pix_left = output_model.data[:, :4]
+            output_model.border_ref_pix_top = output_model.data[:4, :]
+            output_model.border_ref_pix_bottom = output_model.data[-4:, :]
```
PaulHuwe, line 94: These need to go back to being 3D.. otherwise you get output objects like:
  border_ref_pix_top: !core/ndarray-1.0.0
    shape: [4, 4096, 4096]
    
When it should be number_of_groups, 4 , 4096

romancal/dq_init/dq_init_step.py
```diff
@@ -83,6 +83,21 @@ def process(self, input):
...
+
+            output_model.dq_border_ref_pix_right = output_model.pixeldq[:, -4:]
+            output_model.dq_border_ref_pix_left = output_model.pixeldq[:, :4]
+            output_model.dq_border_ref_pix_top = output_model.pixeldq[:4, :]
+            output_model.dq_border_ref_pix_bottom = output_model.pixeldq[-4:, :]
+
```
PaulHuwe, line 100: These 8 lines need tests.

romancal/ramp_fitting/ramp_fit_step.py
PaulHuwe, line 99: These lines need testing.

PaulHuwe (changes requested):

romancal/ramp_fitting/ramp_fit_step.py
nden: `im` here is the flattened 2D image.

romancal/dq_init/dq_init_step.py
ddavis-stsci: Correct the test_ramp_fit.py needs to be expanded to test the new code.

ddavis-stsci (changes requested):
The test_ramp_fit.py unit test needs to cover the new roman cases.

romancal/ramp_fitting/ramp_fit_step.py
cshanahan1: I added tests for these but codecov still says the lines aren't covered. Could it be because I used pytest skip if CI? the test i wrote runs locally

nden:
There's one code style issue but the rest looks good.

romancal/ramp_fitting/ramp_fit_step.py
ddavis-stsci: The CI skip is only for runs external to the institute. The roman test CRDS server is only available on vpn so it will fail for the CI runs that are not on VPN.

cshanahan1: hmm ok, i still dont understand why these lines are marked as uncovered then

romancal/dq_init/dq_init_step.py
nden: At some level these outputs are validates by the schemas., not the size but the dimension.

nden: I suggest we update the ramp_fit tests in a different PR and "unxfail"  them.

PaulHuwe: I think it is fine to XFail these and put tests in another PR. I do think we need tests beyond validation, as it is easy to accidentally grab the wrong values (or mess up their order), which the tests should cover.

romancal/ramp_fitting/ramp_fit_step.py
PaulHuwe: Did you commit the test file? This ticket only has 4 files attached to it at present:
CHANGES.rst
romancal/dq_init/dq_init_step.py 
romancal/dq_init/tests/test_dq_init.py 
romancal/ramp_fitting/ramp_fit_step.py 
There is no ramp test file added here.

cshanahan1: the test i added is in test_dq_init.py that should cover the untested lines in dq_init_step.py. I'm going to add the coverage to ramp in a separate PR since the existing test (which i wanted to just add assert statements to) is marked as xfail and needs to be fixed.

ddavis-stsci: The test files are created on-the-fly, the generate_wfi_reffiles needs to be updated.

nden (approved):
I suggest changes to the ramp_fitting test are implemented in a follow up PR so at least we have the code running for others to test locally.

ddavis-stsci (approved):

PR #479, opened 2022-04-20
RCAL-322: Tests for SOC-588 (Absolute photometric calibration)

### LINKED ISSUE #474: Tests for SOC-588 (Absolute photometric calibration)
_Issue [RCAL-322](https://jira.stsci.edu/browse/RCAL-322) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

PaulHuwe:
GitHub issue, Closes #474

Resolves [RCAL-322](https://jira.stsci.edu/browse/RCAL-322)

**Description**

This PR adds SOC test for absolute photometric calibration. 
Also, tweaked logging in photom step.
Updated rad version compatibility in setup.cfg.

Checklist
- [x] Tests

- [x] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     pyparsing>=2.2
     requests>=2.22
     roman_datamodels>=0.11.0
-    romanad==0.10.0
+    romanad>=0.11.0
```
ddavis-stsci, line 33: This should now be rad not romanad.

ddavis-stsci (approved):
How did you pick your tolerances? I'm surprised that the tolerance for the megaJy and micro Jy uncertainties are the same for an absolute tolerance. 
Overall it looks like it tests what we need.

PaulHuwe:
Files are in artifactory.

PaulHuwe:
> 

I chose the tolerances relative to the values calculated from the matching photom file. I agree that ideally the tolerances for the uncertainties would be different than for the absolute values, but given that the uncertainties are 0 in the CRDS selected photom file, there wasn't a good base to set them off of - hence being the same.

setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     pyparsing>=2.2
     requests>=2.22
     roman_datamodels>=0.11.0
-    romanad==0.10.0
+    rad>=0.11.0
```
nden, line 33: Should we remove rad altogether since it comes with romand_datamodels. Is romancal using anything directly from rad or even importing it?

nden (commented):
Tests pass now.

nden (approved):

setup.cfg
ddavis-stsci: Rad can be removed. It was there when we needed an earlier version of rad to pass the tests and is no longer needed.

PR #504, opened 2022-05-17
Rcal-340 Updates to flat field code to handle no flat for grism data & remove skip

### LINKED ISSUE #502: Review Regression tests for 22Q3_B6 release
_Issue [RCAL-340](https://jira.stsci.edu/browse/RCAL-340) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Make sure that all the DMS numbered tests are running and passing.

ddavis-stsci:
GitHub issue, Closes #502 

Resolves [RCAL-340](https://jira.stsci.edu/browse/RCAL-340)

**Description**

This PR updates to flat field code to handle the case of grism observations with no flat field reference files and removes the skip so that DMS-86, 87, 90, & 91 are covered. 
The results of the regression tests can be found at
https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-Developers-Pull-Requests/12/

Checklist
- [x] Tests

- [x] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/romancal/flatfield/flat_field_step.py
+++ b/romancal/flatfield/flat_field_step.py
@@ -2,10 +2,10 @@
 Flat-field a science image
 """
 
+from crds.core.exceptions import CrdsLookupError
 from ..stpipe import RomanStep
 from . import flat_field
 import roman_datamodels as rdm
-from crds.core.exceptions import CrdsLookupError
 
 __all__ = ["FlatFieldStep"]
 
--- a/romancal/flatfield/flat_field_step.py
+++ b/romancal/flatfield/flat_field_step.py
@@ -5,6 +5,7 @@ Flat-field a science image
 from ..stpipe import RomanStep
 from . import flat_field
 import roman_datamodels as rdm
+from crds.core.exceptions import CrdsLookupError
 
 __all__ = ["FlatFieldStep"]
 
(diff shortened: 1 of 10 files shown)


romancal/flatfield/flat_field_step.py
```diff
@@ -20,7 +21,10 @@ def process(self, step_input):
         input_model = rdm.open(step_input)
         # Get reference file paths
         reference_file_names = {}
-        reffile = self.get_reference_file(input_model, "flat")
+        try:
+            reffile = self.get_reference_file(input_model, "flat")
+        except CrdsLookupError:
+            reffile = None
```
nden, line 27: If this is to cover spectral data, I think we have two other options
- Change the flat rmap in CRDS to include `exposure.type` as a selector and return N/A for spectral types
- Look at exposure.type in the beginning of the step and if it's `grism` or `prism` skip the step.
I think `CrdsLookupError` is reserved for when a value is expected from CRDS but not returned, for example because of a wrong rmap.
I would go with the second option now but we should consider the first one too.

nden (changes requested):

nden (approved):

PR #511, opened 2022-05-31
use GitHub Actions to cache environments and use dependent CI workflow structure

zacharyburnett:
image

Notable changes:

- workflow jobs are defined in GitHub Actions to simplify the environment caching and construction
- all environments are cached to speed up subsequent runs; cache key is set to hash the build files so that if a build requirement changes the environment is built from scratch
- matrices are dependent in a workflow such that a commit that fails style checks will not run tests, and so on
- matrices test a more comprehensive set of platforms and Python versions

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [ ] Milestone
- [ ] Label(s)


zacharyburnett:
Tests in the `numpy` version matrix are failing for some reason. The differences in `pip freeze` are below:
`numpy==*`
```diff
asdf==2.11.1
...
numpy==1.22.4
...
```
`numpy==1.20.*`
```diff
asdf==2.12.0
...
numpy==1.20.3
...
```
`numpy==1.19.*`
```diff
asdf==2.12.0
...
numpy==1.19.5
...
```

zacharyburnett:
I commented-out tests for older Numpy versions (`1.19.*`, `1.20.*`, `1.21.*`) as they all fail due to dependency version specification and issues with entrypoint population. Is `romancal` expected to support these `numpy` versions?

zacharyburnett:
it looks like Numpy less than version `1.22.*` does not pass the unit tests

zacharyburnett:
`VersionConflict`s appear in the tests pre-Numpy `1.22.3`, due to the version requirement for `numpy`:
https://github.com/spacetelescope/romancal/blob/0405b72a2eb684f485d3ee6f2a549adeed4d41cd/setup.cfg#L29

```cfg
# roman-datamodels setup.cfg
numpy>=1.21
```
https://github.com/spacetelescope/roman_datamodels/blob/696ac384d82b6b221d5e7cdee27295cdfe2ee178/setup.cfg#L28

nden:
I'm not sure why we even have versions for numpy set in those. Shouldn't the appropriate numpy version be picked up automatically? Well, unless there are issues with a specific version but I'm not aware of any.

zacharyburnett:
I would imagine it would be helpful for users who might be installing in an old environment, perhaps? However all of the documentation is geared towards making a new environment from scratch. Either way, the current tests don't install older `numpy` anyway:
https://github.com/spacetelescope/romancal/runs/7240702288?check_suite_focus=true#step:7:65

nden:
We should make at least the roman software requirements consistent.

nden (approved):

PR #512, opened 2022-06-02
move build configuration into `pyproject.toml`

zacharyburnett:
**Description**

`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [ ] Milestone
- [ ] Label(s)


zacharyburnett:
copying from the `stcal` PR:

Now that Python 3.11 includes `tomli` in built-ins, and given that `romancal` doesn't currently build C extensions, I am comfortable with readying this for review. `flake8` requires its own file, but if we ever move to a different linter in the future (e.g. `ruff`) we will likely be able to use `pyproject.toml` for its configuration

If `setuptools` changes their interface with `pyproject.toml` it would involve updating the `[tool.setuptools]` table, i.e. changing the following values:
```yaml
 [tool.setuptools]
 zip-safe = true

 [tool.setuptools.packages.find]
 where = ['src']
 ```
 
 The `[project]` table  would remain unchanged as it is defined by PEP621.

ddavis-stsci (approved):
LGTM

mairanteodoro (approved):
Looking good to me.

PR #528, opened 2022-06-21
RCAL-341: Update Flat field code and tests to use N/A from CRDS

### LINKED ISSUE #505: Update Flat field code and tests to use N/A from CRDS
_Issue [RCAL-341](https://jira.stsci.edu/browse/RCAL-341) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Currently CRDS throws and error when searching for a flat field reference file for spectroscopic data. We are discussing with INS to update the rmap files for CRDS so that the value 'N/A' is returned. This will match what JWST is doing and the code is setup to handle the value of N/A. Once the CRDS changes have been make revert or update the flat field code to use the value of N/A and remove the test for CrdsLookupError in the pipeline code.

PaulHuwe:
GitHub issue, Closes #505 

Resolves [RCAL-341](https://jira.stsci.edu/browse/RCAL-341)

**Description**

This PR removed try/except condition on Flat Reference file CRDS lookup.

Checklist
- [x] Tests

- [ ] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,9 +6,13 @@ assign_wcs
 
 - Add distortion transform to assign_wcs step. [#510]
 
+flat
+----
+
+- Removed try/except condition on Flat Reference file CRDS lookup. [#528]
+  
 general
 -------
-
 
 - Update pipeline steps to define the default suffix when saving the step results [#521]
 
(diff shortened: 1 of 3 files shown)


ddavis-stsci (approved):
LGTM

PR #539, opened 2022-06-27
Rcal 295 compute s region

mairanteodoro:
Resolves [RCAL-295](https://jira.stsci.edu/browse/RCAL-295)

This PR addresses the creation of a detector's `S_REGION`. We use the same approach used by JWST to calculate the aperture's footprint from WCS. The result is saved under `meta.wcsinfo`.

A simple test was also implemented inside of `test_wcs.py` that verifies if the length and format of `S_REGION` are in accordance with the expectation. We also check for a $\le$ 10% accuracy between the `S_REGION` results and the ICRS reference coordinates for the aperture.

@ddavis-stsci to review.

_N.B._: There is a discrepancy in the footprint results that we get for our simulated images (based on the telescope pointing and WCS info available in the metadata) and the ones that we get from [this website](http://gsss.stsci.edu/webservices/footprints/webform.aspx); the results seem to be a mirror image over the $\alpha$ axis. As suggested by Nadia (@nden), since this issue does not affect the calculation of the footprint at this point, we are keeping it as it is and bringing it up in our next meeting with Tyler (@tddesjardins).

### DIFF
--- a/romancal/assign_wcs/assign_wcs_step.py
+++ b/romancal/assign_wcs/assign_wcs_step.py
@@ -170,7 +170,7 @@ def add_s_region(model):
     update_s_region_keyword(model, footprint)
 
 def update_s_region_keyword(model, footprint):
-    s_region = (f'POLYGON IRCS ' + ' '.join([str(x) for x in footprint.ravel()]) + ' ')
+    s_region = ('POLYGON IRCS ' + ' '.join([str(x) for x in footprint.ravel()]) + ' ')
     if "nan" in s_region:
         # do not update s_region if there are NaNs.
         log.info("There are NaNs in s_region, S_REGION not updated.")
--- a/romancal/assign_wcs/assign_wcs_step.py
+++ b/romancal/assign_wcs/assign_wcs_step.py
@@ -171,6 +171,7 @@ def add_s_region(model):
 
 def update_s_region_keyword(model, footprint):
     s_region = ('POLYGON IRCS ' + ' '.join([str(x) for x in footprint.ravel()]) + ' ')
+    log.info("S_REGION VALUES: {}".format(s_region))
     if "nan" in s_region:
         # do not update s_region if there are NaNs.
         log.info("There are NaNs in s_region, S_REGION not updated.")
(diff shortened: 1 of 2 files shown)


romancal/assign_wcs/tests/test_wcs.py
```diff
@@ -73,3 +72,15 @@ def test_wcs(tmpdir, distortion, step):
     ra, dec = l2_wcs.meta.wcs(-5, 3)
     assert np.isnan(ra)
     assert np.isnan(dec)
+
```
ddavis-stsci, line 75: I'd add a test for bbox = None and in the sregion has nan's in it.

ddavis-stsci (approved):
I've added a few minor points but overall it looks good.

tddesjardins:
FYI I looked into the difference between the MAST footprint overlay tool and the s_region computed from the gwcs (which uses the distortion reference files made from the SIAF), and spoke with @mairanteodoro about the difference. It's actually an issue with the MAST tool, and I'm going to notify them about it. The s_region work looks correct.

romancal/assign_wcs/tests/test_wcs.py
mairanteodoro: Added a check for `meta.bounding_box` $\ne$ `None`.
We already check for NaNs in `S_REGION` in the `assign_wcs_step.update_s_region_keyword` method.

romancal/assign_wcs/assign_wcs_step.py
```diff
@@ -133,3 +137,43 @@ def wfi_distortion(model, reference_files):
...
+        footprint[0][negative_ind] = 360 + footprint[0][negative_ind]
+
+    footprint = footprint.T
+    update_s_region_keyword(model, footprint)
+
+def update_s_region_keyword(model, footprint):
+    s_region = ('POLYGON IRCS ' + ' '.join([str(x) for x in footprint.ravel()]) + ' ')
+    if "nan" in s_region:
+        # do not update s_region if there are NaNs.
+        log.info("There are NaNs in s_region, S_REGION not updated.")
```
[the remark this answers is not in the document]
mairanteodoro: Added a log message with the entire content of `S_REGION` prior to checking for its content (before if statement block).

PR #542, opened 2022-07-06
Docs update on the distortion file.

### LINKED ISSUE #463: Documentation of assign_wcs
_Issue [RCAL-320](https://jira.stsci.edu/browse/RCAL-320) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Update romancal.assign_wcs documentation for distortion.

mairanteodoro:
Resolves [RCAL-320](https://jira.stsci.edu/browse/RCAL-320)
Closes #463 

This PR updates the docs to include information about the distortion reference file used in the `assign_wcs` step.

Since most of the info was already present in the docs, this PR should be a minor update.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ assign_wcs
 
 - Add distortion transform to assign_wcs step. [#510]
 
+Documentation
+-------------
+
+- include information about the distortion reference file used in the ``assign_wcs`` step [#542]
+
 flat
 ----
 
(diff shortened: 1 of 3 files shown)


PaulHuwe (dismissed):
The changelog needs updating.

zacharyburnett:
yes it looks like this is just a docs change and not an API change

zacharyburnett:
I see now that other releases have a dedicated `Documentation` section, so I will remove the label; @mairanteodoro since that is the case all you have to do is add an entry like the following to `CHANGES.rst` under this current release:
```rst
Documentation
-------------
- include information about the distortion reference file used in the `assign_wcs` step [#542]
```

mairanteodoro:
> Yes, it looks like this is just a docs change and not an API change, so I don't think a changelog entry is necessary. However I might be wrong about that

It's worth noting that this PR is just minor formatting of the contents already available.

PR #545, opened 2022-07-14
update `stpipe` requirement to `>=0.4.1`

zacharyburnett:
Addresses [JP-2675](https://jira.stsci.edu/browse/JP-2675)

This PR updates the `stpipe` dependency requirement to utilize the newest patch (https://github.com/spacetelescope/stpipe/pull/62).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     #roman_datamodels>=0.12.2
     roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git@main
     stcal>=0.7.2
-    stpipe==0.3.1
+    stpipe>=0.4.0
 
 
 [options.extras_require]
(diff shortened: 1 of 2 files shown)


nden:
@zacharyburnett We can't update stpipe for now because of spacetelescope/stpipe#38 which uses JWST specific code. This only affects tests of the entire pipeline and they are all in the regression test system, that's why tests here pass. We'll update after a fix in stpipe (hopefully soon).

setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
     #roman_datamodels>=0.12.2
     roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git@main
     stcal>=0.7.2
-    stpipe==0.3.1
+    stpipe>=0.4.1
```
nden, line 35: Now that stpipe was released we should update this to >=0.4.2 <1

zacharyburnett: done in 918904e

nden (approved):

PR #551, opened 2022-07-27
use `tox` and set up CRDS caching when server is accessible

zacharyburnett:
This PR turns the existing dependent-workflow structure back into `tox`, and caches CRDS data for the same operational context, so it is not downloaded every time. This functionality is not currently useful (it is skipped currently) but will become useful when Roman CRDS becomes publicly available.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/.github/workflows/roman_ci.yml
+++ b/.github/workflows/roman_ci.yml
@@ -52,7 +52,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        toxenv: [ test ]
+        toxenv: [ test-xdist ]
         python-version: [ '3.x' ]
         os: [ ubuntu-latest ]
     steps:
@@ -81,7 +81,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        toxenv: [ test-cov, test-devdeps, test-numpy120, test-numpy121, test-numpy122 ]
+        toxenv: [ test-cov, test-devdeps, test-numpy120-xdist, test-numpy121-xdist, test-numpy122-xdist ]
         python-version: [ '3.8', '3.9', '3.10', '3.11' ]
         os: [ ubuntu-latest, macos-latest ]
         exclude:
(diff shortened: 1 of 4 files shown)


zacharyburnett:
this should be a draft until the Roman CRDS test server is available outside the VPN (and is used in tests)

zacharyburnett:
commit 456a3c64ad180c974c95d3406bffd621a867bc39 skips CRDS caching if the context retrieval fails, so as soon as Roman CRDS are publicly available it should start caching

ddavis-stsci (commented):
I think that having the crds sync setup is a good idea but we might want to be a bit careful here. Doing a full crds sync can take a while (several hours?). 
We might want to be more targeted and allow sync to a specific context or contexts, e.g. 
crds sync  --contexts roman_0001.pmap roman_0002.pmap

zacharyburnett:
Ok, right now it's using the result of `--operational-context`, but is there a set list of contexts that should be used for the tests specifically? https://github.com/spacetelescope/romancal/blob/456a3c64ad180c974c95d3406bffd621a867bc39/.github/workflows/roman_ci.yml#L94-L101

ddavis-stsci:
I think for now the operational contect is fine but we may want more 
flexibility at some point. So for instance if we wanted to test with an 
early
context it seems overkill to have the CRDS server have to go and mark an 
old version as "operational"

The context needed may depend on the test being run. That is why I was 
thinking it might be good if we could specify the context
or override the defaults somehow.

zacharyburnett:
> That is why I was thinking it might be good if we could specify the context or override the defaults somehow.

Ok, I added a `workflow_dispatch:` input with 6b1f27dcd672cec876ab4792703a1755f0df34f1
https://github.com/spacetelescope/romancal/blob/6b1f27dcd672cec876ab4792703a1755f0df34f1/.github/workflows/roman_ci.yml#L17-L22

nden (approved):
This looks good to me. 
@ddavis-stsci Are all your concerns addressed?

The PR needs a rebase.

ddavis-stsci (approved):
Yes I'm good with the changes.

zacharyburnett:
doing some updates to match the JWST workflow; I'll merge when tests pass

PR #553, opened 2022-07-29
fix input file

cshanahan1:
The variable `input_filename` was only being assigned if the input to ExposurePipeline was a string, but it was being checked so inputting a model wasn't working.

### DIFF
--- a/romancal/pipeline/exposure_pipeline.py
+++ b/romancal/pipeline/exposure_pipeline.py
@@ -62,6 +62,8 @@ class ExposurePipeline(RomanPipeline):
         log.info('Starting Roman exposure calibration pipeline ...')
         if isinstance(input, str):
             input_filename = basename(input)
+        else:
+            input_filename = None
 
         # open the input file
         input = rdd.open(input)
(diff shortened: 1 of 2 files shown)


romancal/pipeline/exposure_pipeline.py
```diff
@@ -62,6 +62,8 @@ def process(self, input):
         log.info('Starting Roman exposure calibration pipeline ...')
         if isinstance(input, str):
             input_filename = basename(input)
+        else:
+            input_filename = None
```
ddavis-stsci, line 66: I'm not sure that is what we want here. That will later set result.meta.filename to None. I think we want
`input_filename = input.meta.filename `
in the case we are passing a model.

cshanahan1: If passing a model, input.meta.filename is already set

cshanahan1: And I think its only re-set if its not None, on line 75

ddavis-stsci: Oops my error, you are correct.

ddavis-stsci (approved):

PR #564, opened 2022-08-26
RCAL-351 Update flat field documentation

### LINKED ISSUE #525: Flat fielding documentation is unclear?
_Issue [RCAL-351](https://jira.stsci.edu/browse/RCAL-351) was created on JIRA by [Eddie Schlafly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eschlafly):_

The documentation of the flat fielding is unclear to me.

<https://roman-pipeline.readthedocs.io/en/latest/roman/flatfield/main.html#error-propagation>

The first three equations are probably okay but they look weird in that they are all of the form a = a / b; something like a' = a / b would probably be more legible.

I was surprised that the flat field image was called SCI_FLAT, and might have guessed just FLAT (no SCI) would better distinguish it from a science image.

The 4th equation is ambiguous.  I get that V{~}f{~} = I{~}r{~}{^}2{^} * E{~}f{~}{^}2{^} / F{^}4{^}, where F is the flat, E_f is its uncertainty, and I_r is the "raw" science image (i.e., before flattening).  This is probably exactly the same as what is listed on the web page, but it's confusing because of the earlier issue about these quantities being redefined.  i.e., I = I / F.

ddavis-stsci:
Resolves [RCAL-351](https://jira.stsci.edu/browse/rcal-351)

Closes #525 

This PR updates the documentation for the error propagation in the flat field step. 

**Checklist**
- [N/A ] added entry in `CHANGES.rst` under the corresponding subsection
- [N/A ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/docs/roman/flatfield/main.rst
+++ b/docs/roman/flatfield/main.rst
@@ -32,8 +32,13 @@ are divided by the square of the flat-field value for each pixel.
 A flat-field variance array, VAR_FLAT, is created from the science exposure
 and flat-field reference file data using the following formula:
 
+ The flat-field is applied to the science data, in-place, according to:
+
 .. math::
-   SCI_{science} = SCI_{science} / SCI_{flat}
+   SCI^{\prime}_{science} = SCI_{science} / SCI_{flat}
+
+The flat-field data is also applied to the VAR_POISSON and VAR_RNOISE
+variance arrays,
 
 .. math::
    VAR\_POISSON_{science} = VAR\_POISSON_{science} / SCI_{flat}^2
@@ -41,8 +46,13 @@ and flat-field reference file data using the following formula:
 .. math::
    VAR\_RNOISE_{science} = VAR\_RNOISE_{science} / SCI_{flat}^2
 
+The variance for the flat-fielded data associated with the science
+data is determined using,
+
 .. math::
-   VAR\_FLAT_{science} = ( SCI_{science}^{2} / SCI_{flat}^{2} ) * ERR_{flat}^{2}
+   VAR\_FLAT_{science} = ( (SCI^{\prime}_{science})^{2} / SCI_{flat}^{2} ) * ERR_{flat}^{2}
+
+and finally the error that is associated with the science data is given by,
 
 .. math::
    ERR_{science} = \sqrt{VAR\_POISSON + VAR\_RNOISE + VAR\_FLAT}


hbushouse (approved):
Looks correct to me. Verified that it is in fact the *corrected* SCI_science data that are used to compute VAR_FLAT.

hbushouse:
If you don't want/need a change log entry for this, then you need to add the "no_changelog" (or whatever the exact name is) label so that the CI test doesn't fail.

schlafly:
This PR looks correct to me, but some words about the documentation...
- Do we think that SCI_{flat} is the right way to describe the flat field?  I feel like just calling it FLAT would be more informative.
- SCI_{flat} really sounds like it should be a flattened science image.
- SCI_{science} seems like a weird name if we don't have any other subscripts on SCI, except for SCI_{flat}, which doesn't seem to me to have anything to do with "SCI."
- Maybe we just want to delete all of the subscripts at present, since they don't currently distinguish much useful and are dropped in line 5?  Or keep them, but use the "science" subscript to denote flattened quantities relative to unflattened quantities, so SCI_flat = SCI / FLAT, VAR_POISSON_SCIENCE = VAR_POISSON / FLAT^2, etc..

I'm happy to take this if you'd rather, given that you've confirmed that the code is in fact computing what it should be.  And feel free to tell me that I'm being overly pedantic!

ddavis-stsci:
I think the issue is that  we are referencing the arrays in the science file and there are 7 in the file. So SCI_{flat} is just to refer to the array called flat in the science file. 

Would it make more sense if SCI_{SCI} --> SCI_{data} to reflect the array name? 

The other issue is that we are piggybacking on the JWST code, and their descriptions and memos about the process. So mapping the JWST and Roman nomenclature is clearer if we try to remain close to the JWST description but seeing I've managed to confuse myself quite successfully maybe that is not really working.

hbushouse:
In this case the name in caps is referring to the actual EXTNAME value for each array in its FITS extension, e.g. SCI, ERR, VAR_FLAT, etc. and the subscript indicates which dataset type we're referring to (e.g. the science target exposure that's being corrected or the flatfield reference file), such that "SCI_flat" indicates the "SCI" data array from the flat ref file. It's just unfortunate that "SCI" in this case could be misinterpreted as referring to the data for the science exposure.

Can anyone suggest a less ambiguous and/or confusing syntax, e.g. something like "science_data", "science_err", "flat_data", "flat_err", etc.?

PaulHuwe (approved):
LGTM

PR #569, opened 2022-09-30
Rcal 394 Refactor Saturation checks and pipeline code

ddavis-stsci:
Resolves [RCAL-394](https://jira.stsci.edu/browse/rcal-394)

Closes #

This PR refactors the pipeline code and saturation checks so that the pipeline handles the logic of running the pipeline and the step code modifies/updates the data, at least as far as it is practical. I've also reduce the amount of duplicated code. 
All unit and regression tests are passing, 
https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-Developers-Pull-Requests/78/

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 general
 -------
 
-- Update pipeline steps to define logic and code needed for fully sturated data [#]
+- Update pipeline steps to define logic and code needed for fully sturated data [#569]
 -
 
 0.8.1 (2022-08-23)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 0.8.2 (unreleased)
 ==================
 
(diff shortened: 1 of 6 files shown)


PaulHuwe (approved):
LGTM. All unit and regression tests passed.

PaulHuwe:
I saw no problems with the code and it passed all the tests, but I have concerns about this change. It would mean that initially fully saturated data will progress all the way to the ramp step before exiting - including the jump step, which can be rather lengthy. The present full saturation code has a bail-out after the saturation step in order to avoid running lengthy steps like jump pointlessly.

ddavis-stsci:
After review it was decided that the current code is fine so I'll close this issue without merging.

PR #591, opened 2022-10-21
RCAL-416: Overwrite Photometric Keywords for Spectroscopic Data

PaulHuwe:
Resolves [RCAL-416](https://jira.stsci.edu/browse/JP-nnnn)

Closes #

This PR addresses Change photom step to forcibly set the following photometric keywords to None for spectroscopic data.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


ddavis-stsci:
I think these keywords are all "null" in the schema, rad/resources/schemas/photometry-1.0.0.yaml, and that should be "None" when read into python. Is that not happening or am I not understanding yaml null's and python?

schlafly:
Just following along, you're saying you'd rather that, e.g., this line
https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/photometry-1.0.0.yaml#L13
should be
type: null
?  i.e., yaml reserved keyword that maybe gets translated to None, instead of string "null"?

ddavis-stsci:
That is correct. In the pipeline we do test to see if attributes are None, but if we have the string "null" we'll have to have a different test.

ddavis-stsci (approved):
LGTM

nden:
Needs a rebase

PR #608, opened 2022-12-06
Adding source detection step

cshanahan1:
Resolves [RCAL-423](https://jira.stsci.edu/browse/RCAL-423)

Adds the source catalog step, which uses DAOStarFinder to detect point sources and writes out and returns a catalog. 

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


schlafly:
This looks reasonable to me.  It looks like the style failure has to do with some kind of configuration error with the new pre-commit step?  And then we need to add photutils as a dependency so that the tests can run?

romancal/source_detection/source_detection_step.py
```diff
@@ -43,7 +43,7 @@ class SourceDetectionStep(RomanStep):
         roundhi = float(default=1.0)  # DAOStarFinder: Upper bound for roundness.
         peakmax = float(default=1000.0)  # Upper limit on brightest pixel in sources.
         max_sources = float(default=None)  # Max number of sources, choosing brightest.
-        scalar_threshold = float(default=None) # Detection threshold to be used 
+        scalar_threshold = float(default=None) # Detection threshold to be used
```
mairanteodoro, line 46: Which unit should we use for the threshold? Is it absolute or relative?

mairanteodoro (approved):
Looks good to me.

docs/roman/source_detection/arguments.rst
```diff
@@ -0,0 +1,55 @@
+Arguments
+=========
+The source detection fitting step has several arguments. These can be specified
+by the user by passing them to the step in a Python session, or setting them
+in a parameter file.
+
+* ``--coverage_mask``: True/False to mask pixels with DO_NOT_USE flags in their
+                       dq array. Default is True.
```
schlafly, line 8: Just out of curiosity, can you envision cases where we don't want this set to True?

docs/roman/source_detection/arguments.rst
```diff
@@ -0,0 +1,55 @@
...
+* ``--kernel_fwhm``: A parameter for DAOStarFinder: size of Gaussian kernel in
+                     pixels. Default is 2.0.
+* ``--sharplo``: A parameter for DAOStarFinder: lower bound for sharpness.
+                 Default is 0.0.
+* ``--sharphi``: A parameter for DAOStarFinder: upper bound for sharpness.
+                 Default is 1.0.
+* ``--roundlo``: A parameter for DAOStarFinder: lower bound for roundness.
+                 Default is -1.0.
+* ``--roundhi``: A parameter for DAOStarFinder: upper bound for roundness.
+                 Default is 1.0.
```
schlafly, line 18: If these parameters have natural scales or centers, it would be nice to mention them.  I imagine that 'roundness' is centered at zero (based on hi/lo); not sure what the scale is (i.e., +1 corresponds to an aspect ratio of 2 or something??).  I really don't have any intuition for sharpness, neither in scale nor center.

docs/roman/source_detection/description.rst
```diff
@@ -0,0 +1,62 @@
...
+`meta.source_catalog.tweakreg_catalog_name` attribute will be added to the file
+to point Tweakreg to the catalog on disk. To do this, set `save_catalogs` to
+True. Output catalogs will be saved in the same directory as input files, and
+are also 4D numpy arrays representing, in order, source ID, x centroid position,
+y centroid positon, and flux. Output catalogs can be in ASDF or ECSV format.
+
+NOTE: The intermediate resulting ImageModel from SourceDetectionStep can
+only be saved if it does not contain an attached catalog - to do this, use the
+`save_catalogs` option to seperate the catalog from the file and save them
+seperately.
```
schlafly, line 28: seperately -> separately

romancal/source_detection/source_detection_step.py
```diff
@@ -0,0 +1,232 @@
...
+                threshold=threshold,
+                sharplo=self.sharplo,
+                sharphi=self.sharphi,
+                roundlo=self.roundlo,
+                roundhi=self.roundhi,
+                brightest=self.max_sources,
+                peakmax=self.peakmax,
+            )
+
+            sources = daofind(self.data, mask=self.coverage_mask)
```
schlafly, line 116: My recollection was that we found this behaved better if one subtracted the sky from the data rather than built it into the threshold?

romancal/source_detection/tests/test_source_detection_step.py
```diff
@@ -0,0 +1,252 @@
...
+    with random amplitudes from `min_amp` to  `max_amp`. Assumes
+    units of e-/s."""
+
+    for i, x in enumerate(x_positions):
+        y = y_positions[i]
+        gauss = Gaussian2DKernel(2, x_size=20, y_size=20).array
+        amp = np.random.randint(200, 700)
+        arr[y - 10 : y + 10, x - 10 : x + 10] += (
+            u.Quantity(gauss, ru.electron / u.s, dtype=np.float32) * amp
+        )
```
schlafly, line 66: I think you'll have better centroid matching if you make this 21x21 instead of 20x20; even sized arrays get centered at a pixel corner rather than a pixel center.

romancal/source_detection/tests/test_source_detection_step.py
```diff
@@ -0,0 +1,252 @@
...
+    xcentroid = sorted(xcentroid)
+
+    # check that the number of input and output sources are the same
+    assert len(flux) == len(true_x)
+
+    # check that their locations agree
+    # the differences between real/actual are always around 0.5. im assuming
+    # this is because photutils uses zero-indexed pixels so the centers are
+    # (0.5, 0.5)? atol=0.2 seems to be the lowest safe value for this right now.
+
```
schlafly, line 109: We do need to get this convention correct.  The important thing is that we choose the same thing as the distortion model.  The gWCS convention is that (0, 0) is the center of the first pixel.  My guess is that the issue is that you're putting these stars in off center in add_random_guess, and that if you change the kernel size to 21x21 (i.e., something odd), this will go away.  (This has bitten me before!)

FWIW, for noise-free symmetric Gaussians, we might guess that agreement could be very good.  Your images have noise and 0.2 pix might be fine.

romancal/source_detection/tests/test_source_detection_step.py
```diff
@@ -0,0 +1,252 @@
...
+    assert np.allclose(np.abs(xcentroid - true_x), 0.5, atol=0.2)
+    assert np.allclose(np.abs(ycentroid - true_y), 0.5, atol=0.2)
```
schlafly, line 156: Same 0.5/0 comments as before.

schlafly (commented):
This looks good to me.  I added some comments in-line that may resolve the half-pixel offsets you're seeing.  Most important outstanding questions for me are:
- Do you still think there are especially fragile bits of source detection that we should be tracking?  I think I'd probably record them with comments.
- I had the impression in our testing that we found that subtracting the background and then photometering the sky-subtracted image was better performing than including the background in the threshold.  Has your view there developed?

docs/roman/source_detection/arguments.rst
cshanahan1: I can't think of any. I think my original idea was to allow someone to pass in an array of their own mask, but I'll just remove it as an option for now.

docs/roman/source_detection/arguments.rst
cshanahan1: I will mention that perfectly round == 0.0, but I also couldn't find a clear description of what sharpness means other than: 

'Nearly always, adequate limits on the sharpness and roundness criteria for real stars are 0.2 < sharp <1.0, and — 1.0 < round < +1.0' according to the DAOPhot paper.

romancal/source_detection/source_detection_step.py
cshanahan1: correct, fixed.

docs/roman/source_detection/arguments.rst
mairanteodoro: Some insights into what sharpness is:
- "[...] sharpness provides a measure of the intrinsic size of the object with respect to the PSF" ([Annibali et al. 2008](https://iopscience.iop.org/article/10.1088/0004-6256/135/5/1900))
- "The sharpness statistic compares the central pixel to the mean of the surrounding pixels. If this difference is greater than the originally estimated height of the Gaussian or less than 0.2 the height of the Gaussian (for the default values of SHARPLIM) then the star will be rejected." ([DAOPhot manual](http://astroa.physics.metu.edu.tr/MANUALS/daophot_manual/help_daophot.html))

schlafly: Thanks Clare, Mairan.  Okay, I think I read that as sharpness being (central pix - <surrounding pix>)/central pix.  In which case 1 is a really high upper bound---it's impossible to get higher without having negative pixels, though of course those will be present if there is noise.  Then conversely a perfectly flat image has sharpness = 0 everywhere.  That sounds like a reasonable definition.  There would be no natural central value in that case; it would depend on the PSF and its centering in the pixels.  For the doc, I might just say something like "typical values of sharpness range from 0 (flat) to 1 (delta function)".

romancal/source_detection/tests/test_source_detection_step.py
cshanahan1: when i changed the kernel size to odd this was resolved, i was also able to reduce the tolerance to 0.1 pixels

docs/roman/source_detection/arguments.rst
cshanahan1: thanks, added clarification.

romancal/source_detection/source_detection_step.py
cshanahan1: updated for clarification, its absolute

docs/roman/source_detection/arguments.rst
mairanteodoro: Usually, sources with sharpness close to 1 are discarded as cosmic rays.

docs/roman/source_detection/description.rst
```diff
@@ -0,0 +1,55 @@
...
+in the image.
+
+
+Outputs / Returns
+=================
+
+By default, the resulting source catalog will be temporarily attached to the
+output ImageModel in the `meta.source_catalog.tweakreg_catalog` attribute as 4D
+numpy array representing, in order, source ID, x centroid position, y centroid
+positon, and flux. This catalog will then be deleted from the model in the
```
mairanteodoro, line 15: Typo: "positon" -> "position"

docs/roman/source_detection/description.rst
```diff
@@ -0,0 +1,55 @@
...
+Tweakreg step.
+
+Optionally, the catalog can be saved to disk in which case a
+`meta.source_catalog.tweakreg_catalog_name` attribute will be added to the file
+to point Tweakreg to the catalog on disk. To do this, set `save_catalogs` to
+True. Output catalogs will be saved in the same directory as input files, and
+are also 4D numpy arrays representing, in order, source ID, x centroid position,
+y centroid positon, and flux. Output catalogs can be in ASDF or ECSV format.
```
mairanteodoro, line 23: Typo: "positon" -> "position"

docs/roman/source_detection/description.rst
```diff
@@ -0,0 +1,55 @@
...
+
+NOTE: The intermediate resulting ImageModel from SourceDetectionStep can
+only be saved if it does not contain an attached catalog - to do this, use the
+`save_catalogs` option to seperate the catalog from the file and save them
```
mairanteodoro, line 27: Typo: "seperate" -> "separate"

mairanteodoro (approved):
Found a few typos in the docs.

schlafly (commented):
Looks good to me.  I imagine we can't actually merge until the rad & roman_datamodels updates are in so we don't need to pin to your private repo.  @ddavis-stsci also mentioned regression tests, which I've never worked with.

mairanteodoro (approved):

cshanahan1:
is this ready to merge?

mairanteodoro (approved):

cshanahan1:
The only checks here failing are codecov. I did write tests that should cover every line, but they are not enabled under CI because they require a connection to roman CRDS. Is there a way to run the step and not require this connection so the tests can run? this step doesnt require a reference file.

cshanahan1:
> > The only checks here failing are codecov. I did write tests that should cover every line, but they are not enabled under CI because they require a connection to roman CRDS. Is there a way to run the step and not require this connection so the tests can run? this step doesnt require a reference file.
> 
> I think that if the test doesn't need a reference file then it should be safe to remove the`skipif` decorator.

I tried skipping the decorator but it still tries to make the connection

schlafly (approved):

cshanahan1:
oh my bad, i guess they do run.  i'll leave them unskipped then for now

any objections to merging?

mairanteodoro:
> 

If the tests pass, I'm good with merging this PR.

cshanahan1:
i'm going to merge this once checks pass

PR #611, opened 2022-12-08
consolidate CI jobs into `include:` in initial test matrix

zacharyburnett:
This PR moves redundant sections of the YAML into a single job matrix

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/roman_ci.yml
+++ b/.github/workflows/roman_ci.yml
@@ -41,7 +41,6 @@ jobs:
       - run: tox -e ${{ matrix.toxenv }}
   test:
     name: ${{ matrix.toxenv }} (Python ${{ matrix.python-version }}, ${{ matrix.os }})
-    needs: [ check ]
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


ddavis-stsci:
Do we want to have the tests based on python 3.x? I'm assuming that this will pick up the latest released python, currently 3.11. I'm asking because sometime it can take significant work to bring the software into compliance with a python upgrade and I'm not sure having the tests broken until that happens is desirable. I would think we'd leave that in the devdeps build until it is ok to bring that into the main tests.

zacharyburnett:
> Do we want to have the tests based on python 3.x? I'm assuming that this will pick up the latest released python, currently 3.11. I'm asking because sometime it can take significant work to bring the software into compliance with a python upgrade and I'm not sure having the tests broken until that happens is desirable. I would think we'd leave that in the devdeps build until it is ok to bring that into the main tests.

Good point; I added a `test-devdeps` for Python `3.x` and made the `check-` and `-pyargs` envs pin to `3.11`. Do you think they should be `3.9` / `3.10` instead?

ddavis-stsci (approved):
Look good.

PR #612, opened 2022-12-12
refactor toxenvs, use OpenAstronomy workflows, and move MacOS jobs to schedule

zacharyburnett:
This PR abstracts CI workflows to the OpenAstronomy workflows; this should reduce maintenance for updating and maintaining actions. Additionally, these changes move the majority of MacOS jobs to the weekly scheduled workflow. This should alleviate the issue where the limited MacOS runners for the organization are not available for CI jobs.

image

Also, updates Tox to 4.0 and uses the new PEP517 build system. Mainly, this requires little change to `tox.ini` because usage of `isolated_build` and `usedevelop` were already updated previously.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


.github/workflows/roman_ci_cron.yaml
```diff
@@ -0,0 +1,54 @@
+
+name: Weekly cron
+
+on:
+  schedule:
+    # Weekly Monday 6AM build
+    - cron: "0 6 * * 1"
```
zacharyburnett, line 7: move to midnight
```suggestion
    - cron: "0 0 * * 1"
```

WilliamJamieson (approved):

.github/workflows/roman_ci.yml
```diff
@@ -14,110 +14,85 @@ on:
...
+    with:
+      setenv: |
+        CRDS_PATH: ${{ needs.crds.outputs.path }}
+        CRDS_SERVER_URL: ${{ needs.crds.outputs.server }}
+        CRDS_CLIENT_RETRY_COUNT: 3
+        CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
+      cache-path: /tmp/crds_cache
+      cache-key: crds-${{ needs.crds_context.outputs.pmap }}
+      envs: |
+        - linux: test-numpy120
```
braingram, line 89: Are these necessary? If so, should 1.23 and 1.24 be added?

.github/workflows/roman_ci_cron.yaml
```diff
@@ -0,0 +1,67 @@
...
+      envs: |
+        - macos: test-xdist
+          python-version: 3.9
+        - macos: test-sdpdeps-xdist
+          python-version: 3.9
+        - macos: test-xdist
+          python-version: 3.10
+        - linux: test-pyargs-xdist
+        - linux: test-devdeps-xdist
+          python-version: 3.8
```
braingram, line 55: It looks like there was a commit to drop 3.8. https://github.com/spacetelescope/romancal/pull/612/commits/355374697f6759515c657b75804a03dfb1bac099
Should these jobs also be dropped and the minimum version bumped?

.github/workflows/roman_ci.yml
```diff
@@ -14,110 +14,85 @@ on:
...
+        CRDS_SERVER_URL: ${{ needs.crds.outputs.server }}
+        CRDS_CLIENT_RETRY_COUNT: 3
+        CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
+        DD_SERVICE: romancal
+        DD_ENV: ci
+        DD_GIT_REPOSITORY_URL: ${{ github.repositoryUrl }}
+        DD_GIT_COMMIT_SHA: ${{ github.sha }}
+        DD_GIT_BRANCH: ${{ github.ref_name }}
+      cache-path: /tmp/crds_cache
+      cache-key: crds-${{ needs.crds_context.outputs.pmap }}
```
braingram, line 64: The lack of a public CRDS server means that when run on github this creates a key `crds-`.
https://github.com/spacetelescope/romancal/actions/runs/4440790860/jobs/7795041648?pr=612#step:3:9
I'm not sure this is the cause of the `Warning: Path Validation Error` (or what that means):
https://github.com/spacetelescope/romancal/actions/runs/4440790860/jobs/7795041648?pr=612#step:21:2
I also see no cache with the key `crds-`. Are any of these issues?

braingram (approved):
Overall looks good to me.

I added comments about numpy jobs, python 3.8 jobs and the crds cache key.

.github/workflows/roman_ci.yml
WilliamJamieson: I agree that these are not really necessary.

.github/workflows/roman_ci_cron.yaml
WilliamJamieson: Right now python 3.8 is the min python: https://github.com/spacetelescope/romancal/blob/f52caa3528c5f7dbf88a80414c9c750c55915f25/pyproject.toml#L5. I don't think a python version change should be part of a refactor of CI. That's out of scope.

.github/workflows/roman_ci.yml
zacharyburnett: this fails now because of the lack of a server, but it should start to work when the server is available and CRDS is enabled in the CI tests for `romancal`

.github/workflows/roman_ci_cron.yaml
zacharyburnett: good point, I removed 3.8 tests because of an error but they seem to work now. Added back

PR #613, opened 2022-12-14
Rcal 406 science array quantities

PaulHuwe:
Resolves [RCAL-406](https://jira.stsci.edu/browse/JP-nnnn)

Closes #

This PR adds support for Quantities for data arrays.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


ddavis-stsci (approved):
LGTM

romancal/flatfield/flat_field.py
```diff
@@ -109,21 +111,21 @@ def apply_flat_field(science, flat):
     flat_data[np.where(flat_bad)] = 1.0
     # Now let's apply the correction to science data and error arrays.  Rely
     # on array broadcasting to handle the cubes
-    science.data /= flat_data
+    science.data = u.Quantity((science.data.value / flat_data), ru.electron / u.s, dtype=science.data.dtype)
```
schlafly, line 114: What units does science.data start with  We already forced it to have e/s in the schema?  flat_data is probably dimensionless, so I would have guessed that science.data /= flat_data would just work?  Am I missing something?

romancal/flatfield/flat_field.py
```diff
@@ -109,21 +111,21 @@ def apply_flat_field(science, flat):
...
+    science.var_poisson = u.Quantity((science.var_poisson.value / flat_data_squared), ru.electron**2 / u.s**2,
+                                     dtype = science.var_poisson.dtype)
+    science.var_rnoise = u.Quantity((science.var_rnoise / flat_data_squared), ru.electron**2 / u.s**2,
+                                    dtype = science.var_rnoise.dtype)
+
+    science['var_flat'] = u.Quantity((science.data.value ** 2 / flat_data_squared * flat_err ** 2),
+                                  ru.electron ** 2 / u.s ** 2, dtype=science.data.value.dtype)
+
+    err_sqrt = np.sqrt(science.var_poisson.value + science.var_rnoise.value + science.var_flat)
+    science.err = u.Quantity(err_sqrt, ru.electron / u.s, dtype=err_sqrt.dtype)
```
schlafly, line 128: Likewise, I think my expectation throughout here is that the old code would be fine
e.g. science.var_poisson /= flat_data_squared should preserve the units?
Same for science.var_rnoise, science.var_flat, and only the exception should require explicit quantities?

romancal/jump/tests/test_jump_step.py
```diff
@@ -142,7 +144,7 @@ def test_one_CR(generate_wfi_reffiles, max_cores, setup_inputs):
                           deltatime=grouptime)
 
     for i in range(ngroups):
-        model1.data[i, :, :] = deltaDN * i
+        model1.data.value[i, :, :] = deltaDN * i
```
schlafly, line 147: I would have said it's more idiomatic to change deltaDN -> deltaDN * u.DN that it is to have all the data.values.

romancal/jump/tests/test_jump_step.py
```diff
@@ -153,8 +155,8 @@ def test_one_CR(generate_wfi_reffiles, max_cores, setup_inputs):
     # Add CRs to the SCI data
     for i in range(len(CR_x_locs)):
         CR_group = next(CR_pool)
-        model1.data[CR_group:, CR_y_locs[i], CR_x_locs[i]] = \
-            model1.data[CR_group:, CR_y_locs[i], CR_x_locs[i]] + 500.
+        model1.data.value[CR_group:, CR_y_locs[i], CR_x_locs[i]] = \
+            model1.data.value[CR_group:, CR_y_locs[i], CR_x_locs[i]] + 500.
```
schlafly, line 159: Likewise my default here would have been 500 -> 500 * u.DN rather than working directly with the values.

schlafly (commented):
This looks good to me, with some comments.

My biggest concern is that the value of having units is that they prevent you from doing things that don't make sense, e.g., 1*u.DN + 1*u.electron triggers an exception.  For the most part this PR avoids this feature by working primarily with quantity.value, and just treating things as ordinary numpy arrays.  I see that as a last resort and think we should be trying to avoid ".value" and Quantity(...) when possible.

Of course, it's not always possible---stcal will want just the plan numpy arrays, and we'll need to redecorate them with units when we get them back from stcal.  But I noted a number of cases in the code where I think we could be treating the units more naturally.

PR #620, opened 2022-12-21
move scripts into subpackage to make them entry points

zacharyburnett:
This PR addresses an issue seen in the JWST pipeline where a PEP517 build does not properly install scripts packaged using the legacy `setup(scripts=...)` method for standalone scripts; this changes them to be packaged with `[project.scripts]` as modules. 

This change **should** be transparent to the user, but just in case I added `test_scripts.py` to test if scripts are accessible under the same name after installation.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


zacharyburnett:
Are these scripts meant to be run in a JWST environment? https://github.com/spacetelescope/romancal/actions/runs/3752061594/jobs/6373794155#step:5:244
```
FAILED romancal/tests/empty_config/::test_module_import[romancal.scripts.schema_editor] - ModuleNotFoundError: No module named 'jwst'
[244](https://github.com/spacetelescope/romancal/actions/runs/3752061594/jobs/6373794155#step:5:245)
FAILED romancal/tests/empty_config/::test_module_import[romancal.scripts.schemadoc] - ModuleNotFoundError: No module named 'jwst'
```
https://github.com/spacetelescope/romancal/blob/ed62a13d7501a7d2d7750919a814dd918b84691f/scripts/schema_editor#L32
https://github.com/spacetelescope/romancal/blob/ed62a13d7501a7d2d7750919a814dd918b84691f/scripts/schemadoc#L34

ddavis-stsci:
I don't think we use either of those scripts and so we can drop the references to them as well.

ddavis-stsci (approved):

PR #622, opened 2023-01-06
Start using pre-commit

WilliamJamieson:
This PR adds a [pre-commit](https://pre-commit.com/) configuration and switches the `check-style` CI to using it instead of only running `flake8`. The pre-commit includes additional useful checks. The checks added by this PR are:

- check-added-large-filesz: checks that "large" files are not added, which github might reject.
- check-ast: checks that all the python code will at least syntactically parse.
- check-case-conflict: checks that files will not conflict in case-insensitive file systems.
- check-yaml: checks that all yaml files are valid yaml.
- check-toml: checks that all toml files are valid toml.
- check-merge-conflict: checks that no merge-conflict statements have been left behind.
- check-symlinks: checks that all symlinks are valid and contained within the repository.
- debug-statements: checks that no python debugger statements are left in the code.
 - detect-private-key: checks that no private keys have been commit
- end-of-file-fixer: Properly fixes the end of files for the correct spacing (usually ending with empty line)
- trailing-whitespace: Checks that lines do not end with trailing whitespace (spaces, tabs, etc)
- python-check-blanket-noqa: checks that all `# noqa` include specific error codes to ignore.
- python-check-mock-methods: checks for common mistakes with mocks
- rst-directive-colons: checks that rst directives end with double colons
- rst-inline-touching-normal: checks that inline code code does not touch text in rst
- text-unicode-replacement-char: Forbits files with UTF-8 unicode replacement character.
- pyupgrade: automatically updates old python syntax to modern python syntax
- flake8: runs flake8 on the code
- bandit: runs bandit security checs on the code.

Note that by providing a pre-commit configuration, developers can easily configure their local version of git to run all of the above checks (which are relatively fast) on any file changes prior to committing them. The majority of the above checks are automatically fixed by pre-commit if they fail and can then be subsequently re-staged, so that the commit succeeds. Alternately pre-commit can be used to run all the above checks with one command instead of several, while automatically fixing most errors.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


mairanteodoro (commented):
Looks good to me.

I suppose most of the changes were done by the new `pre-commit` step. Am I correct, @WilliamJamieson?

I noted that whenever the string modulo operator was used in the committed code, it was replaced by `"".format()`, whereas the use of `"".format()` was replaced by f-strings. Should we standardize which string format system to use so that the `pre-commit` step uses only one kind of string formatting?

zacharyburnett:
> I noted that whenever the string modulo operator was used in the committed code, it was replaced by `"".format()`, whereas the use of `"".format()` was replaced by f-strings. Should we standardize which string format system to use so that the `pre-commit` step uses only one kind of string formatting?

That's a good point. Perhaps we should run it twice to move everything to f strings.

WilliamJamieson:
> > I noted that whenever the string modulo operator was used in the committed code, it was replaced by `"".format()`, whereas the use of `"".format()` was replaced by f-strings. Should we standardize which string format system to use so that the `pre-commit` step uses only one kind of string formatting?
> 

In general, f-strings are both faster and easier to read. However, there are some circumstances; notably, some error messages and logging messages where it is in-fact slower (though I believe this is fixed in python 3.11+). There are other cases where f-strings are less advantageous.

> That's a good point. Perhaps we should run it twice to move everything to f strings.

I don't understand what is meant by this comment. `pyupgrade` is run every time pre-commit is run, so the style job would be failing if there was anything left to move to f-strings. `pyupgrade` does not change things to f-strings in the limited set of cases where f-strings are not ideal. In the case of this PR, I have limited `pyupgrade` to only targeting changes it can make which fully work in Python 3.8, so some of the f-string conversions will not be made automatically.

WilliamJamieson:
> I suppose most of the changes were done by the new `pre-commit` step. Am I correct, @WilliamJamieson?

The vast majority of changes where made automatically by the various pre-commit tools. There are some manual changes to fix issues uncovered by some of the checks (such as adding specific `#noqa` flags rather than using blanket statements).

zacharyburnett:
> `pyupgrade` does not change things to f-strings in the limited set of cases where f-strings are not ideal. In the case of this PR, I have limited `pyupgrade` to only targeting changes it can make which fully work in Python 3.8, so some of the f-string conversions will not be made automatically.

ah ok, then this LGTM

PaulHuwe (approved):
LGTM

PR #624, opened 2023-01-12
RCAL-440 Add support for Reference File Quantities

PaulHuwe:
Resolves [RCAL-440](https://jira.stsci.edu/browse/RCAL-440])

Closes #

This PR implements support for quantities in reference files. Updated unit tests for these changes.

NOTE: pyproject.toml is temporarily pointing to dev branches to help tests pass.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/romancal/dark_current/dark_current_step.py
+++ b/romancal/dark_current/dark_current_step.py
@@ -1,6 +1,8 @@
 #! /usr/bin/env python
 
+from astropy import units as u
 import numpy as np
+
 from romancal.stpipe import RomanStep
 from roman_datamodels import datamodels as rdd
 from stcal.dark_current import dark_sub
@@ -90,9 +92,9 @@ def save_dark_data_as_dark_model(dark_data, dark_model):
 
     # Create DarkRef object and copy dark data to it
     out_dark = testutil.mk_dark(shape=dark_data.data.shape)
-    out_dark.data = dark_data.data
+    out_dark.data = u.Quantity(dark_data.data, out_dark.data.unit, dtype=out_dark.data.dtype)
     out_dark.dq = dark_data.groupdq
-    out_dark.err = dark_data.err
+    out_dark.err = u.Quantity(dark_data.err, out_dark.err.unit, dtype=out_dark.err.dtype)
 
     # Temporary patch to utilize stcal dark step until MA table support is fully implemented
     out_dark.meta.exposure['nframes'] = dark_data.exp_nframes
@@ -133,10 +135,10 @@ def dark_output_data_as_ramp_model(out_data, input_model):
     # Removing integration dimension from variables (added for stcal
     # compatibility)
     # Roman 3D
-    out_model.data = out_data.data[0]
+    out_model.data = u.Quantity(out_data.data[0], out_model.data.unit, dtype=out_model.data.dtype)
     out_model.groupdq = out_data.groupdq[0]
     # Roman 2D
     out_model.pixeldq = out_data.pixeldq
-    out_model.err = out_data.err[0]
+    out_model.err = u.Quantity(out_data.err[0], out_model.err.unit, dtype=out_model.err.dtype)
 
     return out_model
--- a/romancal/jump/jump_step.py
+++ b/romancal/jump/jump_step.py
@@ -49,10 +49,10 @@ class JumpStep(RomanStep):
 
             # Extract the needed info from the Roman Data Model
             meta = input_model.meta
-            r_data = input_model.data
+            r_data = input_model.data.value
             r_gdq = input_model.groupdq
             r_pdq = input_model.pixeldq
-            r_err = input_model.err
+            r_err = input_model.err.value
             result = input_model
 
             frames_per_group = meta.exposure.nframes
@@ -100,7 +100,7 @@ class JumpStep(RomanStep):
             gain_filename = self.get_reference_file(input_model, 'gain')
             self.log.info('Using GAIN reference file: %s', gain_filename)
             gain_model = rdd.GainRefModel(gain_filename)
-            gain_2d = gain_model.data
+            gain_2d = gain_model.data.value
 
             readnoise_filename = self.get_reference_file(input_model,
                                                          'readnoise')
@@ -108,7 +108,7 @@ class JumpStep(RomanStep):
                           readnoise_filename)
             readnoise_model = rdd.ReadnoiseRefModel(readnoise_filename)
             # This is to clear the WRITEABLE=False flag?
-            readnoise_2d = np.copy(readnoise_model.data)
+            readnoise_2d = np.copy(readnoise_model.data.value)
 
             # DG 0810/21:  leave for now; make dqflags changes in a later,
             #              separate PR
--- a/romancal/ramp_fitting/ramp_fit_step.py
+++ b/romancal/ramp_fitting/ramp_fit_step.py
@@ -2,12 +2,14 @@
 #
 import logging
 import numpy as np
+from astropy import units as u
 
 from romancal.stpipe import RomanStep
 from romancal.lib import dqflags
 from roman_datamodels import datamodels as rdd
 from roman_datamodels import stnode as rds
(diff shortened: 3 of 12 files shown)


romancal/jump/jump_step.py
```diff
@@ -49,10 +49,10 @@ def process(self, input):
 
             # Extract the needed info from the Roman Data Model
             meta = input_model.meta
-            r_data = input_model.data
+            r_data = input_model.data.value
             r_gdq = input_model.groupdq
             r_pdq = input_model.pixeldq
-            r_err = input_model.err
+            r_err = input_model.err.value
```
schlafly, line 55: It would be good form in these two lines to do something like r_data = input_model.data.to(u.DN).value, essentially making sure that the unit is the one you expect before taking the .value.  For other cases with units this is more important (e.g., you don't want to treat meters and kilometers the same here); here it would be more of a consistency check.

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -87,11 +89,11 @@ def create_image_model(input_model, image_info):
...
     inst = {'meta': meta,
-            'data': data,
+            'data': u.Quantity(data, ru.electron / u.s, dtype=data.dtype),
             'dq': dq,
-            'var_poisson': var_poisson,
-            'var_rnoise': var_rnoise,
-            'err': err,
+            'var_poisson': u.Quantity(var_poisson, ru.electron**2 / u.s**2, dtype=var_poisson.dtype),
+            'var_rnoise': u.Quantity(var_rnoise, ru.electron**2 / u.s**2, dtype=var_rnoise.dtype),
+            'err': u.Quantity(err, ru.electron / u.s, dtype=err.dtype),
```
schlafly, line 96: Just thinking about alternatives here... the other way to do this would be to require that image_info have units, and have the equivalent of this code in ramp_fit_step.  I prefer that slightly aesthetically, since ramp_fit_step is closer to knowing what units it's outputting.  It'd _really_ be better for stcal.ramp_fitting to be aware of units, but obviously we're avoiding that route for now.

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -101,6 +103,15 @@ def create_image_model(input_model, image_info):
...
             'dq_border_ref_pix_bottom': input_model.dq_border_ref_pix_bottom,
+            # 'amp33': u.Quantity(input_model.amp33, ru.DN, dtype=amp33.dtype),
+            # 'border_ref_pix_left': u.Quantity(input_model.border_ref_pix_left, ru.DN, dtype=amp33.dtype),
+            # 'border_ref_pix_right': u.Quantity(input_model.border_ref_pix_right, ru.DN, dtype=amp33.dtype),
+            # 'border_ref_pix_top': u.Quantity(input_model.border_ref_pix_top, ru.DN, dtype=amp33.dtype),
+            # 'border_ref_pix_bottom': u.Quantity(input_model.border_ref_pix_bottom, ru.DN, dtype=amp33.dtype),
+            # 'dq_border_ref_pix_left': input_model.dq_border_ref_pix_left,
+            # 'dq_border_ref_pix_right': input_model.dq_border_ref_pix_right,
+            # 'dq_border_ref_pix_top': input_model.dq_border_ref_pix_top,
+            # 'dq_border_ref_pix_bottom': input_model.dq_border_ref_pix_bottom,
```
schlafly, line 114: I'm not immediately following the intent here.

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -157,7 +168,7 @@ def process(self, input):
             buffsize = ramp_fit.BUFSIZE
             image_info, integ_info, opt_info, gls_opt_model = ramp_fit.ramp_fit(
                 input_model, buffsize, self.save_opt,
-                readnoise_model.data, gain_model.data, self.algorithm,
+                readnoise_model.data.value, gain_model.data.value, self.algorithm,
```
schlafly, line 171: These would be better with .to()'s, in my opinion.

romancal/saturation/saturation.py
```diff
@@ -34,15 +34,15 @@ def flag_saturation(input_model, ref_model):
...
+    data = input_model.data.value[np.newaxis, :]
 
     # Create the output model as a copy of the input
     output_model = input_model
     gdq = output_model.groupdq[np.newaxis, :]
     pdq = output_model.pixeldq[np.newaxis, :]
 
     # Copy information from saturation reference file
-    sat_thresh = ref_model.data.copy()
+    sat_thresh = ref_model.data.value
```
schlafly, line 45: There's a change in behavior here; do you want .value.copy() ?

romancal/dark_current/dark_current_step.py
```diff
@@ -90,9 +92,9 @@ def save_dark_data_as_dark_model(dark_data, dark_model):
 
     # Create DarkRef object and copy dark data to it
     out_dark = testutil.mk_dark(shape=dark_data.data.shape)
-    out_dark.data = dark_data.data
+    out_dark.data = u.Quantity(dark_data.data, out_dark.data.unit, dtype=out_dark.data.dtype)
     out_dark.dq = dark_data.groupdq
-    out_dark.err = dark_data.err
+    out_dark.err = u.Quantity(dark_data.err, out_dark.err.unit, dtype=out_dark.err.dtype)
```
schlafly, line 97: Can these changes be instead pushed to DarkCurrentStep.process(...) after dark_sub.do_correction?

romancal/dark_current/dark_current_step.py
```diff
@@ -133,10 +135,10 @@ def dark_output_data_as_ramp_model(out_data, input_model):
     # Removing integration dimension from variables (added for stcal
     # compatibility)
     # Roman 3D
-    out_model.data = out_data.data[0]
+    out_model.data = u.Quantity(out_data.data[0], out_model.data.unit, dtype=out_model.data.dtype)
     out_model.groupdq = out_data.groupdq[0]
     # Roman 2D
     out_model.pixeldq = out_data.pixeldq
-    out_model.err = out_data.err[0]
+    out_model.err = u.Quantity(out_data.err[0], out_model.err.unit, dtype=out_model.err.dtype)
```
schlafly, line 142: If these changes were after dark_sub.do_correction, these lines could be deleted?

ddavis-stsci (approved):
Looks Good

romancal/jump/jump_step.py
PaulHuwe: I like this, but want to spin it off into another issue to do throughout the code.

PaulHuwe: Put it in this issue #644

romancal/ramp_fitting/ramp_fit_step.py
PaulHuwe: Well, image_info receives the output of stcal, so requiring it to have units could be a pain. Maybe we should ponder this one more in #644 - since those lines kind of relate to unit translation?

romancal/ramp_fitting/ramp_fit_step.py
PaulHuwe: Deleted.

romancal/ramp_fitting/ramp_fit_step.py
PaulHuwe: Relegating to #644.

romancal/saturation/saturation.py
PaulHuwe: Changed.

romancal/dark_current/dark_current_step.py
PaulHuwe: No, one is dark_data and one is out_data.

romancal/dark_current/dark_current_step.py
PaulHuwe: I think it may be cleaner where it presently is (the out & dark are split up this way) - potentially moreso with changes from #644.

PR #625, opened 2023-01-19
RCAL-460 Update default CR rejections in jump_step

ddavis-stsci:
Resolves [RCAL-460](https://jira.stsci.edu/browse/rcal-460)

Closes #

This PR addresses updates the default jump CR rejection threshold to give "reasonable" run times and approximately the correct number of cosmic rays. This is a known issue with the code in common with JWST and the JWST team is looking at why the parameters are not working as expected. We will adopt their solution once available. 

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/romancal/jump/jump_step.py
+++ b/romancal/jump/jump_step.py
@@ -25,9 +25,9 @@ class JumpStep(RomanStep):
     """
 
     spec = """
-        rejection_threshold = float(default=4.0,min=0) # CR sigma rej thresh
-        three_group_rejection_threshold = float(default=6.0,min=0) # CR sigma rej thresh
-        four_group_rejection_threshold = float(default=5.0,min=0) # CR sigma rej thresh
+        rejection_threshold = float(default=180.0,min=0) # CR sigma rej thresh
+        three_group_rejection_threshold = float(default=185.0,min=0) # CR sigma rej thresh
+        four_group_rejection_threshold = float(default=190.0,min=0) # CR sigma rej thresh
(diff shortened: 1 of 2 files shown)


romancal/jump/jump_step.py
```diff
@@ -25,9 +25,9 @@ class JumpStep(RomanStep):
     """
 
     spec = """
-        rejection_threshold = float(default=4.0,min=0) # CR sigma rej thresh
-        three_group_rejection_threshold = float(default=6.0,min=0) # CR sigma rej thresh
-        four_group_rejection_threshold = float(default=5.0,min=0) # CR sigma rej thresh
+        rejection_threshold = float(default=180.0,min=0) # CR sigma rej thresh
+        three_group_rejection_threshold = float(default=185.0,min=0) # CR sigma rej thresh
+        four_group_rejection_threshold = float(default=190.0,min=0) # CR sigma rej thresh
```
mairanteodoro, line 30: @ddavis-stsci Did you run any tests when choosing these values? Just curious as to how you determined them.

mairanteodoro (approved):
Looks good to me.

romancal/jump/jump_step.py
ddavis-stsci: I just used the values from JWST as a start and just ran trial and error till things "looked" right. Nothing deep.

PaulHuwe (commented):
LGTM

PR #411, opened 2022-01-21
Rcal 292 Update regression test inputs parameters to reduce runtime

ddavis-stsci:
GitHub issue, Closes #409

Resolves RCAL-292

**Description**

This PR updates the jump regression test input parameter to reduce the runtime. The parameters are
-steps.jump.rejection_threshold=180. 
--steps.jump.three_group_rejection_threshold=185. 
--steps.jump.four_group_rejection_threshold=190.

Running the local regression tests with the updated parameters, 
pytest --bigdata ~/src/Roman/romancal/romancal/regtest/ --basetemp=/Users/ddavis/src/Roman/develop/test/tmp_tc/ \
-k jump  2>&1 | tee regression_test_Jan21_01.log
============================= test session starts ==============================
platform darwin -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/ddavis/src/Roman/romancal, configfile: pytest.ini
plugins: arraydiff-0.3, asdf-2.8.1, remotedata-0.3.2, ci-watson-0.5, mock-3.6.1, xdist-2.4.0, filter-subpackage-0.1.1, hypothesis-6.26.0, openfiles-0.5.0, forked-1.3.0, doctestplus-0.11.1, astropy-header-0.1.2, cov-3.0.0
collected 17 items / 16 deselected / 1 selected

../../../romancal/romancal/regtest/test_jump_det.py .                    [100%]

================= 1 passed, 16 deselected in 487.06s (0:08:07)

takes 8 min, without 3.7 hours, a reduction of ~28x 

Checklist
- [x] Tests

- [ ] Documentation, N/A

- [x] Change log

- [x] Milestone

- [x] Label(s)

PR #647, opened 2023-02-15
move pytest markers to pyproject.toml

braingram:
remove pytest.ini

The existence of pytest.ini (introduced in https://github.com/spacetelescope/romancal/pull/411) prevented pytest from
reading options from pyproject.toml that for one
example disabled doctests

~~I'm expecting this will fail ci as at least one doctest contains a use of jwst:~~ Edited: fixed this reference
https://github.com/spacetelescope/romancal/blob/393d60f4702489d9e1d2f8b1fa7054e678b619b2/romancal/associations/registry.py#L63

which appears as a failure in the stpipe downstream ci testing of romancal:
https://github.com/spacetelescope/stpipe/actions/runs/4161774078/jobs/7200153620#step:8:39

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/docs/roman/datamodels/datamodels_asdf.rst
+++ b/docs/roman/datamodels/datamodels_asdf.rst
@@ -144,9 +144,9 @@ asdf tree.
 
 .. code:: python
 
-    >>> from roman_datamodels import datamodels as rdm
-    >>> d_uncal = rdm.open('r0000101001001001001_01101_0001_WFI01_uncal.asdf')
-    >>> d_uncal.info()
+    >>> from roman_datamodels import datamodels as rdm  # doctest: +SKIP
+    >>> d_uncal = rdm.open('r0000101001001001001_01101_0001_WFI01_uncal.asdf')  # doctest: +SKIP
+    >>> d_uncal.info()  # doctest: +SKIP
     root (AsdfObject)
     ├─asdf_library (Software)
     │ ├─author (str): The ASDF Developers
@@ -181,7 +181,7 @@ For example, to show all top-level nodes and 5 of each's children:
 
 .. code:: python
 
-    >>> d_uncal.info(max_rows=(None,5))
+    >>> d_uncal.info(max_rows=(None,5))  # doctest: +SKIP
     root (AsdfObject)
     ├─asdf_library (Software)
     │ ├─author (str): The ASDF Developers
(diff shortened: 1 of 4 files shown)


braingram:
@ddavis-stsci pinging you as this pertains to the markers in pytest.ini. My apologies for turning on what appear to be broken tests :) Please let me know if I can help.

braingram:
I did not find the data file used in the doctests in docs/roman/datamodels/datamodels_asdf
https://github.com/spacetelescope/romancal/blob/393d60f4702489d9e1d2f8b1fa7054e678b619b2/docs/roman/datamodels/datamodels_asdf.rst?plain=1#L148
so I added a doctest skip to each line. This might be better handled as a python code block if the intention is to not have these as doctests.

ddavis-stsci:
> I did not find the data file used in the doctests in docs/roman/datamodels/datamodels_asdf
> 
> https://github.com/spacetelescope/romancal/blob/393d60f4702489d9e1d2f8b1fa7054e678b619b2/docs/roman/datamodels/datamodels_asdf.rst?plain=1#L148
> 
> so I added a doctest skip to each line. This might be better handled as a python code block if the intention is to not have these as doctests.

I'm curious about what kind of code block, right now it is in 
```
.. code:: python

    >>> from roman_datamodels import datamodels as rdm
    >>> d_uncal = rdm.open('r0000101001001001001_01101_0001_WFI01_uncal.asdf')
    >>> d_uncal.info()
```
Is there one that pyproject prefers?

ddavis-stsci (approved):
Looks good. I'll write a ticket to update the code blocks so we don't have to use the skips.

braingram:
> Is there one that pyproject prefers?

I believe the issue is that the `>>>` makes pytest see it as a doctest (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#doctest-blocks) and try to execute the contained test as code to check that the output matches the output in the text.

Unfortunately the formatting is a bit different if the `>>>` is removed or replaced with `>>` as sphinx then complains `WARNING: Could not lex literal_block as "python". Highlighting skipped.`. Someone with more experience with sphinx may know how to coax it into produced highlighted non-doctest code blocks without the skips but I don't immediately see a solution.

PR #645, opened 2023-02-01
Start using code formatters

WilliamJamieson:
Apply code formatters to `romancal`. This applies:

- `isort` for import sorting
- `black` for code formatting

I also noticed that for the most part his code respects a much tighter line length constraint than the 130 that is listed for flake8. So I used the code formatters to reduce the line length to the default for black of 88 characters.

**Checklist**
- [ ] added entry in `CHANGES.rst` under the corresponding subsection
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)

PR #650, opened 2023-02-24
use `ruff` in `pre-commit`

zacharyburnett:
a `ruff` configuration is already added to `pyproject.toml`; this PR uses `ruff` in `pre-commit` with `--fix`

`ruff` rules complement `black` from #645


zacharyburnett:
obsolete

PR #651, opened 2023-02-27
Bug fix: make sure ASDF is not closed prematurely.

mairanteodoro:
This PR addresses a bug in `TweakReg` where an ASDF object was being closed prematurely.
The object was being closed while `TweakReg` was determining the common name of the ASDF files in the input list.

**Checklist**
- [ ] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/romancal/tweakreg/tweakreg_step.py
+++ b/romancal/tweakreg/tweakreg_step.py
@@ -526,8 +526,10 @@ class TweakRegStep(RomanStep):
 def _common_name(group):
     file_names = []
     for im in group:
-        im = im if isinstance(im, datamodels.DataModel) else datamodels.open(im)
-        file_names.append(path.splitext(im.meta.filename)[0].strip('_- '))
+        if isinstance(im, datamodels.DataModel):
+            file_names.append(path.splitext(im.meta.filename)[0].strip('_- '))
+        else:
+            raise TypeError("Input must be a list of datamodels list.")
 
     cn = path.commonprefix(file_names)
     return cn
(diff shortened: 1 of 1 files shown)


romancal/tweakreg/tweakreg_step.py
```diff
@@ -528,7 +528,6 @@ def _common_name(group):
     for im in group:
         im = im if isinstance(im, datamodels.DataModel) else datamodels.open(im)
         file_names.append(path.splitext(im.meta.filename)[0].strip('_- '))
```
braingram, line 530: What about something like this instead?
```suggestion
        if isinstance(im, datamodels.DataModel):
            fn = im.meta.filename
        else:
            with datamodels.open(im) as dm:
                fn = dm.meta.filename
        file_names.append(path.splitext(fn)[0].strip('_- '))
```
This should handle both cases:
- when `im` is a data model, don't close the model
- when `im` is a filename, open, and then close the data model (using a with context)

ddavis-stsci: I like Brett's version.

braingram (approved):
Looks good to me.

Regarding the failing ci job: https://github.com/spacetelescope/romancal/actions/runs/4294214505/jobs/7482939142#step:9:42
I just had a coverage upload fail in stdatamodels so maybe something is up with the codecov website.

mairanteodoro:
> 



> Looks good to me.
> 
> Regarding the failing ci job: https://github.com/spacetelescope/romancal/actions/runs/4294214505/jobs/7482939142#step:9:42 I just had a coverage upload fail in stdatamodels so maybe something is up with the codecov website.

I just re-ran the job and it passed.

PR #653, opened 2023-03-01
Raise exception whenever tweakreg_catalog is missing.

mairanteodoro:
This PR addresses the case when the user wants to run TweakRegStep without running source detection first.

In addition to creating an output file with the results, the source detection step adds an attribute (`tweakreg_catalog`) to the `meta` indicating the filename that should be used in `TweakRegStep` for alignment with Gaia. Such attribute will always be present when the romancal pipeline is executed through the `ExposurePipeline` class.

In the case where the user wants to execute TweakRegStep separately without first running source detection, then `tweakreg_catalog` will be missing. For now, we will only raise an `AttributeError` and stop processing.

**Checklist**
- [ ] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/romancal/tweakreg/tweakreg_step.py
+++ b/romancal/tweakreg/tweakreg_step.py
@@ -141,7 +141,6 @@ class TweakRegStep(RomanStep):
 
         # Build the catalogs for input images
         for i, image_model in enumerate(images):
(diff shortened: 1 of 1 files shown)


ddavis-stsci (approved):
LGTM

PR #655, opened 2023-03-03
RCAL-467: Implement tests for Roman's TweakReg and ModelContainer.

mairanteodoro:
This PR implement unit tests for RomanCal's `TweakReg`.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


ddavis-stsci (changes requested):
I don't think this is the way to go here. In general the unit tests should be self contained and not depend on the romanisim package. The input asdf file should be generated by the roman_datamodels utilities and not an external input. 

I'd generate the input image with mk_level2_image in the mk_test module and then update the header parameters as needed. 

Custom input data files are more appropriate for the regression testing and we should try to use the input files provided by INS. If these are not working we should open a discussion with INS to fix/update the input files.

schlafly:
I don't think we expect files from INS that can do this until ~next month.

@ddavis-stsci , how do you recommend we get a WCS object for these tests?  I feel like this needs a distortion file from CRDS, which isn't nice in these tests?  I don't think mk_level2_image includes a WCS object.  One could start making it via the assign_wcs step, but that will depend on the CRDS distortion?  For the related simulator WCS tests, I made up a fake distortion.
https://github.com/spacetelescope/romanisim/blob/main/romanisim/tests/test_wcs.py#L14-L33
which I guess could be plumbed in to assign_wcs_step, though it gets a bit involved.

ddavis-stsci:
Or we can copy the attributes needed from a level 2 file with the 
distortions added.

mairanteodoro:
I created a dummy WCS for the unit tests using `maker_utils.mk_distortion()` for the distortion file.

mairanteodoro:
> I don't think this is the way to go here. In general the unit tests should be self contained and not depend on the romanisim package. The input asdf file should be generated by the roman_datamodels utilities and not an external input.
> 
> I'd generate the input image with mk_level2_image in the mk_test module and then update the header parameters as needed.
> 
> Custom input data files are more appropriate for the regression testing and we should try to use the input files provided by INS. If these are not working we should open a discussion with INS to fix/update the input files.

I deleted the ASDF file and refactored the test to use `mk_distortion` function available in `maker_utils`.

ddavis-stsci (changes requested):
I think this is a good start but the unit tests should also test the basic functionality of the module. I'd suggest adding tests for 
def compute_radius
_is_wcs_correction_small 
test for the custom catalog error

test that the catalog type must be one of the three specified and that the correct error is generated when this is not true.

test for when the input contains no image model

test for when the catalog does not contain the correct columns
and there are 25 items listed in the tweakreg_step specs some of these should be tested in the unit tests at least to make sure that the correct units are being used specifically if there are any conversions being done. 
Others involving input and output files can be left to the regression tests. 

So I just think we should have more complete coverage in the unit tests.

mairanteodoro:
> 

For comprehensive coverage, we can't use a dummy WCS.

schlafly (commented):
Looking through these tests... do any of them test the core functionality of:
- I have some initial set of (x, y) positions
- I have some initial WCS that is wrong (e.g., shifted)
- I have some Gaia (ra, dec)
- I call tweakreg and get a new WCS
- The new WCS is much closer to right?

I missed that test if it's in there!

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -0,0 +1,428 @@
...
+        "269.02457173246125 65.69201346248587 "
+        "269.0333096074621 65.56870823657276 "
+    )
+
+
+def _create_tel2sky_model(input_dm):
+    """Create the transform from telescope to sky.
+
+    The transform is defined with a reference point in a Frame
+    associated tih the telescope (V2, V3) in arcsec, the corresponding
```
ddavis-stsci, line 43: tih the ??

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -0,0 +1,428 @@
...
+        | SphericalToCartesian(wrap_lon_at=180)
+        | rot
+        | CartesianToSpherical(wrap_lon_at=360)
+    )
+    model.name = "v23tosky"
+    return model
+
+
+def create_wcs(input_dm, shift_1=0, shift_2=0):
+    # create a WCS object
```
ddavis-stsci, line 70: Add a doc string here with units for the shifts, I assume pixels at this point.

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Fixed.

mairanteodoro:
> 

I added `test_tweakreg_rotated_plane` to test for accuracy when the original WCS is wrong by a small amount of rotation and/or offset.

PR #664, opened 2023-03-22
run tests on minimum dependencies

zacharyburnett:
This PR adds tests for minimum dependencies using the `minimum_dependencies` package

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


pyproject.toml
```diff
@@ -12,12 +12,12 @@ classifiers = [
...
 ]
 dependencies = [
-    'asdf >=2.13.0',
+    'asdf >=2.14.2',
     'asdf-astropy >=0.4.0',
     'astropy >=5.0.4',
     'crds >=11.16.16',
     'gwcs >=0.18.1',
-    'jsonschema >=3.0.2',
+    'jsonschema >=4.0.1',
```
zacharyburnett, line 20: these were caught by the tests as being incompatible with upstream dependencies

braingram (approved):
Looks good to me. Thanks for adding the use of this tool and checking the minimum dependencies.

pyproject.toml
WilliamJamieson, line 20: Probably worth adding a changelog entry for these.

WilliamJamieson (changes requested):

PR #665, opened 2023-03-27
Bugfix for source_detection's test_outputs

WilliamJamieson:
When I ran the unit tests locally, this test:
https://github.com/spacetelescope/romancal/blob/dec39b6fc756a49cff18c1bcf2a5ad650037c1d1/romancal/source_detection/tests/test_source_detection_step.py#L159-L185

started failing the checks for the created file.

This was caused by the step not being run in the temporary directory created for the test. This left the file behind after the test finished and the test for the output file was with respect to this directory. This PR fixes this bug by using a context manager to temporarily change python's working directory to the test's temporary directory for the duration of the test.

The reason this was not caught by the CI in #608 is because this test is not run by the GitHub CI at this time due to the lack of a public CRDS server.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/source_detection/tests/test_source_detection_step.py
+++ b/romancal/source_detection/tests/test_source_detection_step.py
@@ -164,25 +164,50 @@ def test_source_detection_scalar_threshold(setup_inputs):
 def test_outputs(tmp_path, setup_inputs):
     """Make sure `save_catalogs` and `output_cat_filetype` work correctly."""
 
-    model = setup_inputs()
+    try:
+        # The contextlib.chdir() context manager was added in Python 3.11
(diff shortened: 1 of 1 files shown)


cshanahan1 (approved):

cshanahan1:
just a thought - i could add an output directory arg, currently the catalog can only be output in the pwd. i originally didn't do this because i thought it would be confusing to have this in addition to the pipeline/step output_dir argument, but i think it makes sense to have this option

ddavis-stsci (approved):
LGTM

PaulHuwe (approved):
LGTM

PR #666, opened 2023-03-27
RCAL-481: Remove use of pytest-openfiles

WilliamJamieson:
Resolves [RCAL-481](https://jira.stsci.edu/browse/RCAL-481)

`pytest-openfiles` has been deprecated and is interfering with the use of fixtures in our test code (ones that handle files). This PR switches `ResourceWarnings` into errors which will catch almost all cases of files left open. The only cases it does not catch is if an open file handle is assigned to a global variable as the "warning" will not be emitted until python stops running, which occurs after pytest has completed. 

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


mairanteodoro (approved):
Looking good to me.

PR #673, opened 2023-04-06
TweakReg: doc updates.

mairanteodoro:
Resolves [RCAL-527](https://jira.stsci.edu/browse/RCAL-527)

This PR addresses updates to the RomanCal version of TweakReg.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/docs/roman/tweakreg/README.rst
+++ b/docs/roman/tweakreg/README.rst
@@ -68,7 +68,7 @@ attributes to the catalog file names provided in either in the ASN table or
 
 Alignment
 ---------
-The source catalog (either created by `SourceDetectionStep` or user-provided)
+The source catalog (either created by `SourceDetectionStep` or provided by the user)
 gets cross-matched and fit to an astrometric reference catalog
 (set by ``TweakRegStep.abs_refcat``). The pipeline initially supports fitting to
 any Gaia Data Release (defaults to `GAIADR3`).
--- a/docs/roman/tweakreg/README.rst
+++ b/docs/roman/tweakreg/README.rst
@@ -1,0 +1,231 @@
+Description
+===========
+
+:Class: `roman.tweakreg.TweakRegStep`
+:Alias: tweakreg
+
+Overview
+--------
+This step uses the coordinates of point-like sources from an input catalog
+(i.e. the result from `SourceDetectionStep` saved in the
+`meta.tweakreg_catalog` attribute) and compares them with the
+coordinates from a Gaia catalog to compute corrections to
+the WCS of the input images such that sky catalogs obtained from the image catalogs
+using the corrected WCS will align on the sky.
+
+Custom Source Catalogs
+----------------------
+The default catalog used by ``tweakreg`` step can be disabled by
+providing a file name to a custom source catalog in the
+``meta.tweakreg_catalog`` attribute of input data models.
+The catalog must be in a format automatically recognized by
+:py:meth:`~astropy.table.Table.read`. The catalog must contain
+either ``'x'`` and ``'y'`` or ``'xcentroid'`` and ``'ycentroid'`` columns which
+indicate source *image* coordinates (in pixels). Pixel coordinates are
+0-indexed.
+
+For the ``tweakreg`` step to use user-provided input source catalogs,
+``use_custom_catalogs`` parameter of the ``tweakreg`` step must be set to
+`True`.
+
+In addition to setting the ``meta.tweakreg_catalog`` attribute of input data
+models to the custom catalog file name, the ``tweakreg_step`` also supports two
+other ways of supplying custom source catalogs to the step:
+
+1. Adding ``tweakreg_catalog`` attribute to the ``members`` of the input ASN
+   table - see `~roman.datamodels.ModelContainer` for more details.
+   Catalog file names are relative to ASN file path.
+
+2. Providing a simple two-column text file, specified via step's parameter
+   ``catfile``, that contains input data models' file names in the first column
+   and the file names of the corresponding catalogs in the second column.
+   Catalog file names are relative to ``catfile`` file path.
+
+Specifying custom source catalogs via either the input ASN table or
+``catfile``, will update input data models' ``meta.tweakreg_catalog``
+attributes to the catalog file names provided in either in the ASN table or
+``catfile``.
+
+.. note::
+    When custom source catalogs are provided via both ``catfile`` and
+    ASN table members' attributes, the ``catfile`` takes precedence and
+    catalogs specified via ASN table are ignored altogether.
+
+.. note::
+    1. Providing a data model file name in the ``catfile`` and leaving
+       the corresponding source catalog file name empty -- same as setting
+       ``'tweakreg_catalog'`` in the ASN table to an empty string ``""`` --
+       would set corresponding input data model's ``meta.tweakreg_catalog``
+       attribute to `None`. In this case, ``tweakreg_step`` will automatically
+       generate a source catalog for that data model.
+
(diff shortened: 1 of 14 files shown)


mairanteodoro:
I forgot to add the parameters to select the reference files ("Reference Files" section at the very end of the docs):
```
Reference Files
CRDS selects appropriate pars-tweakregstep references based on the following keywords.
```

docs/roman/tweakreg/README.rst
```diff
@@ -0,0 +1,293 @@
...
+  attribute. If `False`, new catalogs will be generated by the ``tweakreg``
+  step. (Default=`False`)
+
+* ``catalog_format``: A `str` indicating catalog output file format.
+  (Default='ecsv')
+
+* ``catfile``: Name of the file with a list of custom user-provided catalogs.
+  (Default='')
+
+**Optimize alignment order:**
```
schlafly, line 148: Is this variable something we use right now?  I think we've only discussed L2 alignment to Gaia; I don't know what this means in this context.

docs/roman/tweakreg/README.rst
```diff
@@ -0,0 +1,293 @@
...
+    + ``'CTL'`` (TESS Candidate Target List);
+
+    + ``'PanSTARRS'`` (DR1 and DR2);
+
+    + ``'Plato'`` (All-Sky PLATO Input Catalog DR1).
+
+  .. _note:
+  .. note::
+    When ``abs_refcat`` is `None` or an empty string, alignment to the
+    absolute astrometry catalog will be turned off.
```
schlafly, line 246: What happens in the L2 case where if abs_refcat is None but there is only one image, so no overlaps?

schlafly (approved):
Looks good.  Two minor questions about tweakreg features that make sense at L3 time but which I don't really understand at L2 time.

docs/roman/tweakreg/README.rst
mairanteodoro: I updated the docs to remove the option for a user-provided catalog.

docs/roman/tweakreg/README.rst
mairanteodoro: With the latest changes, if `abs_refcat = None`, `TweakRegStep` will use the default catalog as set by `tweakreg.DEFAULT_ABS_REFCAT`.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -47,10 +46,9 @@ class TweakRegStep(RomanStep):
     class_alias = "tweakreg"
 
     spec = f"""
-        save_catalogs = boolean(default=False) # Write out catalogs?
-        use_custom_catalogs = boolean(default=False) # Use custom user-provided catalogs?
-        catalog_format = string(default='ecsv') # Catalog output file format
-        catfile = string(default='') # Name of the file with a list of custom user-provided catalogs
+        save_catalog = boolean(default=False) # Write out the input catalog?
```
nden, line 49: Is it necessary to change the name of the parameter (compared to jwst.tweakreg)? 
`save_catalogs` -> `save_catalog`

romancal/tweakreg/tweakreg_step.py
```diff
@@ -74,60 +72,13 @@ class TweakRegStep(RomanStep):
         abs_fitgeometry = option('shift', 'rshift', 'rscale', 'general', default='rshift')
         abs_nclip = integer(min=0, default=3) # Number of clipping iterations in fit when performing absolute astrometry
         abs_sigma = float(min=0.0, default=3.0) # Clipping limit in sigma units when performing absolute astrometry
-        output_use_model = boolean(default=True)  # When saving use `DataModel.meta.filename`
     """  # noqa: E501
 
     reference_file_types = []
 
     def process(self, input):
-        use_custom_catalogs = self.use_custom_catalogs
```
nden, line 83: Are we removing the option to pass user catalogs?

docs/roman/tweakreg/README.rst
schlafly: I agree that I'm focused on the "optimize alignment order" bit here, which doesn't have meaning to me at L2.  I think we're currently imagining not needing further tweaking at L3, but we'll see.  I wasn't really advocating for removing anything, but I wanted the docs to address our usual case, where the alignment order isn't obviously a thing.

schlafly (commented):
These docs still look good to me.  I am a bit confused, though---mostly in my last review I wanted some clarification about what a couple things did when tweakreg is used at L2 time, which is the default case for Roman.  There we don't have overlaps and some steps presumably get skipped.  I think we should make this default case very clear.

I also thought that some user facing documentation of the custom catalog option would be valuable, since a lot of the docs talked about the API but as a user I would have liked to see e.g. an invocation that used that feature.  I don't think that additional documentation is strictly necessary if it's hard or if we don't have a CLI to that bit of the step, though.

romancal/tweakreg/tweakreg_step.py
mairanteodoro: This option would allow the user to save the source catalog created by `SourceDetection`.
Since that option is already available in `SourceDetectionStep`, we removed it from `TweakRegStep`.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -47,10 +46,9 @@ class TweakRegStep(RomanStep):
     class_alias = "tweakreg"
 
     spec = f"""
-        save_catalogs = boolean(default=False) # Write out catalogs?
-        use_custom_catalogs = boolean(default=False) # Use custom user-provided catalogs?
-        catalog_format = string(default='ecsv') # Catalog output file format
-        catfile = string(default='') # Name of the file with a list of custom user-provided catalogs
+        save_catalog = boolean(default=False) # Write out the input catalog?
+        catalog_format = string(default='ascii.ecsv') # Catalog output file format
+        catalog_path = string(default='~') # Catalog output file path
```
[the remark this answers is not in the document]
mairanteodoro: The current implementation will default to the current directory.

romancal/tweakreg/tweakreg_step.py
mairanteodoro: No, that was a mistake.

docs/roman/tweakreg/README.rst
```diff
@@ -0,0 +1,189 @@
...
+(i.e. the result from `SourceDetectionStep` saved in the
+`meta.tweakreg_catalog` attribute) and compares it with the
+coordinates from a Gaia catalog to compute corrections to
+the WCS of the input images such that sky catalogs obtained from the image catalogs
+using the corrected WCS will align on the sky.
+
+Alignment
+---------
+The source catalogs for each input image are compared to each other
+and linear (affine) coordinate transformations that align these
```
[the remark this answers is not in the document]
mairanteodoro: The code does skip the relative alignment and performs absolute astrometry in all input images.

That sentence was a remnant from the JWST docs. Updated.

mairanteodoro:
> These docs still look good to me. I am a bit confused, though---mostly in my last review I wanted some clarification about what a couple things did when tweakreg is used at L2 time, which is the default case for Roman. There we don't have overlaps and some steps presumably get skipped. I think we should make this default case very clear.
> 
> I also thought that some user facing documentation of the custom catalog option would be valuable, since a lot of the docs talked about the API but as a user I would have liked to see e.g. an invocation that used that feature. I don't think that additional documentation is strictly necessary if it's hard or if we don't have a CLI to that bit of the step, though.

We added an "Examples" page to the `TweakReg` docs with three scenarios, including the case where the user provides the source catalog instead of using the one provided by `SourceDetection`.

nden (approved):

PR #676, opened 2023-04-11
RCAL-490

WilliamJamieson:
Resolves [RCAL-490](https://jira.stsci.edu/browse/RCAL-490)

Together with spacetelescope/stpipe#91, this PR will remove the `romancal` dependency on `stdatamodels`.

This is a result of the use of the `stdatamodels.s3_utils` in both `stpipe` and here in `romancal`. However, the `romancal` usage of `s3_utils` is entirely the result of copying test fixtures from `jwst` to `romancal` during its setup, and there is actual usage of these fixtures by `romancal`. The `stpipe` usage of `s3_utils` can be eliminated as `s3_utils` is unused by `jwst` and `romancal`. @eslavich can provide further details.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ general
 
 - Remove the ``codecov`` dependency [#677]
 
+- Remove explicit dependence on ``stdatamodels``. [#676]
+
 source_detection
 ----------------
 - Added SourceDetection Step to pipeline [#608]
--- a/romancal/conftest.py
+++ b/romancal/conftest.py
@@ -4,20 +4,6 @@ import os
 import tempfile
 
 import pytest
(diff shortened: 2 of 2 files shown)


schlafly (commented):
This looks good to me.  We've started asking now for a link to the regression tests to be sure that we catch things failing there, though this looks pretty safe.

WilliamJamieson:
Passes regression tests, see: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/132/

schlafly (approved):
Looks good to me, thanks!

PR #677, opened 2023-04-12
Remove `codecov` test dependency

WilliamJamieson:
The `codecov-python` (`codecov` dependency) uploader has been deprecated for some time and now it has been yanked from PyPi. This means installing `romancal` with the `test` dependencies will no-longer succeed. The `codecov` dependency was only used by the regression tests to upload coverage of the regression tests, this has been replaced by the new non-python-based `codecov-uploader` in just the regression test jobs.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ general
 
 - Remove use of ``pytest-openfiles`` [#666]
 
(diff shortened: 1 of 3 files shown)


nden (approved):
The failure is unrelated, merging.

PR #686, opened 2023-04-25
Add epoch parameter to fetch catalog data.

mairanteodoro:
Resolves [RCAL-553](https://jira.stsci.edu/browse/RCAL-553)

This PR implements proper motion correction to the coordinates returned by `astrometric_utils.get_catalog`.
We also added an additional test to check for the new implementation.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


romancal/tweakreg/tests/test_astrometric_utils.py
```diff
@@ -406,3 +434,49 @@ def test_get_catalog_using_invalid_parameters(ra, dec, sr, catalog_name):
...
+    gaia_ref_epoch_coords = get_catalog(
+        ra, dec, epoch=gaia_ref_epoch, sr=sr, catalog=catalog_name
+    )
+    expected_new_ra = (
+        np.array(
+            gaia_ref_epoch_coords["ra"] * 3600
+            + (epoch - gaia_ref_epoch) * gaia_ref_epoch_coords["pmra"] / 1000
+        )
+        / 3600
+    )
```
schlafly, line 471: I am pretty sure that the pmra's are defined with a factor of cos(dec) so that np.hypot(pmra, pmdec) is the speed of the star's motion on the sky.  This test doesn't have that factor, so I'm surprised it passes?  OTOH, np.isclose is not so stringent when it comes to astrometry, where an arcsecond of motion is a lot but is only 5e-6 radians.

The parallax correction is being neglected in these tests, which will also lead to some expected error.  I'm a bit surprised that there are _no_ nearby stars that are triggering errors there?  Is there a reason that the (0, 0) search radii are different than the 10, 10 and 10, -10 search radii?

mairanteodoro: Thanks for catching the missing spherical correction term in the equation, Eddie.

Regarding the search radii, there's no specific reason for the different values for (ra, dec) = (0, 0). I had copied those parameters from another test and forgot to change them to better suit the new test.

mairanteodoro: We added a more strict condition for `np.isclose()`: `atol=1e-10` and `rtol=1e-9`; these values are used in the following equation to determine if two values are close: `absolute(a - b) <= (atol + rtol * absolute(b))` (from [numpy's docs](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html)).

schlafly:
That looks good to me, modulo the test failures that I don't understand.

I do think that we will want to include parallaxes.  We should probably either get into touch with the MAST folks about including that, or it will force us to roll our own corrections.

ddavis-stsci (approved):
LGTM

mairanteodoro:
A [ticket in the ASB board](https://jira.stsci.edu/browse/ASB-21509) has been created to add parallax correction to the MAST API.

PR #687, opened 2023-04-26
Initial sky matching implementation

bmorris3:
Resolves [RCAL-498](https://jira.stsci.edu/browse/RCAL-498)

This PR provides an initial adaptation of [the `skymatch` module in `jwst`](https://jwst-pipeline.readthedocs.io/en/latest/jwst/skymatch/skymatch.html) implemented by @mcara. I have made every effort to preserve the same functionality, tests, and docs as `jwst.skymatch`. 

The primary difference between the implementations is that I have (temporarily?) dropped support for associations, since the definition of associations for JWST and Roman is likely to be different in important ways that are still partly TBD. 

Locally the tests pass and the docs build successfully **only when using `roman_datamodels` with the fix in:** https://github.com/spacetelescope/roman_datamodels/pull/164

The algorithm is summarized nicely in [Makovoz et al. (2005)](https://ui.adsabs.harvard.edu/abs/2005PASP..117..274M/abstract). In short, it takes a group of images with different background levels (see the left panel below) and solves a system of equations for the best sky background level in each image to bring all backgrounds into agreement. The resulting matched and subtracted images are shown in the right panel below.

demo

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


romancal/skymatch/skymatch_step.py
```diff
@@ -0,0 +1,218 @@
...
+        binwidth = float(min=0.0, default=0.1) # Bin width for 'mode' and 'midpt' `skystat`, in sigma
+    """  # noqa: E501
+
+    reference_file_types = []
+
+    def process(self, input):
+        self.log.setLevel(logging.DEBUG)
+        # for now turn off memory optimization until we have better machinery
+        # to handle outputs in a consistent way.
+        self._is_asn = False
```
ddavis-stsci, line 56: Where is the memory optimization?

ddavis-stsci (changes requested):
When I try and run the pytests from your branch I get these errors,
`TypeError: Input must be a list of either strings (full path to ASDF files) or Roman datamodels.`
and I think the pytests should be self contained.

ddavis-stsci (changes requested):
You also need to add a link to your successful run of the regression tests,
see the "GitHub PR" section of 
https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=SCSB&title=Roman+Calibration+Pipeline

bmorris3:
RE @schlafly in https://github.com/spacetelescope/romancal/pull/687#issuecomment-1523812583: we discussed this a bit in the last pipeline tag-up, but I'll share my response here so others can see too. The slight mismatches that remain in the demo plot above are likely slightly exaggerated by the visualization I made. 

I've updated the visualization, and [the notebook that generates the plot below is available here](https://gist.github.com/bmorris3/bb827c00b7677dd07fbae88acbf187df):

demo

Each image is labeled with a random index, and the color is chosen from blue to red for images from low to high mean background flux. Left column is before sky match, right column is after. Histograms in the bottom row show all pixel values in each image. The image that seems like the worst match is image "6", which has a post-subtraction mean normalized by its stddev of 2%.

docs/roman/skymatch/description.rst
```diff
@@ -0,0 +1,245 @@
...
+   as well as instrumental (e.g. thermal) background, etc.
+
+The step records information in three keywords that are included in the output
+files:
+
+- BKGMETH: records the sky method that was used to compute sky levels
+
+- BKGLEVEL: the sky level computed for each image
+
+- BKGSUB: a boolean indicating whether or not the sky was subtracted from the
```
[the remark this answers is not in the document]
bmorris3: Good point! These aren't the names I gave the attributes, I just hadn't changed this part of the docs to match. Correcting them now.

romancal/skymatch/skymatch_step.py
bmorris3: This was recently implemented in `jwst` by @mcara, but not implemented here. I'll remove this comment.

romancal/skymatch/skymatch_step.py
```diff
@@ -0,0 +1,218 @@
...
+                        gim, "COMPLETE" if gim.is_sky_valid else "SKIPPED"
+                    )
+
+        return input if self._is_asn else img
+
+    def _imodel2skyim(self, image_model):
+        input_image_model = image_model
+
+        if "background" not in image_model.meta:
+            # TODO: remove this block when ``rad`` has a background schema.
```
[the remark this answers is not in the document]
bmorris3: [See here. 👍🏻 ](https://github.com/spacetelescope/rad/issues/247)

mcara:
What was matched: mode? mean? median?  This can affect the quality of the match that you observe. For 'mode', `binwidth` is important too.

bmorris3:
Thanks for the pointers @mcara. I ran the sky match step with:
```python
result = SkyMatchStep.call(image_models, subtract=False, skymethod='global+match')
```
which would choose the default sky statistic `mode`. the bin width is `0.1`.

mcara:
I would suggest that you re-do the test using skystat='mean' since that is what you use in your script when making those plots.

mcara (approved):
It looks like 99.9% of the code from JWST pipeline + PEP8 and style improvements + support for units. Should function like JWST code.

bmorris3:
> I would suggest that you re-do the test using skystat='mean' since that is what you use in your script when making those plots.

Good call @mcara, see below for a version with `skystat='mean'`, which shows even better agreement in this toy example:

demo

schlafly:
Ah, yes, that looks much better!

bmorris3:
Here's the [regression test link](https://plwishmaster.stsci.edu:8081/job/RT/job/romancal/914/).

romancal/skymatch/skymatch_step.py
```diff
@@ -0,0 +1,217 @@
...
+            dm.meta["background"]["level"] = sky
+            dm.meta["background"]["subtracted"] = (
+                self.subtract or dm.meta["background"]["subtracted"]
+            )
+
+            if self.subtract:
+                dm.data[...] = sky_image.image[...]
+
+        dm.meta.cal_step.skymatch = step_status
+
```
PaulHuwe, line 214: Can you make an issue to add skymatch to cal_step in RAD?

romancal/skymatch/skymatch_step.py
```diff
@@ -0,0 +1,217 @@
...
+        if self._is_asn:
+            dm.save(image)
+            dm.close()
```
PaulHuwe, line 217: Are you saving the infrastructure to add this to the pipeline for another ticket? (Possibly for a new pipeline - ELPP vs HLPP.)

romancal/skymatch/tests/test_skymatch.py
```diff
@@ -0,0 +1,378 @@
...
+
+    # fake a background until `rad` implements the schema:
+    l2_im.meta["background"] = dict(level=None, subtracted=False, method=None)
+
+    l2_im.meta.cal_step["skymatch"] = "INCOMPLETE"
+    return l2_im
+
+
+@pytest.fixture
+def wfi_rate():
```
PaulHuwe, line 93: Why do a simple rename here?

PaulHuwe (commented):
I assume regression tests will be done in another PR.

romancal/skymatch/region.py
```diff
@@ -0,0 +1,418 @@
...
+            "Expected vertices to be " "a list of minimum 4 tuples (x,y)"
+        )
+        super().__init__(rid, coord_system)
+
+        # self._shiftx & self._shifty are introduced to shift the bottom-left
+        # corner of the polygon's bounding box to (0,0) as a (hopefully
+        # temporary) workaround to a limitation of the original code that the
+        # polygon must be completely contained in the image. It seems that the
+        # code works fine if we make sure that the bottom-left corner of the
+        # polygon's bounding box has non-negative coordinates.
```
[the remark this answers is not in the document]
mcara: No

mcara:
> I assume regression tests will be done in another PR.

Regression tests for this step were not available in the JWST pipeline until recently: https://github.com/spacetelescope/jwst/pull/7531 They could be added to this PR or in a separate PR but ideally it should be done in this PR.

romancal/skymatch/tests/test_skymatch.py
bmorris3: This might be an artifact of the tortured commit history here, happy to adjust as you see fit. 

I wanted to write the function `mk_image_model` which might be useful (or even moved) elsewhere, and a fixture `wfi_rate` for uniform use in any tests that may need the result of `mk_image_model`. 

Then I realized that in writing `mk_image_model`, I noticed most of the adjustments were adding metadata to the result of `mk_level2_image` (WCS, background, a `cal_step` update). So it occurred to me that there might one day be a `mk_level3_image` method, and then this function might not be necessary, and the fixture can call that function instead.

But even if there's a `mk_level3_image`, it will need a method like `mk_gwcs` in order to synthesize dithers with overlap for these tests. 

So in the end I decided not to prematurely optimize and leave these working methods in the tests.

bmorris3:
For ease of finding them later, I'll collect upstream issues/links here: 
* https://github.com/spacetelescope/rad/issues/247
* https://github.com/spacetelescope/rad/issues/265
* https://github.com/spacetelescope/roman_datamodels/issues/134

romancal/skymatch/tests/test_skymatch.py
PaulHuwe: I am good with leaving it as is for future development.

PaulHuwe (approved):

nden:
Can this be merged or is anyone interested in reviewing it?

ddavis-stsci:
Does this have a Roman-Developers-Pull-Requests run? 
Other than that it has been reviewed several times (I think).

nden:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/220/
regression tests are passing

PR #697, opened 2023-05-11
Compatibility fix in tweakreg test for rdm#164

bmorris3:
spacetelescope/roman_datamodels#164 generalizes the `ModelContainer` constructor, which causes a single test in tweakreg to fail because it expects a different error type when `TweakRegStep.call` is called with incorrect arguments. 

This PR changes the tweakreg test to check that the failure is one of two types – the first error is the expected error after spacetelescope/roman_datamodels#164 is merged, the second error is the old expected error. This test should pass whether the current rdm version is before or after spacetelescope/roman_datamodels#164 gets merged.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/tweakreg/tests/test_tweakreg.py
+++ b/romancal/tweakreg/tests/test_tweakreg.py
@@ -376,8 +376,8 @@ def test_tweakreg_raises_error_on_invalid_input(input, error_type):
     with pytest.raises(Exception) as exec_info:
         TweakRegStep.call(input)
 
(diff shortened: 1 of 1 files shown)


mairanteodoro (approved):
Looks good to me.
Thanks, @bmorris3!

bmorris3:
Thanks @mairanteodoro!

nden (approved):

PR #698, opened 2023-05-15
RCAL-561: improve error-handling for TweakReg when fetching data from VO API service.

mairanteodoro:
Resolves [RCAL-561](https://jira.stsci.edu/browse/RCAL-561)

This PR adds a `try-except` block in `TweakReg` to handle errors when fetching data through `astrometric_utils.get_catalog`, which now has three additional exception catches (all subclasses of `requests.exceptions.RequestException`):
- `ConnectionError` (not able to connect to the VO API server);
- `TimeOut` (service timeout);
- `RequestException` (any other exception raised specifically by `requests`).

Each one has its own specific error message to help with debugging.

If `TweakRegStep` cannot create an absolute reference catalog (in cases where the VO API server is not available), then we'll do the default in situations where the step cannot proceed:
- print a warning message in the logs;
- set the class instance variable `skip` to `True`;
- set `meta.cal_step.tweakreg="SKIPPED"`;
- return a `ModelContainer` with all the datamodels updated.

Additional unit tests were added to check for connection and timeout errors.

Notes:
- `TweakReg` tests are all passing: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/183/testReport/romancal.tweakreg.tests/

- `romancal` regression tests are still the same as they were in the previous build (182) off of the `main` branch: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/183/

- Builds 191 through 206 are consistently showing no errors from `TweakReg`.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 0.10.1 (unreleased)
 ===================
 
+tweakreg
+--------
+- Added logic to handle cases where an absolute catalog cannot be created. [#698]
+
 associations
 ------------
 
--- a/romancal/tweakreg/astrometric_utils.py
+++ b/romancal/tweakreg/astrometric_utils.py
@@ -8,12 +8,6 @@ from astropy.coordinates import SkyCoord
 from ..assign_wcs import utils as wcsutil
 from ..resample import resample_utils
 
-# import logging
-
-# Define logging
-# log = logging.getLogger()
-# log.setLevel(logging.DEBUG)
-
 ASTROMETRIC_CAT_ENVVAR = "ASTROMETRIC_CATALOG_URL"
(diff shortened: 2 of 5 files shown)


mairanteodoro:
I'm running the regtests now; I'll add the link once it's finished.

mairanteodoro:
> I think that this does warrant a change log entry so I removed that label. It is not clear to me what happens after the exception is caught, the program just moves forward and fails when it does not have a catalog? Maybe I'm missing this bit in the code.

We can't do anything without a reference catalog so the step raises an error in that case.

ddavis-stsci (approved):
lgtm

PR #699, opened 2023-05-18
RCAL-548 Include dq flag for GW_AFFECTED_DATA

ddavis-stsci:
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/JP-nnnn)

Closes #

This PR addresses ...

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,10 @@ associations
 
 - Initial association code for GBTDS observations [#661]
 
+Documentation
+-------------
+
+- Update dq flags to include "GW_AFFECTED_DATA"  flag [#]
 
 general
 -------
--- a/docs/roman/dq_init/reference_files.rst
+++ b/docs/roman/dq_init/reference_files.rst
@@ -2,4 +2,8 @@ Reference Files
 ===============
 The ``dq_init`` step uses a MASK reference file.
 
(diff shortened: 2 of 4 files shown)


ddavis-stsci:
Updated tests with fixes for tweakreg (RCAL-561) and those tests now pass.

schlafly (approved):
Looks good to me.  I know in the context of another PR that DO_NOT_USE can have some detail in its meaning that we should spell out in more detail somewhere.

The particular case that I'm remembering is that DO_NOT_USE is some set of ~broken pixels + pixels that saturate in the first group, but are otherwise fine.  

I don't actually know what we do for Roman, but it would be good to better enumerate what DO_NOT_USE does and where it comes from.

I'm not really able to guess what leads to an unreliable error pixel.

Maybe a related bit of valuable documentation would be something like the provenance of the different flags.  i.e., most of these come from reference files and we just pass them forward.  A few (saturated, jump_det) come from our pipeline and are dynamic.  The GW stuff is related to the location of the guide windows but doesn't actually look at the values of the affected pixels themselves.  I can't remember what, if anything, we're doing for persistence.

Anyway, I don't mean to add to a PR that is mostly reorganization, just seeing these makes me remember what documentation we should think about improving going forward.

PR #704, opened 2023-05-30
RCAL-485, RCAL-558: Implement the reference pixel correction

WilliamJamieson:
Resolves [RCAL-485](https://jira.stsci.edu/browse/RCAL-485) and [RCAL-558](https://jira.stsci.edu/browse/RCAL-558)

This PR implements the reference pixel correction. It relies on the changes from spacetelescope/rad#270 and spacetelescope/roman_datamodels#190.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


WilliamJamieson:
The failing CI will be resolved by #706.

schlafly (commented):
I reviewed this before when it was in the private repo, but I had to look at this again in the context of understanding whether the reference pixel correction would also correct any discretization bias, and so I thought I'd add a few more comments.

First, this is very legible code and I like it.

I don't currently see associated read the docs documentation of this step.  I think at present we could add something at roughly the level of detail of the current run_steps(...) function (which is nice!).  FWIW, brief notes I took:
Steps
- Remove an overall offset from each pixel.
- Remove a linear trend across each amplifier in each frame (nframe * nchannel linear fits), based on the top and bottom reference pixels.
- Cosine interpolation: fill in missing times in amp33 stream via smoothing.
- FFT interpolation: further fill in missing times in amp33 stream
- Apply left / right / amp33 reference pixel correction using FFTs and ~window functions expressed via coefficients tuned to each channel.
-  Apply initial offset back to each pixel.
I don't think the docs need to be much more than that.  And of course please correct any errors I've made here.

Some broader points outside the scope of this PR that I think we should discuss with other interested parties at some point:
- I think the left / right / top / bottom reference pixels are real pixels that can see cosmic rays.  I did not notice code that attempts to identify and remove cosmic rays, and I worry that a bright / large cosmic in the reference pixel regions could end up leading to contamination across the detector.  I think the amp33 pixels are virtual and cannot be affected by cosmic rays (?), so this isn't a concern there, but things like the linear trend removal and left/right reference pixel correction seem potentially brittle.
- I didn't understand the motivation for the overall offset removal & re-addition.

schlafly:
One more comment: I like the structure of the PR but we could think about memory overhead.
- It's somewhat unavoidable to read the whole thing into memory; it's floats rather than int16s so we're using ~twice the memory intrinsically needed by the file.  It also makes error and dq arrays (5x original data volume).
- the standardview.from_datamodel(...) will make a copy of the data.
- the conversion to channels will make another copy of the data.

I don't think any other steps use an appreciable amount of memory, though maybe I missed something.  So my expectation is that this would use 9x the original data volume (5x from the data model, 2x & 2x from the two copies).  It would be good to know if that expectation is borne out.  It sounds challenging to me to make progress on the datamodel side, though it might be worth discussing.  I think technically the ramp detection algorithm never needs to look at more than one channel at a time, so one could imagine saving a lot of memory by having an outer loop over channels and only making copies of each channel separately.  But 9x doesn't sound too bad; for a 100 resultant file that would be 256 GB of memory.

WilliamJamieson:
> One more comment: I like the structure of the PR but we could think about memory overhead.
> 
> * It's somewhat unavoidable to read the whole thing into memory; it's floats rather than int16s so we're using ~twice the memory intrinsically needed by the file.  It also makes error and dq arrays (5x original data volume).
> * the standardview.from_datamodel(...) will make a copy of the data.
> * the conversion to channels will make another copy of the data.
> 
> I don't think any other steps use an appreciable amount of memory, though maybe I missed something. So my expectation is that this would use 9x the original data volume (5x from the data model, 2x & 2x from the two copies). It would be good to know if that expectation is borne out. It sounds challenging to me to make progress on the datamodel side, though it might be worth discussing. I think technically the ramp detection algorithm never needs to look at more than one channel at a time, so one could imagine saving a lot of memory by having an outer loop over channels and only making copies of each channel separately. But 9x doesn't sound too bad; for a 100 resultant file that would be 256 GB of memory.

I don't think it is necessary at this time to divide it over channels, outside of the offset computation the step is entirely independent across frames, which are looped over in every step. So a more efficient (in terms of memory mapping from an ASDF file) to compute by frame rather than channel. We can also reduce by channel in addition to looping by frame. That being said I do not think it is worth delaying this step going in to chase small optimizations like this at this time, instead I think we should wait until we have a better idea of the performance bottlenecks later.

WilliamJamieson:
I had to add bigdata skips because when generating the coefficients (a huge dataset) had the possibility of exceeding GitHub actions memory limits (sometimes it would sometimes it would not). This is part of some pseudo regression testing, so I think it is fine to just run it as part of the regression testing for now.

Also, I cannot easily reduce the data sizes because the reference pixel correction makes a lot of intrinsic assumptions about the size and shape of the data (to reshape and pad the data in the right places). Thus the movement of the testing to the regression tests is the best option for stability.

WilliamJamieson:
Since tests have been moved to the regression suite, they do still indeed pass: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/245/

romancal/refpix/refpix_step.py
```diff
@@ -0,0 +1,49 @@
...
+
+    reference_file_types = ["refpix"]
+
+    def process(self, input):
+        """
+        Perform the reference pixel correction
+        """
+
+        # open the input data model
+        log.info(f"Opening the input data model: {input}")
```
nden, line 39: I would set this at `debug` level and change it to `Opening the science data.`

romancal/refpix/data.py
```diff
@@ -0,0 +1,637 @@
...
+        - CHAN_WIDTH: the number of columns in a channel (detector amplifier)
+        - N_CHANNELS: the number of channels for the detector
+        - N_COLUMNS: the total number of columns in the detector
+    """
+
+    REF = 4
+    PAD = 12
+    CHAN_WIDTH = 128
+    N_DETECT_CHAN = 32
+    N_COLUMNS = CHAN_WIDTH * N_DETECT_CHAN
```
nden, line 34: Would it be better to document the constants directly, not in the class string, i.e.
```
REF = 4
# The number of columns used by the detector reference pixels
```

nden:
Before I continue with the review I'd like to understand the reason for the `Control` class. 
Configuration for a step can come from three places - the step, a parameter file in CRDS, user supplied parameter file. The mechanism for this is coded in stpipe. Do I understand correctly that the `Control` class will override this mechanism and read the configuration from the step and does it interfere with the other options?

romancal/refpix/data.py
WilliamJamieson: No, I don't think that would be a good idea in this case. The constants enum should be pulled into the documentation automatically (I don't think I have actually written the documentation yet for refpix), meaning the doc string will get incorporated into the docs. Using a comment will not be pulled into the documentation.

Note that the reason for creating constants this way, is so that they are immutable. If they were not wrapped into the enum, there is the potential risk of them being altered by mistake.

WilliamJamieson:
I do not understand what you are asking. If you look in the step itself a Control instance constructed directly from the step itself where the information is defined within the "spec" for the step. In other words, the Control instance is simply a convenience wrapper around the necessary arguments defined by the "spec" so that they could be passed as a single unit. 

The explicit reason I did this is so that the code I wrote to run the actual computation is both explicit and independent of the stpipe framework. I am not a fan of the fact that the "spec" is incorporated dynamically into the step object at runtime via a not well documented pseudo parser. The independence from the step itself means that I can adapt the computation to whatever changes are made to the dynamic portions of the step while only having to adjust the constructor for the Control instances. Moreover, the explicit (as in variables being defined as actual Python code) nature of the object means that the actual computation code can be statically analyzed outside of runtime in a meaningful way (excluding the constructors for the objects from the step and datamodel itself).

romancal/refpix/data.py
nden: The question then to @schlafly and @tddesjardins is - do you envision end users wanting to change any of the constants so that they can explore different options of the step?

nden:
> The explicit reason I did this is so that the code I wrote to run the actual computation is both explicit and independent of the stpipe framework.

If you are taking about `run_steps`, it can still be stpipe independent, while passing the parameters explicitly in refpix_step. stpipe implements a lot of boilerplate that otherwise steps need to implement. I have not run tests for various use cases, my concern is the `Control` class implements only a subset of those and I don't see the need for it. 

@ddavis-stsci The process method is used by stpipe but it is present in the step.

WilliamJamieson:
> If you are taking about `run_steps`, it can still be stpipe independent, while passing the parameters explicitly in refpix_step. stpipe implements a lot of boilerplate that otherwise steps need to implement. I have not run tests for various use cases, my concern is the `Control` class implements only a subset of those and I don't see the need for it.

I still fail to see the issue with the `Control` object. It is simply a convenience wrapper around information which must be in the step itself. Indeed, it is functionally equivalent to defining a `NamedTuple` (or even a plan tuple) around the values in question, it being a dataclass is simply a matter of style and clarity. In fact, you can see that the `Control` is constructed from the current step instance it self [here](https://github.com/WilliamJamieson/romancal/blob/e4a6e87292019c39009663da476991837a7f54f5/romancal/refpix/refpix_step.py#L47). I am not doing any thing fancy with any external interfaces or reading things out of some strange place, I am simply packing multiple bits if information:  `remove_offset`, `remove_trends`, `cosine_interpolate`, and `fft_interpolate` values which are defined by the [`spec`](https://github.com/WilliamJamieson/romancal/blob/e4a6e87292019c39009663da476991837a7f54f5/romancal/refpix/refpix_step.py#L20-L29), into a single object.

> I agree with Nadia here. In addition for most of the steps the work is done in a routine called "process". While this is rather non-specific it would be good to have the some syntax across the cal code.

I can name the code in question whatever you want or place it within whatever interface you want it. However, that is irrelevant to the fact that I am simply wrapping information into a connivence object that I called `Config`.

romancal/refpix/data.py
WilliamJamieson: I don't think allowing this is a very safe idea. There are a lot of baked in assumptions about how the computation is structured which are related to these values. While the code always references these constants and I tried to make the code independent of these constants, there are subtle assumptions about the relationships among the constants which if violated could cause unexpected behavior of the correction. Hence, my efforts to make the constants actually constant.

schlafly: These constants are properties of the detector and will not change.  Maybe one could imagine changes to the readout mode changing PAD.  But that would be a pretty major change to the detector configuration and isn't something the user would want to change.  I don't expect we'll need to change it.

romancal/refpix/tests/test_refpix.py
```diff
@@ -0,0 +1,26 @@
...
+from astropy import units as u
+
+from romancal.refpix.data import StandardView
+from romancal.refpix.refpix import Control, run_steps
+
+from . import reference_utils
+
+
+# GitHub actions don't have enough memory to run this test
+@pytest.mark.bigdata
```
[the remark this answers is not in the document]
WilliamJamieson: See https://github.com/spacetelescope/romancal/pull/704#issuecomment-1603123731. This correction is very much tied to the physical configuration of the detectors and so it is not simple to use smaller arrays.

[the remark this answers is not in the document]
schlafly: Playing devil's advocate briefly, I think the reference pixel correction algorithm is probably pretty well behaved in the number of rows, so you might be able to make a fake exposure shape that instead of being 4096 rows long were only ~12 rows long.  I think you only need to be >8, otherwise things stop making sense due to padding?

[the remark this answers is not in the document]
WilliamJamieson: @schlafly that will actually work. Though it will be a pain because the reference code I am testing against makes a hard assumption on the number of rows...

romancal/stpipe/integration.py
```diff
@@ -29,6 +29,7 @@ def get_steps():
         ("romancal.step.LinearityStep", None, False),
         ("romancal.step.PhotomStep", None, False),
         ("romancal.step.RampFitStep", None, False),
+        ("romancal.step.RefPixStep", None, False),
```
nden, line 32: @ddavis-stsci I thought we can setup aliases for steps here. Is there a general issue with step aliases?

romancal/refpix/refpix_step.py
```diff
@@ -0,0 +1,55 @@
...
+                # Run the correction
+                log.debug("Running the reference pixel correction")
+                return refpix.run_steps(
+                    datamodel,
+                    refs,
+                    self.remove_offset,
+                    self.remove_trends,
+                    self.cosine_interpolate,
+                    self.fft_interpolate,
+                )
```
nden, line 55: @WilliamJamieson One other thing I forgot to mention is adding a flag to the data if the step completed successfully, [like here](https://github.com/spacetelescope/romancal/blob/main/romancal/assign_wcs/assign_wcs_step.py#L100).

nden: This requires changing the `calstep` schema but perhaps that was already done

WilliamJamieson: Looks like https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/cal_step-1.0.0.yaml will need to be updated.

romancal/refpix/tests/test_refpix.py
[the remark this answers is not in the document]
WilliamJamieson: @schlafly your suggestion will work, see 310cc414628d3026e9d5c399216e8135dbd2a07c. However, it required several modifications to the Goddard code we are testing against. If you are fine with those changes then we can keep them.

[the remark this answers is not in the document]
schlafly: For my two cents those changes look pretty sane and look valuable to me for unit test purposes.

romancal/refpix/refpix_step.py
nden, line 55: When I run the step on the command line using
```
strun romancal.refpix.RefPixStep r0000201001001001003_01101_0001_WFI16_dqinit.asdf --override_refpix refpix_reference.asdf
```
I get  a traceback ending in 
```
...
  File "/Users/dencheva/opt/anaconda3/envs/rb10/lib/python3.9/site-packages/asdf/generic_io.py", line 382, in seek
    result = self._fd.seek(offset, whence)
ValueError: seek of closed file
```

I think the reason is that the datamodel is returned within a couple of context managers which close the file before exiting the function. We need the `process` function to return the datamodel so it can be passed to the next step in the pipeline, or saved to file by the user.

WilliamJamieson: The issue you are seeing is from the outer context manager not the because of the use of nested managers.

I updated the model in place to avoid copying input model. Would you rather I just copy the input model and return the copy?

ddavis-stsci (changes requested):
Unfortunately this is still failing,

```
strun romancal.step.RefPixStep r0000101001001001001_01101_0001_WFI01_saturation.asdf
----------------------------------------------------------------------
ERROR PARSING CONFIGURATION:
    'romancal.step.RefPixStep' is not a path to a config file or a
    Python Step class
----------------------------------------------------------------------
usage: strun [--logcfg LOGCFG] [--verbose] [--debug] [--save-parameters SAVE_PARAMETERS] [--disable-crds-steppars] cfg_file_or_class

Run an stpipe Step or Pipeline

positional arguments:
  cfg_file_or_class     The configuration file or Python class to run

options:
  --logcfg LOGCFG       The logging configuration file to load
  --verbose, -v         Turn on all logging messages
  --debug               When an exception occurs, invoke the Python debugger, pdb
  --save-parameters SAVE_PARAMETERS
                        Save step parameters to specified file.
  --disable-crds-steppars
                        Disable retrieval of step parameter references files from CRDS
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 45, in _get_config_and_class
    step_class = utilities.import_class(
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/utilities.py", line 70, in import_class
    step_class = getattr(imported, class_name)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'romancal.step' has no attribute 'RefPixStep'

During handling of the above exception, another exception occurred:

  File "/Users/ddavis/miniconda3/envs/rcal_dev/bin/strun", line 26, in <module>
    step = Step.from_cmdline(sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/step.py", line 186, in from_cmdline
    return cmdline.step_from_cmdline(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 375, in step_from_cmdline
    step, step_class, positional, debug_on_exception = just_the_step_from_cmdline(
                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 232, in just_the_step_from_cmdline
    step_class, config, name, config_file = _get_config_and_class(
                                            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 49, in _get_config_and_class
    raise ValueError(
ValueError: 'romancal.step.RefPixStep' is not a path to a config file or a Python Step class
```

You also need to update your version of stcal for running the pipeline, 
```
File "/Users/ddavis/src/Roman/romancal/romancal/jump/jump_step.py", line 121, in process
    gdq, pdq = detect_jumps(
    ^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

nden:
Hm, that worked for me when I tested it. Perhaps something changed. I'll test it again.

WilliamJamieson:
> Unfortunately this is still failing,
> 
> ```
> strun romancal.step.RefPixStep r0000101001001001001_01101_0001_WFI01_saturation.asdf
> ----------------------------------------------------------------------
> ERROR PARSING CONFIGURATION:
>     'romancal.step.RefPixStep' is not a path to a config file or a
>     Python Step class
> ----------------------------------------------------------------------
> usage: strun [--logcfg LOGCFG] [--verbose] [--debug] [--save-parameters SAVE_PARAMETERS] [--disable-crds-steppars] cfg_file_or_class
> 
> Run an stpipe Step or Pipeline
> 
> positional arguments:
>   cfg_file_or_class     The configuration file or Python class to run
> 
> options:
>   --logcfg LOGCFG       The logging configuration file to load
>   --verbose, -v         Turn on all logging messages
>   --debug               When an exception occurs, invoke the Python debugger, pdb
>   --save-parameters SAVE_PARAMETERS
>                         Save step parameters to specified file.
>   --disable-crds-steppars
>                         Disable retrieval of step parameter references files from CRDS
> Traceback (most recent call last):
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 45, in _get_config_and_class
>     step_class = utilities.import_class(
>                  ^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/utilities.py", line 70, in import_class
>     step_class = getattr(imported, class_name)
>                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: module 'romancal.step' has no attribute 'RefPixStep'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/bin/strun", line 26, in <module>
>     step = Step.from_cmdline(sys.argv[1:])
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/step.py", line 186, in from_cmdline
>     return cmdline.step_from_cmdline(args)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 375, in step_from_cmdline
>     step, step_class, positional, debug_on_exception = just_the_step_from_cmdline(
>                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 232, in just_the_step_from_cmdline
>     step_class, config, name, config_file = _get_config_and_class(
>                                             ^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/stpipe/cmdline.py", line 49, in _get_config_and_class
>     raise ValueError(
> ValueError: 'romancal.step.RefPixStep' is not a path to a config file or a Python Step class
> ```

@ddavis-stsci and @nden I think I figured out this. On the main branch with any of the steps which are currently on main and functional, I can reproduce this failure for those steps if I install `romancal` with the editable flag (`-e`). If I do a non-editable install of `romancal`, then this runs the step.

This extends to these changes; that is, I can reproduce this reported failure if `romancal` is installed with the `-e` flag and resolve this error if `romancal` is installed normally (without the `-e` flag). 

Is this known behavior, or some bug that has arisen as part of "recent" changes?

WilliamJamieson:
Regression tests pass: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/293/

nden (approved):
LGTM

ddavis-stsci (approved):
Looks good the run is now succeeding,
```
strun romancal.step.RefPixStep r0000101001001001001_01101_0001_WFI01_saturation.asdf
2023-07-26 08:32:45,846 - CRDS - ERROR -  Error determining best reference for 'pars-refpixstep'  =   Unknown reference type 'pars-refpixstep'
2023-07-26 08:32:45,847 - stpipe.RefPixStep - INFO - RefPixStep instance created.
2023-07-26 08:32:45,947 - stpipe.RefPixStep - INFO - Step RefPixStep running with args ('r0000101001001001001_01101_0001_WFI01_saturation.asdf',).
2023-07-26 08:32:45,948 - stpipe.RefPixStep - INFO - Step RefPixStep parameters are: {'pre_hooks': [], 'post_hooks': [], 'output_file': None, 'output_dir': None, 'output_ext': '.asdf', 'output_use_model': False, 'output_use_index': True, 'save_results': True, 'skip': False, 'suffix': None, 'search_output_file': True, 'input_dir': '', 'remove_offset': True, 'remove_trends': True, 'cosine_interpolate': True, 'fft_interpolate': False}
2023-07-26 08:32:46,216 - CRDS - INFO -  Fetching  /Users/ddavis/crds_cache/references/roman/wfi/roman_wfi_refpix_0045.asdf  440.4 M bytes  (1 / 1 files) (0 / 440.4 M bytes)
2023-07-26 08:33:32,451 - stpipe.RefPixStep - WARNING - /Users/ddavis/miniconda3/envs/rcal_dev/lib/python3.11/site-packages/asdf/asdf.py:352: AsdfWarning: File 'file:///Users/ddavis/crds_cache/references/roman/wfi/roman_wfi_refpix_0045.asdf' was created with extension URI 'asdf://stsci.edu/datamodels/roman/extensions/datamodels-1.0' (from package roman-datamodels==0.16.1), but older package (roman-datamodels==0.15.1.dev74+g1773850) is installed.
2023-07-26 08:33:32,451 - stpipe.RefPixStep - WARNING -   warnings.warn(msg, AsdfWarning)
2023-07-26 08:33:32,451 - stpipe.RefPixStep - WARNING -
2023-07-26 08:33:33,812 - stpipe.RefPixStep - WARNING - /Users/ddavis/src/Roman/romancal/romancal/refpix/data.py:417: RuntimeWarning: invalid value encountered in divide
2023-07-26 08:33:33,812 - stpipe.RefPixStep - WARNING -   kern = (cov / mask_conv).reshape(rows, columns)
2023-07-26 08:33:33,812 - stpipe.RefPixStep - WARNING -
2023-07-26 08:33:41,259 - stpipe.RefPixStep - INFO - Saved model in r0000101001001001001_01101_0001_WFI01_refpixstep.asdf
2023-07-26 08:33:41,259 - stpipe.RefPixStep - INFO - Step RefPixStep done
```

PR #714, opened 2023-06-12
RCAL-590: add fit results to datamodel's meta.

mairanteodoro:
Resolves [RCAL-590](https://jira.stsci.edu/browse/RCAL-590)

This PR addresses adding the alignment fit results to the datamodel's `meta` tree under `wcs_fit_results`.

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)
- [ ] provided regression test link

### DIFF
--- a/docs/roman/tweakreg/README.rst
+++ b/docs/roman/tweakreg/README.rst
@@ -70,8 +70,56 @@ Alignment
 ---------
 The source catalog (either created by `SourceDetectionStep` or provided by the user)
 gets cross-matched and fit to an astrometric reference catalog
-(set by ``TweakRegStep.abs_refcat``). The pipeline initially supports fitting to
-any Gaia Data Release (defaults to `GAIADR3`).
+(set by ``TweakRegStep.abs_refcat``) and the results are stored in
+``model.meta.wcs_fit_results``. The pipeline initially supports fitting to any
+Gaia Data Release (defaults to `GAIADR3`).
+
+An example of the content of ``model.meta.wcs_fit_results`` is as follows:
+
+  .. code-block:: python
+
+          model.meta.wcs_fit_results = {
+            "status": "SUCCESS",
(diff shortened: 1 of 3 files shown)


docs/roman/tweakreg/README.rst
```diff
@@ -70,8 +70,56 @@ Alignment
...
+                  65.9862519 , 65.98668115, 65.98882216, 65.98995814, 65.99429907,
+                  65.99490411, 65.99604151, 65.99798312, 65.99986204, 66.00201329,
+                  66.00214747, 66.0030868 , 66.00642501, 66.00807803, 66.00859865,
+                  66.01220723, 66.01447642, 66.01479239]),
+            "nmatches": 22,
+            "matched_ref_idx": array([ 0, 23, 15, 29, 36, 14, 35,  7, 32, 13,  9, 44,
+                   1, 40, 45, 46, 21, 16,  6, 28, 38, 18], dtype=uint64),
+            "matched_input_idx": array([18, 16,  8, 21, 20,  6,  4, 17,  2, 12,  3, 14,
+                  15, 13, 19,  0, 10, 7,  1,  5, 11,  9], dtype=uint64)
+          }
```
schlafly, line 117: This looks great for the most part.  For what it's worth, though, I would not have included the matched_ref_idx, matched_input_idx, or fitmask attributes.  If we go that route I think we start needing to include the whole catalog entries from Gaia & roman for those stars.  e.g. if we're not including the Gaia results in the catalog, matched_ref_idx is very hard to use.

I am not clear on eff_minobj after reading the tweakreg docs.

mairanteodoro: That's the entire output from TwekWCS for reference; we can keep/remove any parameters we want.

ddavis-stsci (approved):
LGTM

PR #770, opened 2023-07-05
Register skymatch with stpipe

WilliamJamieson:
This PR registers `skymatch` with `stpipe` so that it can be run via the `strun` command.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)


ddavis-stsci (approved):
LGTM

PR #771, opened 2023-07-05
Make test randomness deterministic

### LINKED ISSUE #769: Rampfit test intermittently failing
The fix in #725 does not quite work all of the time, on the main branch I am getting intermittent failures in the
```
romancal/ramp_fitting/tests/test_ramp_fit.py::test_one_group_small_buffer_fit_ols
```
test cases. Iterative running of this test locally (it is skipped in CI) reproduces the error locally for me. For example I get the error:
```
E           AssertionError: 
E           Not equal to tolerance rtol=1e-06, atol=0
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 1.06613152e-11
E           Max relative difference: 1.06613152e-06
E            x: array(-1.000001e-05, dtype=float32)
E            y: array(-1.e-05)
```

I believe the root of this error is: https://github.com/spacetelescope/romancal/blob/2396bc7bb78e00242e705ecb588157bf1a182f2a/romancal/ramp_fitting/tests/test_ramp_fit.py#L32-L36
The intermittent failure is caused by the fact that the data/err change from run-to-run of the tests. Instead, the random number generator should be seeded so that every test is exactly reproducible among different test runs.

WilliamJamieson:
Closes #769

This PR transitions all the random number generation in romancal to be both deterministic and to use the recommend numpy [Random Generator](https://numpy.org/doc/stable/reference/random/generator.html) to control random number generation. All the random number generations are seeded so the tests are entirely reproducible.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -50,6 +50,9 @@ general
 
 - Remove unused extras (``ephem``, ``lint``) from build configuration and regression testing [#784]
 
+- Make all random number generation for tests both seeded and use the same random
+  number generation system. [#771]
+
 0.11.0 (2023-05-31)
 ===================
 
--- a/romancal/dq_init/tests/test_dq_init.py
+++ b/romancal/dq_init/tests/test_dq_init.py
@@ -10,6 +10,8 @@ from romancal.dq_init import DQInitStep
 from romancal.dq_init.dq_initialization import do_dqinit
 from romancal.lib import dqflags
 
+RNG = np.random.default_rng(83)
+
 # Set parameters for multiple runs of data
 args = "xstart, ystart, xsize, ysize, ngroups, instrument, exp_type"
 test_data = [(1, 1, 2048, 2048, 2, "WFI", "WFI_IMAGE")]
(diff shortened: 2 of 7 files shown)


schlafly:
Just staring at it, looks good to me.  Are the regression tests passing?  Clearly the docs failure is a separate issue somewhere?

WilliamJamieson:
> Just staring at it, looks good to me.  Are the regression tests passing?  Clearly the docs failure is a separate issue somewhere?

Yep it was fixed by #773.

schlafly:
I'm happy to approve this but I need the regression test link to do so---though I can't imagine this leading to regtest issues.  Thanks!

WilliamJamieson:
> I'm happy to approve this but I need the regression test link to do so---though I can't imagine this leading to regtest issues.  Thanks!

There is no need to run the regression tests. If you look all the changes here are only to tests which are run by the GitHub CI.

schlafly (approved):
LGTM.

PR #772, opened 2023-07-05
Use same changelog workflow as the rest of the roman repositories

WilliamJamieson:
This PR updates the changelog CI to use the same changelog workflow used by `rad` and `roman_datamodels`. This workflow draws from an independent centralized workflow that has several additional features (including milestone verification if we wish to turn that on).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,16 +1,25 @@
-name: Ensure changelog
+name: Changelog
 
 on:
   pull_request:
     types: [labeled, unlabeled, opened, synchronize, reopened]
 
+# Only cancel in-progress jobs or runs for the current workflow
+#   This cancels the already triggered workflows for a specific PR without canceling
+#   other instances of this workflow (other PRs, scheduled triggers, etc) when something
+#   within that PR re-triggers this CI
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  ensure_changelog:
-    name: Verify that a changelog entry exists for this pull request
+  changelog:
+    name: Confirm changelog entry
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-        with:
-          submodules: true
-      - run: grep -P '\[[^\]]*#${{github.event.number}}[,\]]' CHANGES.rst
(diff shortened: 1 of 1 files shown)


WilliamJamieson:
The independent workflow is managed by @pllim.

pllim:
FWIW Brigitta asked if I am interested to upstream this to Scientific Python but I was hesitant because I am scared it is going to get refactored and break stuff. But if STScI is not comfortable with this workflow in my own account, I can revisit that idea with Brigitta. It should not go under spacetelescope because it is used by Astropy too.

zacharyburnett:
> FWIW Brigitta asked if I am interested to upstream this to Scientific Python but I was hesitant because I am scared it is going to get refactored and break stuff. But if STScI is not comfortable with this workflow in my own account, I can revisit that idea with Brigitta. It should not go under spacetelescope because it is used by Astropy too.

would it be possible to put it under OpenAstronomy?

WilliamJamieson:
> FWIW Brigitta asked if I am interested to upstream this to Scientific Python but I was hesitant because I am scared it is going to get refactored and break stuff. But if STScI is not comfortable with this workflow in my own account, I can revisit that idea with Brigitta. It should not go under spacetelescope because it is used by Astropy too. 

We can fork and change it if it starts breaking stuff for space telescope.

pllim:
I am not a member of OpenAstronomy

pllim:
Let me ask around. But in the meantime, nothing stops you from using it as-is right now.

WilliamJamieson:
> Let me ask around. But in the meantime, nothing stops you from using it as-is right now.

Its used by the rest of roman, asdf, and jdaviz... so we (STScI) have motivation to fork back if there ends up being an issue like you fear.

pllim:
Yeah but even if I move it, the existing call would still work. So no reason to wait around for a political move here.

pllim:
I'll probably just batch PR to all the affected repos when the move happens.

bsipocz:
> FWIW Brigitta asked if I am interested to upstream this to Scientific Python but I was hesitant because I am scared it is going to get refactored and break stuff.

Well, just to chime in, the format is used in my repos, too, so I'm motivated enough not to break the current format :)

zacharyburnett (approved):

PR #773, opened 2023-07-05
Remove myst parser from docs config

WilliamJamieson:
This PR removes the `myst_parser` from the docs as it is not needed. `sphinx-asdf` removed this the `myst_parser` recently.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)


braingram (approved):
Thanks for the quick fix.

Looks good to me!

PR #725, opened 2023-06-28
RCAL-604 update ramp fitting unit test for stcal 1.4.0

ddavis-stsci:
Resolves [RCAL-604](https://jira.stsci.edu/browse/RCAL-604)

Closes #

This PR updates the ramp fitting unit tests for changes in stcal needed by JWST (JP-3242)

**Checklist**
- [x] added entry in `CHANGES.rst` under the corresponding subsection
- [x] updated relevant tests
- [x] updated relevant documentation
- [x] updated relevant milestone(s)
- [x] added relevant label(s)

```
pytest --pdb ~/src/Roman/romancal/romancal/ramp_fitting/tests/test_ramp_fit.py
=============================================================== test session starts ================================================================
platform darwin -- Python 3.11.0, pytest-7.3.1, pluggy-1.0.0
rootdir: /Users/ddavis/src/Roman/romancal
configfile: pyproject.toml
plugins: remotedata-0.4.0, env-0.8.1, asdf-2.15.0, mock-3.10.0, filter-subpackage-0.1.2, astropy-header-0.2.2, hypothesis-6.75.3, doctestplus-0.12.1, astropy-0.10.0, cov-4.0.0, ci-watson-0.6.1, openfiles-0.5.0, arraydiff-0.5.0
collected 9 items

../../../romancal/romancal/ramp_fitting/tests/test_ramp_fit.py .........                                                                     [100%]

================================================================ 9 passed in 7.13s
```
The regression tests all pass,
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/259/

PR #774, opened 2023-07-05
Try to avoid copying data models.

schlafly:
This PR tries to reduce memory usage in romancal.  My approach has been pretty aggressive so far and should be reined in.  I have focused on the steps preceding jump detection and ramp fitting, which we are replacing.  See
https://github.com/spacetelescope/romanisim/pull/64
for a memory efficient ramp fitting approach.  A jump detection approach following plans for Roman based around that routine should also be very memory efficient.

I have been using this is conjunction with
https://github.com/spacetelescope/roman_datamodels/pull/232
to reduce memory usage.  On my memray runs today, up until jump detection the new version uses 3.5 GB in comparison to 10 GB for romancal default.  But the datamodel size is O(1 GB) so I should be able to get closer to 2 GB.  To do better than that would require not reading the complete dark model into memory and instead removing it plane-by-plane or something, which is certainly an option.  I'll poke a bit further tomorrow, but I wanted other people to be able to see what damage I'm doing.

I re-implemented dark subtraction because I didn't want to have to visit stcal's dark subtraction and there really isn't a lot going on there.  I have some ridiculous bits like `out_data = input_model` that are guaranteed to cause confusion but more closely follow the original flow; were we to actually want to go in this direction I would clean that up.  Some tests needed to change to reflect that now the steps operate in place rather than making copies (due to my change to roman_datamodels https://github.com/spacetelescope/roman_datamodels/pull/232).  

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/linearity/linearity_step.py
+++ b/romancal/linearity/linearity_step.py
@@ -42,6 +42,8 @@ class LinearityStep(RomanStep):
 
             # copy poly coeffs from linearity model so Nan's can be updated
             lin_coeffs = lin_model.coeffs.copy()
+            # I feel like we should be able to do this in place, but I can't figure
+            # out how.  asdf internal arrays want to be immutable.
             lin_dq = lin_model.dq  # 2D pixeldq from linearity model
 
             gdq = input_model.groupdq  # groupdq array of input model
@@ -49,7 +51,7 @@ class LinearityStep(RomanStep):
 
             gdq = gdq[np.newaxis, :]
 
-            output_model = input_model.copy()
+            output_model = input_model
             output_model.data = output_model.data[np.newaxis, :]
 
             # Call linearity correction function in stcal
--- a/romancal/dark_current/dark_current_step.py
+++ b/romancal/dark_current/dark_current_step.py
@@ -44,28 +44,21 @@ class DarkCurrentStep(RomanStep):
                     "groupgap"
                 ] = input_model.meta.exposure.groupgap
 
-            # Reshaping data variables for stcal compatibility
-            input_model.data = input_model.data[np.newaxis, :]
-            input_model.groupdq = input_model.groupdq[np.newaxis, :]
-            input_model.err = input_model.err[np.newaxis, :]
-
             # Do the dark correction
-            out_data, dark_data = dark_sub.do_correction(
-                input_model, dark_model, self.dark_output
-            )
+
+            # out_data, dark_data = dark_sub.do_correction(
+            #     input_model, dark_model, self.dark_output
+            # )
+            # this is somewhat committed to doing a deep copy.
+            out_data = input_model
+            out_data.data -= dark_model.data
+            out_data.pixeldq |= dark_model.dq
+            out_data.meta.cal_step.dark = 'COMPLETE'
 
             # Save dark data to file
-            if dark_data is not None and dark_data.save:
-                save_dark_data_as_dark_model(dark_data, dark_model)
+            if self.dark_output is not None:
+                dark_model.save(self.dark_output)  # not clear to me that this makes any sense
             dark_model.close()
-
-            # Reshaping data variables back from stcal
-            input_model.data = input_model.data[0]
-            input_model.groupdq = input_model.groupdq[0]
-            input_model.err = input_model.err[0]
-
-            # Convert data to RampModel
-            out_ramp = dark_output_data_as_ramp_model(out_data, input_model)
 
         if self.save_results:
             try:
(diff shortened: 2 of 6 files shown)


romancal/linearity/linearity_step.py
```diff
@@ -42,14 +42,16 @@ def process(self, input):
 
             # copy poly coeffs from linearity model so Nan's can be updated
             lin_coeffs = lin_model.coeffs.copy()
+            # I feel like we should be able to do this in place, but I can't figure
+            # out how.  asdf internal arrays want to be immutable.
```
braingram, line 46: I'm fairly certain this is because line 41 opened the LinearityRefModel as read only and memory mapped (I'm running the test locally now but it's very slow).

If this is the case, the write protections will prevent the array from being modified.

I don't believe any `kwargs` passed to the model (via __init__ when called with a filename) make it to asdf:
https://github.com/spacetelescope/roman_datamodels/blob/c0a084d990f5904f577d9c6c29391fcb44718112/src/roman_datamodels/datamodels/_core.py#L172
so asdf will use the defaults (read only, copy_arrays=False).

schlafly: *laugh*, I was just thinking about your talk and copy_arrays=True, which indeed makes this work in place without a copy.  I haven't decided yet if that's actually a win.

braingram: Excellent!

Let me know if there's anything I can do to help. My attempts to run the regtest locally (on the vpn) failed with permission errors for `/grp/crds/cache/mappings/roman' (after running for a while). I'm not sure what's going on there so it's good to hear you were able to test out the change.

schlafly:
Okay, I'm feeling better about this PR.  Regression tests are passing, here:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/281/

The overall goal here is to make the pipeline modify everything in place and pass to the extent possible a single datamodel through that gets updated until it reaches its final form and is returned.  Most of the changes in this PR are related to that goal.  The new version uses 2 GB through dark subtraction on my test file, while the old version used 10 GB.  With planned updates to ramp fitting and jump detection I do not expect those steps to add meaningfully to the total memory usage.  However, I haven't touched those steps yet and they currently use more memory than the earlier steps I have addressed.

There are some issues with the stpipe design and the desire to have the steps be able to process either file names or data models.  In the first case, we need to open the file name, and then are responsible for closing the file when we're done with it.  If we use features like lazy loading and memory mapping, returning an asdf object with a closed file causes problems.  So I've changed things so that steps turn lazy loading off, essentially detaching the datamodel objects from their backing files and allowing the files to be closed and the models returned.  This is kind of sad---lazy loading makes sense in the pipeline, as does memory mapping.  But the previous approach was to make a deep copy of each object, which also ended up removing the benefits of these features.  I think if we want to do better here we need to change the pipeline so that it only ever deals with open datamodels and callers are responsible for opening and closing files (following @WilliamJamieson's suggestions).  Then we would remove the context management on the input objects from the pipeline.

In the current implementation I've kept the context managers to keep changes to a minimum.  It's still the case that if someone sends us an open datamodel with a backing file, it would be bad form for us to close it.  The associated PR https://github.com/spacetelescope/roman_datamodels/pull/232 does a shallow copy when opening an already open datamodel that prevents exiting the context manager for the proxy copy from closing the open file.  It feels surprising and un-idiomatic that opening an already open file returns a shallow copy of that file, and we're doing work to use a context manager when we don't have any reason to use one for this case.  But doing better there seems to me like it would require some redesign of the pipeline / stpipe to only work from datamodels and not from files.

@braingram notes that lazy loading is not available for files with quantities.  This is unfortunate.  This means that blocks like this https://github.com/spacetelescope/stpipe/pull/97 end up reading the whole file when passed a file as an argument.  If possible, we should try to re-enable that behavior.  However, in the pipeline steps we end up turning lazy loading off anyway since we want to return full, functional, detached objects.

There are warnings in the regression tests related to files not being closed in the regression tests.  Those are also present in main.  They stem from this issue https://github.com/spacetelescope/stpipe/pull/97 and are ultimately unrelated to this PR.

There are minor unrelated changes to group_time and the unit test here https://github.com/spacetelescope/romancal/pull/774/files#diff-0ba58d845e38535dbb7496d66fccd7af20bdb15d3c5baa4fda2f863c90d99ea1R48 which I needed to pass tests.  I found that this was the root cause of the error in https://github.com/spacetelescope/romancal/pull/725/files .

ddavis-stsci:
I'm not sure I'm seeing much improvement in the memory usage. I'm attaching the plot using your PR branch
git status
On branch reduce-copies
memprof_pipeline_July19_1

WilliamJamieson:
Can you provide the code you are using to generate this plot?

schlafly:
I'll look more at this tomorrow, but can you highlight the beginning of the jump detection step here?  jump detection and ramp fitting are still very inefficient and are by far the high water mark.  But with our planned changes to these and this PR those will go down too.  My claims about memory usage focused on three point immediately before jump detection.

ddavis-stsci:
The jump step starts at about the 50s mark. If you look at
https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=SCSB&title=RCAL+pipeline+Timing+and+Memory
you can see that the early steps use 2-4x less memory which is a 
significant improvement.
I assume that the ramp fitting is still the old version for this PR and 
you only refer to the improved version in romanisim?
Do you have a number for the romanisim ramp fitting memory usage?

stscieisenhamer (commented):
LGTM in principle. The open file issue seems to be handled different in roman than jwst; jwst does a reference count method which mitigates many of the issues, if I am reading all this right.

Also note that I agree with the thoughts on future modification of `stpipe`, but such mods may have impact on `jwst` also.

schlafly:
Jonathan has the new version of ramp fitting (related to the simulator version) in his stcal / romancal PRs.  That code doesn't make any copies of the ramp structure.  It does need to make an ImageModel which obviously has a few 4k x 4k arrays of floats.  Probably a GB or two.  Jump detection will be in the same boat.

schlafly:
@ddavis-stsci , can you approve?  It looks like I can't approve my own PR.  Thanks!

ddavis-stsci (approved):

PR #779, opened 2023-07-13
RCAL-511 Inititial implementation of the Uneven Ramp fitting

stscieisenhamer:
Resolves [RCAL-511](https://jira.stsci.edu/browse/JP-511)

This PR implements the the `ols_cas22` algorithm option for uneven/multi-accum table ramp fitting.

The PR is tied to the corresponding `stcal` [PR#175](https://github.com/spacetelescope/stcal/pull/175).

The intent of these PR's is to get the initial algorithm, as implemented in `romanisim` in so that testing, experimenting, and algorithmic improvement/refactoring can occur in subsequent PR's. As such, needed components such as documentation have yet to be filled out.

For reviewers, the intent of these PRs are to identify all future necessary work. Issues found, unless critical, will be documented in issues and handled in future PRs.

**Checklist**
- [x] updated relevant tests
- ~updated relevant documentation~
- [x] added relevant label(s)


romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -7,76 +7,179 @@
...
+        """
+        resultants = input_model.data.value
+        dq = input_model.groupdq
+        read_noise = readnoise_model.data.value
+        gain = gain_model.data.value
+        read_pattern = input_model.meta.exposure.read_pattern
+
+        # add in the gain
+        resultants_adjusted = resultants * gain
+        read_noise_adjusted = read_noise * gain
```
schlafly, line 159: Can we do these in place?  In particular, the resultants_adjusted, which is large?

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -7,76 +7,179 @@
...
+
+        # Fit the ramps
+        ramppar, rampvar = ols_cas22_fit.fit_ramps_casertano(
+            resultants_adjusted, dq, read_noise_adjusted, read_pattern=read_pattern
+        )
+
+        # Break out the information and fix units
+        slopes = ramppar[..., 1]
+        var_rnoise = rampvar[..., 0, 1, 1]
+        var_poisson = rampvar[..., 1, 1, 1]
```
schlafly, line 169: This reminds me that let's go ahead and abandon my decision to structure these arrays unnecessarily largely.  Maybe parallel the stcal OLS output more closely, or something?  But I feel bad that rampvar has 12 different covariances for every pixel, only one of which is actually populated, which are then dropped.  The RampFitInterpolator populated all of these, of course, and maybe we will down the line for Roman as well, but let's not assume that we'll support that now.

schlafly (commented):
This looks good to me.  I've left a few comments in line.  We should merge the data construction code that's been duplicated in test_ramp_fit_ols / test_ramp_fit_cas22?

ddavis-stsci (approved):
LGTM

kmacdonald-stsci (approved):

romancal/ramp_fitting/ramp_fit_step.py
stscieisenhamer: Sorry, not clear as to what is desired? It is that `rampvar` needs to have the [...,1 ,1] dimensions removed? Or is there something more needed? Should the `RampData` object be fully supported, or more supported?

schlafly: Presently these lines
https://github.com/spacetelescope/stcal/pull/175/files#diff-9b179303cde3ddf2b749be9d17d7b44e2f93d1207cb0306e8bd955ad41683ff7R233-R238
build 14 detector planes worth of parameters, but only populate four of those planes.  The rationale was that the other ramp fitting code in the simulator computed all 14 planes, but the Casertano+22 paper only discusses the computation of the four done here.  Since we aren't supporting the other ramp fitting code here, there isn't a reason to make these larger shapes.  So the suggestion is to change the interface there to just return the four par / var terms that are computed, and not to package them in these bigger mostly empty objects and then unpackage them.

It would be virtuous to support RampData, though there are a number of quantities there where I don't actually know what they do.  Let's just stick with some simple tuple or dictionary of four items for now unless you see things differently.

stscieisenhamer:
Due to illness, I have not been able to address the following:
- handle multiplication in-place
- fix the rampvar return

If this needs to be handled sooner, I would suggested merging as-is, and handling the two issues in a separate PR.

stscieisenhamer:
@schlafly One issue that I noticed while simplifying the `var` return. In this code, the `err` is [calculated here](https://github.com/spacetelescope/romancal/blob/73d11024e8c08c5efc416feecffd6b9a963c7593/romancal/ramp_fitting/ramp_fit_step.py#L171).

However, in the `stcal` code, a similar [calculation is done](https://github.com/spacetelescope/stcal/blob/10b8398a590ec747956fb57827f861c9aee4fbb8/src/stcal/ramp_fitting/ols_cas22_fit.py#L138), the only difference is that the romancal code does the square root. Should `err` be what is done in the stcal version, or is the stcal version wrong? Or are they just different?

schlafly:
I think those both look good to me?  Your first link refers to an standard error (err), and your second link refers to a variance (var), and it makes sense that the one is the square root of the other.

stscieisenhamer:
Have modified to point to the `stcal` development. Also fixed the test. Local regressions are clean.

[Full regression here](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/332/).

stscieisenhamer:
OK, clean regression. @ddavis-stsci I have requested a re-review only because I do not know what milestone should be added here.

Otherwise, I believe this is good to go.

PR #787, opened 2023-07-20
Rcal 504 develop the resample step

mairanteodoro:
Resolves [RCAL-504](https://jira.stsci.edu/browse/RCAL-504)

This PR addresses the implementation of the resampling step in RomanCal.

**Notes**
This PR depends on [stcal's PR 196](https://github.com/spacetelescope/stcal/pull/196) (it uses the newly implemented `alignment` sub-package)

**Checklist**
- [X] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     'tweakwcs >=0.8.0',
     'spherical-geometry >= 1.2.22',
     'stsci.imagestats >= 1.6.3',
+    'drizzle >= 1.13.7',
     'webbpsf == 1.1.1',
 ]
 dynamic = ['version']
@@ -46,6 +47,7 @@ docs = [
     'sphinx-automodapi',
     'sphinx-rtd-theme',
     'stsci-rtd-theme',
+    'sphinx-autobuild',
     'tomli; python_version <="3.11"',
 ]
 test = [
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -1,0 +1,592 @@
+import logging
+
+import numpy as np
+from astropy import units as u
+from drizzle import cdrizzle, util
+from roman_datamodels import datamodels
+from roman_datamodels.maker_utils import mk_datamodel, mk_resample
+
+from ..datamodels import ModelContainer
+from . import gwcs_drizzle, resample_utils
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+__all__ = ["OutputTooLargeError", "ResampleData"]
+
+
+class OutputTooLargeError(MemoryError):
+    """Raised when the output is too large for in-memory instantiation"""
+
+
+class ResampleData:
+    """
+    This is the controlling routine for the resampling process.
+
+    Notes
+    -----
+    This routine performs the following operations:
+
+      1. Extracts parameter settings from input model, such as pixfrac,
+         weight type, exposure time (if relevant), and kernel, and merges
+         them with any user-provided values.
+      2. Creates output WCS based on input images and define mapping function
+         between all input arrays and the output array. Alternatively, a custom,
+         user-provided WCS object can be used instead.
+      3. Updates output data model with output arrays from drizzle, including
+         a record of metadata from all input models.
+    """
+
+    def __init__(
+        self,
+        input_models,
+        output=None,
+        single=False,
+        blendheaders=True,
+        pixfrac=1.0,
+        kernel="square",
+        fillval="INDEF",
+        wht_type="ivm",
+        good_bits=0,
+        pscale_ratio=1.0,
+        pscale=None,
+        **kwargs,
+    ):
+        """
+        Parameters
+        ----------
+        input_models : list of objects
+            list of data models, one for each input image
+
+        output : str
+            filename for output
+
+        kwargs : dict
+            Other parameters.
+
+            .. note::
+                ``output_shape`` is in the ``x, y`` order.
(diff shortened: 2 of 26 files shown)


pyproject.toml
```diff
@@ -28,11 +28,12 @@ dependencies = [
     # 'roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git@main',
     'scipy >=1.11',
     # 'stcal >=1.4.0',
-    'stcal @ git+https://github.com/spacetelescope/stcal.git@main',
+    'stcal @ git+https://github.com/mairanteodoro/stcal.git#egg=stcal-alignment',
```
zacharyburnett, line 31:
```suggestion
    'stcal @ git+https://github.com/mairanteodoro/stcal.git@stcal-alignment',
```

romancal/resample/resample.py
```diff
@@ -0,0 +1,569 @@
...
+                inwht,
+                model.meta.wcs,
+                output_wcs,
+                resampled_variance,
+                outwht,
+                outcon,
+                pixfrac=self.pixfrac,
+                kernel=self.kernel,
+                fillval=np.nan,
+            )
```
mcara, line 350: again, `xmin, xmax, ymin, ymax` have been removed. why?

romancal/resample/resample.py
```diff
@@ -0,0 +1,569 @@
...
+        if outcon.ndim == 3:
+            outcon = outcon[planeid]
+
+        if xmin is xmax is ymin is ymax is None:
+            bb = input_wcs.bounding_box
+            ((x1, x2), (y1, y2)) = bb
+            xmin = int(min(x1, x2))
+            ymin = int(min(y1, y2))
+            xmax = int(max(x1, x2))
+            ymax = int(max(y1, y2))
```
mcara, line 539: Now I see why `xmin, xmax, ...` have been dropped from calls to these functions. This code block seems to replicate  JWST's `_resample_range()` but I have questions about this:
1. Is it even possible that `x1 > x2`? (and similar for `y`)
2. [Old code](https://github.com/spacetelescope/jwst/blob/b16e2b78bd601290b633ca74aeaf7383a2c169a5/jwst/resample/resample_utils.py#L315-L319) adds 0.5 pixels before converting `x2` to `int`. That is done to round up so that all input pixels within the bounding box will be resampled. For example, if `x2 = 499.6` then 0.1 (along the `x`-axis) of the 500th pixel has data that could be resampled. With proposed formula you would have excluded that pixel.

romancal/resample/resample.py
```diff
@@ -0,0 +1,569 @@
...
+            for each output pixel. Should be integer zero on first call.
+            Subsequent calls hold intermediate results.  This is modified
+            in-place.
+
+        uniqid : int, optional
+            The id number of the input image. Should be one the first time
+            this function is called and incremented by one on each subsequent
+            call.
+
+        xmin : float, optional
```
mcara, line 453:
```suggestion
        xmin : float, None, optional
```

romancal/resample/resample.py
```diff
@@ -0,0 +1,569 @@
...
+        ymin : float, optional
+            Sets the minimum value in the y dimension on the bounding box. The
+            y dimension varies less rapidly than the x and represents the line
+            index on the input image. If the value is zero, no minimum  will be
+            set in the y dimension.
+
+        ymax : float, optional
+            Sets the maximum value in the y dimension. If the value is zero, no
+            maximum will be set in the y dimension,  the full x dimension
+            of the output image is the bounding box.
```
mcara, line 477: None of these docs explain what happens when `xmin,xmax, ...=None`.

mcara (changes requested):
With exception of how `xmin, xmax, ...=None` are documented and how they are computed from the bounding box, I think this looks OK (i.e., my other comments are minor and optional).

nden (commented):
I think this is ready to be merged after comments are addressed.

nden (changes requested):
The resample step fails when `weight_type='exptime'` with an error:
```
File ~/dev/romancal/romancal/resample/gwcs_drizzle.py:435, in dodrizzle(insci, input_wcs, inwht, output_wcs, outsci, outwht, outcon, expin, in_units, wt_scl, uniqid, xmin, xmax, ymin, ymax, pixfrac, kernel, fillval)
    430 # Call 'drizzle' to perform image combination
    431 log.info(f"Drizzling {insci.shape} --> {outsci.shape}")
    433 _vers, nmiss, nskip = cdrizzle.tdriz(
    434     insci.astype(np.float32).value,
--> 435     inwht.astype(np.float32).value,
    436     pixmap,
    437     outsci,
....
AttributeError: 'numpy.ndarray' object has no attribute 'value'
```

`resample_utils.build_driz_weight` builds a weight image and attaches units to it when `weight_type='ivm'`. However, when `weight_type='exptime'` the weight image is a numpy array. 
Is it necessary to attach units to the weight image for `ivm`? It looks like they are stripped right after that when passed to drizzle and unless I'm missing something there's no other use of units.

mairanteodoro:
> Is it necessary to attach units to the weight image for `ivm`? It looks like they are stripped right after that when passed to drizzle and unless I'm missing something there's no other use of units.

You are correct; I don't think we need to add units for the weighted image. I will update the code to reflect that.

nden:
The problem with `weight_type='exptime'` is fixed. I did not review in detail this time around but had done so previously.

nden (approved):

Issue #793, opened 2023-07-26
Devdeps CI job is failing.


WilliamJamieson:
Between the jobs (for #704):

- https://github.com/spacetelescope/romancal/actions/runs/5661779177/job/15340406708 (PR action run)
- https://github.com/spacetelescope/romancal/actions/runs/5670945339/job/15366991545 (Merge action run)

The devdeps job has begun to fail.

I suspect that this is partially the result of a recent numpy 2.0.dev release.

WilliamJamieson:
Working `pip freeze`
```
py311-devdeps: commands_pre[1]> pip freeze
asdf @ git+https://github.com/asdf-format/asdf@1e9a0497f80fa1a13c496a1727049af5a5ebae5d
asdf-astropy @ git+https://github.com/astropy/asdf-astropy@0c502517a5cca9a6ce03150ad0c6472ae12a2e1f
asdf-coordinates-schemas==0.2.0
asdf-standard==1.0.3
asdf-transform-schemas @ git+https://github.com/asdf-format/asdf-transform-schemas@992fa8c47a68a2e31fe8c350aa0666cc7602f681
asdf-unit-schemas==0.1.0
asdf-wcs-schemas @ git+https://github.com/asdf-format/asdf-wcs-schemas@646bdff47b3d90a378308c118ad4de6e7bd9c96d
astropy @ git+https://github.com/astropy/astropy@d827727cb007ae398da57a1609bfd9944dff08ba
astropy-iers-data==0.2023.7.24.0.33.15
attrs==23.1.0
certifi==2023.7.22
charset-normalizer==3.2.0
ci-watson==0.6.1
coverage==7.2.7
crds==11.17.2
filelock==3.12.2
gwcs @ git+https://github.com/spacetelescope/gwcs@929478d8152574699e57e5498a2e4c079583a1e4
hypothesis==6.82.0
idna==3.4
importlib-metadata==6.8.0
iniconfig==2.0.0
jmespath==1.0.1
jsonschema==4.17.3
numpy==2.0.0.dev0
opencv-python-headless==4.8.0.74
packaging==23.1
Parsley==1.3
photutils==1.8.0
pluggy==1.2.0
psutil==5.9.5
pyerfa==2.0.0.3
pyparsing==3.1.0
pyrsistent==0.19.3
pytest==7.4.0
pytest-arraydiff==0.5.0
pytest-astropy==0.10.0
pytest-astropy-header==0.2.2
pytest-cov==4.1.0
[...]

PR #795, opened 2023-07-26
Actually use the `astropy` nightly wheel.

WilliamJamieson:
Currently, the Jenkins devdeps job is failing because something on the Jenkins machine has changed enough so that it can no longer build `astropy` wheels natively (some sort of compiler/base-library issue). Unlike the previous attempts at using the `astropy` nightly wheel this PR actually uses that wheel.

Also as a workaround to #793, this PR forces `numpy` 2.0.dev+ versions to use the legacy `numpy` 1.25 floating point (print) representation. Note that this will only work around this issue for the unit tests. ASDF will need to resolve this issue in order to have a general solution.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 git+https://github.com/asdf-format/asdf
 git+https://github.com/asdf-format/asdf-transform-schemas
 git+https://github.com/asdf-format/asdf-wcs-schemas
-git+https://github.com/astropy/astropy
+--extra-index-url https://pypi.anaconda.org/astropy/simple astropy --pre
 git+https://github.com/astropy/asdf-astropy
 git+https://github.com/spacetelescope/crds
 git+https://github.com/spacetelescope/gwcs
--- a/romancal/conftest.py
+++ b/romancal/conftest.py
@@ -3,7 +3,15 @@ import inspect
 import os
 import tempfile
 
+import numpy as np
 import pytest
+from astropy.utils import minversion
+
+# HACK: This is a temporary workaround for ASDF not being able to handle how
+#       numpy 2.0.dev+ represents (prints) floating point numbers. This simply
+#       forces numpy to use the old printing method while running the tests only.
+if minversion(np, "2.0.dev"):
+    np.set_printoptions(legacy="1.25")
 
 
 @pytest.fixture
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,6 +60,8 @@ general
 
 - Make steps operate in place rather than copying.  [#774]
 
+- Fix devdeps Jenkins job. [#795]
+
 0.11.0 (2023-05-31)
 ===================
 


WilliamJamieson:
Proof that it is now using a prebuilt dev wheel: https://github.com/spacetelescope/romancal/actions/runs/5674451888/job/15378017238#step:10:269

WilliamJamieson:
This passes the regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/298/

Note that I did not run the Roman-devdeps tests on this because that would require me to modify the configuration of that test suite. Instead I modified the Roman-Developers-Pull-Requests job to both point at this version of `romancal` AND use the `JenkinsfileRT_dev` instead of its normal `JenkinsfileRT`.

romancal/conftest.py
```diff
@@ -3,7 +3,15 @@
...
 
+import numpy as np
 import pytest
+from astropy.utils import minversion
+
+# HACK: This is a temporary workaround for ASDF not being able to handle how
+#       numpy 2.0.dev+ represents (prints) floating point numbers. This simply
+#       forces numpy to use the old printing method while running the tests only.
+if minversion(np, "2.0.dev"):
+    np.set_printoptions(legacy="1.25")
```
WilliamJamieson, line 14: This has been added only so that the `devdeps` CI and regression tests can be demonstrated to fully work by covering over a known issue in ASDF.

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 git+https://github.com/asdf-format/asdf
 git+https://github.com/asdf-format/asdf-transform-schemas
 git+https://github.com/asdf-format/asdf-wcs-schemas
-git+https://github.com/astropy/astropy
+--extra-index-url https://pypi.anaconda.org/astropy/simple astropy --pre
```
WilliamJamieson, line 4: The extra-index-url listed in the tox.ini really ought to be moved here.

zacharyburnett (approved):
LGTM, sorry this has been a runaround for so long. I'll work on removing `--upgrade-strategy eager` as well as that has seemed to cause more problems than its worth

ddavis-stsci (changes requested):
I think this is a bit premature. If Jenkins is having trouble building astropy wheels shouldn't we file an SPB ticket and if they can't fix it then do the workaround?

WilliamJamieson:
> I think this is a bit premature. If Jenkins is having trouble building astropy wheels shouldn't we file an SPB ticket and if they can't fix it then do the workaround? 

This is not a work around. Astropy builds a new dev wheel every week or so and uploads it here.

It is my understanding that @nden wanted us to move all the pipeline development testing from building wheels ourselves to using the prebuilt development wheels. I do not think that we should be in the business of attempting to support building libraries that we do not have to.

nden:
I think the goal is to test working with the latest available dev version of astropy. We can assume they are testing their build system. Let's use the wheels for packages where available to speed up testing, since what we are testing is new functionality changes in dependencies, not their build systems. This is not a new way of running the devdeps tests afaik.

pllim:
Hi, `astropy` core maintainer here. 👋 

It is my advice that you use the dev wheel that we provide instead of trying to build it yourself. We generate wheels for all the major architectures on versions that we support. If you switch to using our wheel and the system is unable to find one, it might be an indication that you have to upgrade stuff on your end.

In addition, using wheel better reflects how your users will actually run the whole stack. No one sane is going to try to build `astropy` from source downstream because the package has both C-extensions and Cython codes.

Hope this helps!

PR #802, opened 2023-08-01
Rcal 596 Add association processing to the pipeline code

### LINKED ISSUE #755: Add association processing to exposure pipeline
_Issue [RCAL-596](https://jira.stsci.edu/browse/RCAL-596) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

We now have a container for association files. Add the ability to process the members of an association file to the exposure level pipeline. Initially this should process all of the 18 detectors in the FOV to create 18 calibrated detector files.

ddavis-stsci:
Resolves [RCAL-596](https://jira.stsci.edu/browse/rcal-596)

Closes #755 

This PR adds the initial association processing to the pipeline. This is an initial implementation and once rcal-620 and rcal-621 are done we will use ModelContainer and the open function to load the association. 
The current usage requires the use of disable-crds-steppar option but should be relaxed on the two Jira issues above are 
addressed. To run the pipeline with an association the command is, 
strun --disable-crds-steppar roman_elp r90001-a3001_image_001_asn.json
```
2023-08-01 13:51:57,520 - stpipe.ExposurePipeline - INFO - ExposurePipeline instance created.
2023-08-01 13:51:57,521 - stpipe.ExposurePipeline.dq_init - INFO - DQInitStep instance created.
2023-08-01 13:51:57,522 - stpipe.ExposurePipeline.saturation - INFO - SaturationStep instance created.
...
2023-08-01 14:11:21,769 - stpipe.ExposurePipeline.source_detection - INFO - Found 3869 sources.
2023-08-01 14:11:21,802 - stpipe.ExposurePipeline.source_detection - INFO - Step source_detection done
2023-08-01 14:11:21,803 - stpipe.ExposurePipeline - INFO - Roman exposure calibration pipeline ending...
2023-08-01 14:11:23,380 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI01_cal.asdf
2023-08-01 14:11:24,925 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI02_cal.asdf
2023-08-01 14:11:26,485 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI03_cal.asdf
2023-08-01 14:11:28,130 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI04_cal.asdf
2023-08-01 14:11:28,130 - stpipe.ExposurePipeline - INFO - Step ExposurePipeline done
```

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

pytest romancal/tests/
=============================================================== test session starts ================================================================
platform darwin -- Python 3.11.0, pytest-7.4.0, pluggy-1.2.0
rootdir: /Users/ddavis/src/Roman/romancal
configfile: pyproject.toml
plugins: remotedata-0.4.0, asdf-2.15.0, hypothesis-6.82.0, doctestplus-0.13.0, cov-4.1.0, filter-subpackage-0.1.2, mock-3.11.1, astropy-header-0.2.2, astropy-0.10.0, ci-watson-0.6.1, openfiles-0.5.0, env-0.8.2, arraydiff-0.5.0
collected 92 items

romancal/tests/test_import.py ............................................................................................                   [100%]

================================================================ 92 passed in 1.59s ================================================================
(rcal_dev) bash>pytest romancal/datamodels/tests/
=============================================================== test session starts ================================================================
platform darwin -- Python 3.11.0, pytest-7.4.0, pluggy-1.2.0
rootdir: /Users/ddavis/src/Roman/romancal
configfile: pyproject.toml
plugins: remotedata-0.4.0, asdf-2.15.0, hypothesis-6.82.0, doctestplus-0.13.0, cov-4.1.0, filter-subpackage-0.1.2, mock-3.11.1, astropy-header-0.2.2, astropy-0.10.0, ci-watson-0.6.1, openfiles-0.5.0, env-0.8.2, arraydiff-0.5.0
collected 42 items

romancal/datamodels/tests/test_datamodels.py .........................................                                                       [ 97%]
romancal/datamodels/tests/test_filetype.py .                                                                                                 [100%]

=============================================================== 42 passed in 23.78s

Since plwishmaster is down I ran the regression tests locally, 
platform darwin -- Python 3.11.0, pytest-7.4.0, pluggy-1.2.0
rootdir: /Users/ddavis/src/Roman/romancal
configfile: pyproject.toml
plugins: remotedata-0.4.0, asdf-2.15.0, hypothesis-6.82.0, doctestplus-0.13.0, cov-4.1.0, filter-subpackage-0.1.2, mock-3.11.1, astropy-header-0.2.2, astropy-0.10.0, ci-watson-0.6.1, openfiles-0.5.0, env-0.8.2, arraydiff-0.5.0
collected 21 items

../../../romancal/romancal/regtest/test_dark_current.py ....             [ 19%]
../../../romancal/romancal/regtest/test_jump_det.py .                    [ 23%]
../../../romancal/romancal/regtest/test_linearity.py ..                  [ 33%]
../../../romancal/romancal/regtest/test_ramp_fitting.py .                [ 38%]
../../../romancal/romancal/regtest/test_refpix.py .                      [ 42%]
../../../romancal/romancal/regtest/test_tweakreg.py .                    [ 47%]
../../../romancal/romancal/regtest/test_wfi_dq_init.py ..                [ 57%]
../../../romancal/romancal/regtest/test_wfi_flat_field.py ...            [ 71%]
../../../romancal/romancal/regtest/test_wfi_photom.py .                  [ 76%]
../../../romancal/romancal/regtest/test_wfi_pipeline.py ...              [ 90%]
../../../romancal/romancal/regtest/test_wfi_saturation.py ..             [100%]
...
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================= 21 passed, 19 warnings in 7665.81s (2:07:45)


braingram:
FYI: doc failure appears related to an old theme version, similar to: https://github.com/spacetelescope/jwst/pull/7787/files

ddavis-stsci:
Thanks Brett, that fixed the problem!

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,95 @@ def process(self, input):
...
                 # Return fully saturated image file (stopping pipeline)
-                return result
-
-        result = self.assign_wcs(result)
-        if result.meta.exposure.type == "WFI_IMAGE":
-            result = self.flatfield(result)
-        else:
-            log.info("Flat Field step is being SKIPPED")
-            result.meta.cal_step.flat_field = "SKIPPED"
+                return results
```
schlafly, line 147: I'm confused about this.  The old version had this line under the is_fully_saturated condition; here this is outside that condition, so if there is ' groupdq' in result.keys() we stop the pipeline?  I think I would have guessed that this should be indented one block, and that rather than returning, we continue to the next SCA?

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,95 @@ def process(self, input):
...
+            result = self.assign_wcs(result)
+            if result.meta.exposure.type == "WFI_IMAGE":
+                result = self.flatfield(result)
+            else:
+                log.info("Flat Field step is being SKIPPED")
+                result.meta.cal_step.flat_field = "SKIPPED"
 
-        if result.meta.exposure.type == "WFI_IMAGE":
+            result = self.photom(result)
             result = self.source_detection(result)
```
schlafly, line 157: This might be a merge / rebase issue or something, but I think we want to keep this "only do source detection for images" condition?

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,95 @@ def process(self, input):
...
-            result.meta.cal_step.source_detection = "SKIPPED"
 
-        # setup output_file for saving
-        self.setup_output(result)
-        log.info("Roman exposure calibration pipeline ending...")
+            # setup output_file for saving
+            self.setup_output(result)
+            log.info("Roman exposure calibration pipeline ending...")
+
+            self.output_use_model = True
```
schlafly, line 163: I didn't see where this was used.

schlafly (commented):
Looks good to me.  I left a few comments in line in the pipeline stage where I didn't understand the logic.

Do the regression tests really take 2 hours to run on your system?  Is that just the difference between your system and plwishmaster (~20 min, IIRC?)?

Yes, Mostly downloading lots of large files.

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: Yes that must have been wiped out. 
The check is also in the steps so it doesn't matter too much. The logic is here to avoid extra setup.

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: That looks like a rebase/merge issue and I'll fix that.

ddavis-stsci:
Updates for logging and step skips for non WFI_IMAGE data. The pipeline now produces,
```
2023-08-03 14:21:13,575 - stpipe.ExposurePipeline - INFO - Flat Field step is being SKIPPED
2023-08-03 14:21:13,576 - stpipe.ExposurePipeline - INFO - Photom and source detection steps are being SKIPPED
2023-08-03 14:21:13,576 - stpipe.ExposurePipeline - INFO - Roman exposure calibration pipeline ending...
2023-08-03 14:21:15,180 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI01_cal.asdf
2023-08-03 14:21:16,635 - stpipe.ExposurePipeline - INFO - Saved model in r0000201001001001003_01101_0001_WFI01_cal.asdf
2023-08-03 14:21:16,635 - stpipe.ExposurePipeline - INFO - Step ExposurePipeline done
```

and 

```
          assign_wcs	COMPLETE
          flat_field	SKIPPED
                dark	COMPLETE
             dq_init	COMPLETE
                jump	COMPLETE
           linearity	COMPLETE
              photom	SKIPPED
    source_detection	SKIPPED
            ramp_fit	COMPLETE
          saturation	COMPLETE
```

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,100 @@ def process(self, input):
...
-                    "source_detection",
-                ]:
-                    result.meta.cal_step[step_str] = "SKIPPED"
-
-                # Set suffix for proper output naming
-                self.suffix = "cal"
+                log.info("All pixels are saturated. Returning a zeroed-out image.")
+
+                # Return zeroed-out image file (stopping pipeline)
+                return self.create_fully_saturated_zeroed_image(result)
```
schlafly, line 123: I think there's still something messed up about the fully saturated logic.  Here we check if the image is saturated and if so return.  I think we want to instead do:
results.append(self.create_fully_saturated_zeroed_image(result))
continue
?  Maybe additionally adding 'SKIPPED' everywhere as is done a few lines later.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,100 @@ def process(self, input):
...
+                        "assign_wcs",
+                        "flat_field",
+                        "photom",
+                        "source_detection",
+                    ]:
+                        result.meta.cal_step[step_str] = "SKIPPED"
+
+                    # Set suffix for proper output naming
+                    self.suffix = "cal"
+                    results.append(result)
```
schlafly, line 144: It feels to me like this block never executes, since it's below the return self.create_fully_saturated... bit above.  I suspect you want to get rid of this block and incorporate the 'SKIPPED' bit into the earlier block?

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: I agree but I think this is separate ticket from the association implementation. 
FWIW I disagree with adding the saturation code check here but ...

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: See above

romancal/associations/load_as_asn.py
```diff
@@ -0,0 +1,207 @@
...
+        Along with the attributes belonging to a Level2 association, the
+        following are added:
+
+        filename : str
+            The name of the association file, if such a file
+            where passed in. Otherwise a default value is given.
+
+        Returns
+        -------
+        association : DMSLevel2bBase
```
nden, line 131: Should we get away from JWST names and levels, since there's noLevel 2b for Roman?

romancal/associations/load_as_asn.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        product_name_func = cls.model_product_name
+        if basename is not None:
+            product_name_func = partial(cls.name_with_index, basename)
+
+        # if the input string is a FITS file create an asn and return
+        if isinstance(obj, str):
+            file_name, file_ext = os_path.splitext(obj)
+
+            if file_ext == ".fits":
```
nden, line 142: And what about not using FITS files or do you see this code being combined in the future with the code for JWST?

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,100 @@ def process(self, input):
...
-
-        # Test for fully saturated data
-        if "groupdq" in result.keys():
+        file_type = filetype.check(input)
+        asn = None
+        if file_type == "asdf":
+            try:
+                input = rdm.open(input)
+            except TypeError:
+                log.debug("Error opening file:")
```
nden, line 84: What should happen in the case of one asdf file which cannot be opened? Should it return?

romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,66 +75,100 @@ def process(self, input):
...
+        if file_type == "asn":
+            try:
+                asn = LoadAsLevel2Asn.load(input, basename=self.output_file)
+            except AssociationNotValidError:
+                log.debug("Error opening file:")
```
nden, line 89: Similarly here - if an association is invalid shouldn't the code exit?

nden: Is there a way to make sure it's a Level 2 asn, to prevent running the code on L2 products? Even if this is desirable I'm not asking to be done in this PR.

romancal/pipeline/exposure_pipeline.py
schlafly: I hear you that some of this could be a different PR, but at least the bit where we still want to process all of the images even if the first one is all saturated applies?  i.e., replace return self.create_fully_saturated_zeroed_image with results.append(self.create_fully_saturated_zeroed_image) ; continue, right?  The step should always return a list of results if fed an association, and never just one?

romancal/associations/load_as_asn.py
ddavis-stsci: That's fine. I did this for expediency and when the associations were much more poorly defined. Can this be a separate ticket to clean this up?

romancal/associations/load_as_asn.py
ddavis-stsci: I kept this to cover a case if/when someone will give romancal a fits file. We can just give a warning and gracefully exit.
There is also the type error in roman_datamodels (TypeError: Roman datamodels does not accept FITS files or objects) should we stick with that?

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: I guess the question is, if any of the inputs in the association fail for any reason will the end user still want the results? 
My choice would be to stop and let the user know. They can fix the error and rerun the pipeline. That way they make the decision if the results are valid without all the inputs. Not very pythonic of me.

romancal/pipeline/exposure_pipeline.py
ddavis-stsci: I left all that out. I'd like to address that once rcal-620 and 621 are in. I think that would be a better place to enforce validity.

ddavis-stsci:
That is not actually true. In some cases you will only want one result.
For this case you should get a list of results.
Updated the code.

schlafly (approved):
I'm still confused about the intended behavior in the fully saturated case.  I've added code suggestions below and am approving.  Thanks!

ddavis-stsci:
I don't understand your objections. When there is a fully saturated 
input file I get

2023-08-07 10:31:44,768 - stpipe.ExposurePipeline.source_detection - 
INFO - Step source_detection done
2023-08-07 10:31:44,768 - stpipe.ExposurePipeline - INFO - Roman 
exposure calibration pipeline ending...
2023-08-07 10:31:46,735 - stpipe.ExposurePipeline - INFO - Saved model 
in r0000101001001001001_01101_0001_WFI01_cal.asdf
2023-08-07 10:31:48,481 - stpipe.ExposurePipeline - INFO - Saved model 
in r0000101001001001001_01101_0001_WFI01_sat_cal.asdf
2023-08-07 10:31:50,110 - stpipe.ExposurePipeline - INFO - Saved model 
in r0000101001001001001_01101_0001_WFI02_cal.asdf
2023-08-07 10:31:50,110 - stpipe.ExposurePipeline - INFO - Step 
ExposurePipeline done

So the unsaturated data is calibrated and output and the saturated data 
output is also output.
and the saturated data file has 0's as expected.
np.max(sat.data)
<Quantity 0. electron / s>

schlafly:
Okay, I guess I'm misunderstanding the code, go ahead.

pyproject.toml
```diff
@@ -22,10 +22,11 @@ dependencies = [
     'photutils >=1.6.0',
     'pyparsing >=2.4.7',
     'requests >=2.22',
-    # 'rad >=0.15.0',
-    'rad @ git+https://github.com/spacetelescope/rad.git@main',
-    # 'roman_datamodels >=0.15.0',
-    'roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git@main',
+    'rad == 0.17.0',
```
braingram, line 25: I think this pin is causing the devdeps job on plwishmaster to pull the older, 0.17.0 rad (and similar for roman_datamodels) instead of testing against the 0.17.1 (or development version, which I think is the same at the moment).

@ddavis-stsci was this a requirement for this PR? If so, is there a follow-up PR or tracking issue to remove the pin?

ddavis-stsci:
The pin is needed until we get new input files that match the latest RAD 
and roman_datamodels.
Once we have that we'll switch back to the latest released version of 
both packages.

braingram:
Thanks! That explains why 22 of the 23 failing tests fixed themselves :)

pyproject.toml
```diff
@@ -22,10 +22,11 @@ dependencies = [
...
+#    'rad >= 0.16.0, < 0.17.1',
+#    'rad @ git+https://github.com/spacetelescope/rad.git@main',
+    'roman_datamodels ==0.17.0',
+#    'roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git@main',
```
WilliamJamieson, line 29: There absolutely should not be exact pins for these dependencies. This will cause us major headaches in development.

PR #804, opened 2023-08-01
Adjust ramps for gain, a la JWST's GainScaleStep.

schlafly:
Resolves [RCAL-608](https://jira.stsci.edu/browse/RCAL-608)

Comparisons between romanisim and romancal indicated that there was a flux normalization issue relating to the gain.  I think that the issue is that the JWST code that the Roman code derives from has a separate GainScaleStep that adjusts the ramps for the gain.  Meanwhile the Roman code does not include this factor.

This adjusts the slopes and corresponding uncertainties and variances to account for this factor.

I have not implemented a separate step, and have instead just updated the relevant arrays in the ramp fit step.

Since the factor occurs in ramp fitting, in a location we expect to replace soon, it's not clear we need this.  On the other hand, https://github.com/spacetelescope/romancal/pull/779 optionally preserves the existing implementation, so we should fix it if we want to preserve it there.

I have not yet run regression tests---they will surely fail.  I can update the relevant documentation if we decide we want to go this way.  @ddavis-stsci , thoughts?

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


ddavis-stsci:
I'm not sure I understand completely. The JWST instruments have the 
ability to adjust the gain and thus they also
need the gain scale step. As I understand it the Roman WFI does not have 
the ability to have the gain adjusted by
command so we should not need the step to adjust the gain factor, it is 
set in the calibration files.

I'll take a closer look at the ramp fitting and see what they are doing 
there.

schlafly:
I can at least say that this resolves the difference between romanisim and romancal.  I read this and the Webb step as straightforwardly applying the gain, though I did see notes in the Webb implementation that it was for "nonstandard" gains.

I don't know what it means for the gain to be applied in the calibration files (for me, that's what the gain reference file is).  But I didn't verify experimentally that the stcal ramp fitting step was preserving the units it was passed (i.e, DN in -> DN/s out).  I can check that tomorrow.

schlafly:
I've run an additional bit of test code to verify that ramp_fit.ramp_fit(...) doesn't apply the gain.
```
gainvals = [1, 10, 100]
for gv in gainvals:
    gain.data.value[...] = gv * u.electron / u.DN
    image_info, integ_info, opt_info, gls_opt_model = ramp_fit.ramp_fit(
        im, ramp_fit.BUFSIZE, False, rn.data.value, gain.data.value,
        'ols', 'optimal', 1, romancal.lib.dqflags.pixel)
    print(gv, np.mean(image_info[0]))
```
which gives the following output:
```
1 0.041118436
10 0.04111842
100 0.04111842
``` 
In this test, changing the gain by a factor of a hundred while leaving the DN constant gives identical slopes.  Here the ramps all look like [0, 1, 2, 3, 4, 5], and the frame time is 3.04 and the number of reads per group is 8, so we expect an answer of 1 / 3.04 / 8 = 0.0411842 DN / s, which is what we're getting.  As far as I can tell, we're being given the answer in DN / s and we need to multiply by the gain in order to get to electron / s.

You're right that I haven't figured out where JWST is applying the spatially variable part of the gain; I hadn't appreciated that GainScaleStep just applies a scalar constant.  I did also do a test with half of the gain image having one value and the other half with another value, and still got 0.0411842 DN / s out everywhere.

ddavis-stsci:
I misspoke above and you are correct that JWST just scales the data by a numerical factor (GAINFACT) along with the err, var_poisson, var_rnoise. So the 2-d gain should be used just like for Roman in calculating the ramps. So a DC offset to the ramp data should not alter the derived slopes. 

For the gls fit routine they just use the gain as 
`data *= gain_2d`
when this is passed to the determine_slope routine.

schlafly:
Sorry, are you saying this PR should go in?  I haven't looked at the gls routines at all (romancal uses the ols ones?).  It is good to see the clear treatment of this in the gls module.  That said, I am missing where jwst is applying the spatially variable part of the gain.  Perhaps gain variation gets folded into the flat field?  In that case the simulator isn't quite doing the right thing in its application of the flat field (it would be applying the gain twice), but since the reference files aren't realistic this isn't a real issue at present.

schlafly:
@ddavis-stsci , following up on this one, I think we decided we wanted it to go in soon-ish before the next round of regression test file updates.

ddavis-stsci (commented):
I think this is correct and if the results are to be in electrons this needs to be applied.

tddesjardins (approved):
LGTM

schlafly:
@ddavis-stsci , can you approve this?  Thanks!

ddavis-stsci (approved):
LGTM

PR #808, opened 2023-08-08
Redo pinning for regression testing

### LINKED ISSUE #807: Avoid exact pins for `rad` and `roman_datamodels`
PR #802 exact pins `rad` and `roman_datamodels`. This is disruptive to development of these packages in conjunction with `romancal`.

WilliamJamieson:
Closes #807

This PR adds a new requirements file to be used when performing non-devdeps regression testing called `requirements-reg.txt`. This file enables us to pin dependencies for regression testing purposes but which will not affect normal pip installs. For example it will enable adding an upper pin on `rad` if it contains changes which will temporally break the regression tests due to the need to regenerate the regression test files (such as when we make structural changes to the files). Normally, `requirements-reg.txt` should be empty, only with temporary entries to smooth regression testing; once regression test files have been updated `requirements-dev.txt` should then be emptied.

Note that this will NOT affect running the regression tests locally. If you wish to employ it locally you will need to run
```
pip install -r requirements-reg.txt
```

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)


ddavis-stsci (changes requested):
Please don't make any changes until after the release.

WilliamJamieson:
> Please don't make any changes until after the release.

If the release is supposed to include the `rad` 0.17.1 changes then this will need to go in to fix the photom test and pins.

PR #836, opened 2023-08-16
Adjust saturation threshold depending on the read pattern

schlafly:
Resolves [RCAL-618](https://jira.stsci.edu/browse/RCAL-618)

This PR in concert with https://github.com/spacetelescope/stcal/pull/188 updates the saturation threshold to depend on the read pattern.

The idea is that the saturation reference file gives the number of counts at which a given pixel saturates.  Roman downlinks resultants, which are usually averages of many reads.  The current saturation flagging checks if a resultant is above the saturation threshold, but we want to know if any read in the resultant is above the threshold.  The resultants in a sense dilute the saturated signal---a saturated last read is averaged with unsaturated earlier reads, bringing the resultant value beneath the saturation threshold.  This PR corrects for that affect by passing stcal an additional argument, read_pattern, that specifies how reads are allocated to resultants, and using that information to adjust the saturation threshold resultant-by-resultant.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


schlafly:
Regression tests are failing for unrelated reasons (presumably the forthcoming release?), though more broadly I think we expect the change to the saturation algorithm to introduce some differences in the saturation flagging.

schlafly:
I think this is good to go, though I doubt we actually want to merge it until the corresponding stcal PR is merged.

ddavis-stsci (approved):
LGTM, but of course the real work is in stcal.

PR #841, opened 2023-08-21
Source detection step option to fit model PSFs

### LINKED ISSUE #830: Add PSF fitting option to Source Detection step
_Issue [RCAL-636](https://jira.stsci.edu/browse/RCAL-636) was created on JIRA by [Brett Morris](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bmmorris):_

[romancal#&#x2060;794](https://github.com/spacetelescope/romancal/pull/794) (RCAL-609) implements PSF fitting methods for accurate astrometry. Add an option to the Source Detection step to refine astrometry for sources detected using solutions from PSF fitting.

bmorris3:
### Description

This PR improves astrometry in the Source Detection step by (optionally) making use of PSF fitting methods added in https://github.com/spacetelescope/romancal/pull/794.

The centroid accuracy of the PSF fitting methods depends on many factors, like the precision of the PSF model used in the fit (which may be traded off for longer runtimes), the flux of the source, the WFI filter, and to a lesser extent, the detector position, etc. The revised tests for source detection in this PR check the following criteria when using PSF fitting on bright sources in the F087 filter: 
* the median centroid residual is smaller than 3 milliarcseconds
* the maximum centroid residual is smaller than 11 milliarcseconds
* the recovered centroids and their errors are all <2-sigma consistent with the injected values

(see [here in the diff](https://github.com/spacetelescope/romancal/pull/841/files#diff-3e72b9b93885321af310a4c1d1641f249282c6b943e1c1d0b53309b4206c7ebbR57-R61) for these assertions).

These upper-limits in the tests can be revised when we:
1. Precompute PSF models and make them available (e.g. on CRDS). With more accurate (and more expensive to compute) PSF models, we can get more precise centroids, and revise the upper limits in the tests downwards.
2. Have more specific requirements for a given filter and/or a given source flux

This PR supersedes #936.

Resolves [RCAL-609](https://jira.stsci.edu/browse/RCAL-609)

Closes #830
Closes #936

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/source_detection/source_detection_step.py
+++ b/romancal/source_detection/source_detection_step.py
@@ -5,6 +5,7 @@ Create a source catalog for tweakreg
 
 import logging
 
+import astropy.units as u
 import numpy as np
 from asdf import AsdfFile
 from astropy.stats import SigmaClip
@@ -35,7 +36,7 @@ class SourceDetectionStep(RomanStep):
     """
 
     spec = """
-        kernel_fwhm = float(default=2.)  # DAOStarFinder:Size of Gaussian kernel,
+        kernel_fwhm = float(default=None)  # DAOStarFinder:Size of Gaussian kernel,
         # in pixels.
         sharplo = float(default=0.)  # DAOStarFinder: Lower bound for sharpness.
         # Typical values of sharpness range from 0 (flat) to 1 (delta function).
@@ -55,7 +56,7 @@ class SourceDetectionStep(RomanStep):
         # the SNR for the threshold image.
         bkg_estimator = string(default='median')  # if calc_threshold_img,
         # choice of mean, median, or mode.
-        bkg_boxsize = integer(default=3)  # if calc_threshold_img,
+        bkg_boxsize = integer(default=9)  # if calc_threshold_img,
         # size of box in pixels for 2D background.
         bkg_sigma = float(default=2.0) # if calc_threshold_img,
         # n sigma for sigma clipping bkgrnd.
@@ -89,6 +90,8 @@ class SourceDetectionStep(RomanStep):
                 (dqflags.pixel["DO_NOT_USE"]) & input_model.dq
             ).astype(bool)
 
+            filt = input_model.meta.instrument["optical_element"]
+
             # if a pre-determined threshold value for detection for the whole
             # image is provided, use this
             if self.scalar_threshold is not None:
@@ -105,8 +108,22 @@ class SourceDetectionStep(RomanStep):
                 threshold_img = bkg.background + self.snr_threshold * bkg.background_rms
                 threshold = np.median(threshold_img)
                 log.info(f"Calculated a detection threshold of {threshold} from image.")
+            else:
+                threshold = np.median(self.data)
 
-            log.info("Detecting sources with DAOFind, using entire image array.")
+            if self.kernel_fwhm is None:
+                # estimate the FWHM of the PSF using rough estimate
+                # from the diffraction limit:
+                central_wavelength = psf.filter_central_wavelengths[filt] * u.um
+                diffraction_limit = float(1.22 * central_wavelength / (2.4 * u.m))
+                assume_pixel_scale = 0.11 * u.arcsec / u.pix
+                expect_psf_pix = np.sqrt(
+                    ((diffraction_limit * u.rad).to(u.arcsec) / assume_pixel_scale) ** 2
+                    + 1 * u.pix**2
+                ).to_value(u.pix)
+                self.kernel_fwhm = expect_psf_pix
+
+            log.info("Detecting sources with DAOStarFinder, using entire image array.")
             daofind = DAOStarFinder(
                 fwhm=self.kernel_fwhm,
                 threshold=threshold,
@@ -125,6 +142,8 @@ class SourceDetectionStep(RomanStep):
             elif self.calc_threshold is not None:
                 # subtract background from data if calculating abs. threshold
                 sources = daofind(self.data - bkg.background, mask=self.coverage_mask)
+            else:
+                sources = daofind(self.data, mask=self.coverage_mask)
 
             # reduce table to minimal number of columns, just source ID,
             # positions, and fluxes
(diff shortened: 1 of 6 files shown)


romancal/source_detection/source_detection_step.py
```diff
@@ -122,8 +140,10 @@ def process(self, input):
                 sources = daofind(self.data, mask=self.coverage_mask)
 
             elif self.calc_threshold is not None:
-                # subtrack background from data if calculating abs. threshold
+                # subtract background from data if calculating abs. threshold
                 sources = daofind(self.data - bkg.background, mask=self.coverage_mask)
+            else:
+                sources = daofind(self.data, mask=self.coverage_mask)
```
schlafly, line 146: Staring a little at the threshold logic, I'm confused.  daofind doesn't do it natively, I think, but morally one wants to find all point sources that are more than N sigma above background, for N ~ 5.  That looks potentially like what 'calc_threshold' mode is doing, but there it looks like the background is both part of the threshold and subtracted from the data, which seems like doubly subtracting.  Meanwhile the mode without calc_threshold doesn't look like it has a notion of RMS and instead only has a notion of background, and the threshold is set exactly at the background?

I'm not sure there are any cases where we don't want to measure and remove a background, so I think I'd recommend just removing the calc_threshold = False branch, and then fixing the calc_threshold = True branch so that doesn't ~doubly subtract the background?

Sorry, I realize that this isn't new code, etc., but we should still fix it.

romancal/source_detection/source_detection_step.py
```diff
@@ -140,17 +160,54 @@ def process(self, input):
...
+                # meta.source_detection the table will be stored as a
+                # 1D array with the four columns concatenated, in order,
+                # with units attached
+                catalog_as_array = np.array(
+                    [
+                        catalog["id"].value,
+                        catalog["xcentroid"].value,
+                        catalog["ycentroid"].value,
+                        catalog["flux"].value,
+                    ]
```
schlafly, line 209: I don't know how hard it is to check, and maybe we defer this to another PR, but we really want to use at least a structured array here eventually, as well as saving all the other columns that may have value.

romancal/source_detection/source_detection_step.py
```diff
@@ -185,6 +242,11 @@ def process(self, input):
                 # and save_catalogs is false, since it is not in the schema
                 input_model.meta.source_detection["tweakreg_catalog"] = catalog_as_array
 
+                if self.fit_psf:
+                    input_model.meta.source_detection[
+                        "psf_catalog"
+                    ] = psf_photometry_table
+
```
schlafly, line 249: I think ultimately we want one L2 source catalog for each image with a ~fixed schema.  That would argue for merging the PSF & initial finder results together---they're row-by-row matched, right?  Could also be part of a separate PR.

schlafly (commented):
This looks good.  Have you run any romanisim or regtest simulated images?  I'm still worried a bit about the DAOStarFinder options, and it would be good to see that we detect most of the real, bright sources.

romancal/source_detection/source_detection_step.py
bmorris3: I honestly did not understand what was happening in this block either. I added the `else` block that you've commented on here to make this PR work, but I agree that the logic needs revision.

romancal/source_detection/source_detection_step.py
bmorris3: I will ping @PaulHuwe on Slack to ask about this.

bmorris3: I confirmed that structured arrays are supported in ASDF. You can test for yourself with:
```python
import numpy as np
import roman_datamodels.datamodels as rdd
from roman_datamodels.maker_utils import mk_level2_image

im = rdd.ImageModel(mk_level2_image())

idx = np.random.randint(0, 100, 10)
x = np.random.uniform(0, 100, 10)
y = np.random.uniform(0, 100, 10)

recarr = np.core.records.fromarrays(
    [idx, x, y], 
    names="idx, x, y", 
    formats=[int, float, float]
)

im.meta['test_recarr'] = recarr

im.to_asdf('test_recarr.asdf')

im_reloaded = rdd.open('test_recarr.asdf')

recarr2 = im_reloaded.meta['test_recarr']

assert np.all(recarr2 == recarr)
```

I propose that we make a follow-up PR for the move to structured arrays, since that will touch tweakreg a bunch. Thoughts?

schlafly: That sounds good, thanks.

romancal/source_detection/source_detection_step.py
bmorris3: I agree that we should have all results (source detection and optionally PSF) reported in one table. Since tweakreg currently expects an unstructured array, I won't implement the update in this PR (as discussed in https://github.com/spacetelescope/romancal/pull/841#discussion_r1386942201).

In the follow-up PR, we can have a structured array with columns that tweakreg uses, which will be the PSF fitting results if they're available and the DAO results if not, and also extra columns that tweakreg ignores.

romancal/source_detection/source_detection_step.py
PaulHuwe: As long as no one cares about accessing this information as text in the metadata, then this is great.

bmorris3: By default, this catalog is later deleted by tweakreg, so it shouldn't be useful to anyone unless they stop the pipeline part-way through.

schlafly: We definitely don't want a potentially huge catalog converted to text anywhere!

bmorris3:
@schlafly: All related [regression tests](https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/469/#showFailuresLink) are passing (failures are unrelated).

schlafly (approved):
Great, looks good to me, thanks!

braingram:
@bmorris3 Do you recall the reasoning for the webbpsf exact pin?
https://github.com/spacetelescope/romancal/blob/0ad965cab7bba1247b9ea3d2be01209ef14fc6ed/pyproject.toml#L38
A new version (1.3.0) is available which appears to be compatible with numpy 2.0 (the version pinned here is not).

PR #843, opened 2023-08-21
Cleanup the requirements-dev

WilliamJamieson:
Update `requirements-dev.txt` so that it contains all the information it needs instead of some of it being split into `tox.ini`.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -53,8 +53,6 @@ pass_env =
     TEST_BIGDATA
     CODECOV_*
     DD_*
-set_env =
-    devdeps: PIP_EXTRA_INDEX_URL = https://pypi.anaconda.org/scientific-python-nightly-wheels/simpl
 
 extras =
     test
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -68,7 +66,7 @@ commands_pre =
     oldestdeps: pip install -r requirements-min.txt
     devdeps: pip install -r requirements-dev.txt -U --upgrade-strategy eager
```
zacharyburnett, line 67:
```suggestion
    devdeps: pip install -r requirements-dev.txt
```

zacharyburnett (approved):

tox.ini
WilliamJamieson: This is what forces pip to install from the alternate indexes in order load the pre-built upstream wheels.

WilliamJamieson: If these requirements were installed before romancal then this would not be needed.

PR #844, opened 2023-08-21
Make it so we can manually trigger the CI

WilliamJamieson:
It is sometimes useful to be able to manually trigger the github CI without a schedule or PR.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


zacharyburnett (approved):
would it also be useful to include optional dependency overrides like in the jenkins job?

PR #857, opened 2023-09-06
RCAL-643 Update roman CI/CD to use CRDS local cache

### LINKED ISSUE #856: Update ci workflow to use crds cache
_Issue [RCAL-643](https://jira.stsci.edu/browse/RCAL-643) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

The ticket <https://jira.stsci.edu/projects/CCD/issues/CCD-1292> to create the roman crds cache area on central store is done. We need to update the CRDS variables in ci.yml to
CRDS_PATH=████████████████████
CRDS_SERVER_URL=https://serverless

ddavis-stsci:
Resolves [RCAL-643](https://jira.stsci.edu/browse/rcal-643)

Closes #856

This updates the roman_ci.yml and roman_ci_cron.yml to use the local roman crds cache with
      CRDS_SERVER_URL: https://serverless
      CRDS_PATH: /grp/crds/roman/test

**Checklist**
- [N/A] added entry in `CHANGES.rst` under the corresponding subsection
- [x] added relevant label(s)

The regression tests pass (https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/338/)

### DIFF
--- a/.github/workflows/roman_ci.yml
+++ b/.github/workflows/roman_ci.yml
@@ -25,8 +25,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       OBSERVATORY: roman
-      CRDS_SERVER_URL: https://roman-crds-test.stsci.edu
-      CRDS_PATH: /tmp/crds_cache
+      CRDS_SERVER_URL: https://serverless
+      CRDS_PATH: /grp/crds/roman/test
     steps:
       - id: context
         run: >
--- a/.github/workflows/roman_ci_cron.yaml
+++ b/.github/workflows/roman_ci_cron.yaml
@@ -22,8 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       OBSERVATORY: roman
-      CRDS_PATH: /tmp/crds_cache
-      CRDS_SERVER_URL: https://roman-crds-test.stsci.edu
+      CRDS_SERVER_URL: https://serverless
+      CRDS_PATH: /grp/crds/roman/test
     steps:
       - id: context
         run: >
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -53,8 +53,8 @@ def pip_install_args = "--index-url ${pip_index} --progress-bar=off"
 env_vars = [
     "TEST_BIGDATA=https://bytesalad.stsci.edu/artifactory",
     "CRDS_CONTEXT=roman_0051.pmap",
-    "CRDS_SERVER_URL=https://roman-crds.stsci.edu",
-    'CRDS_PATH=${WORKSPACE}/roman-crds-test-cache',
+    "CRDS_SERVER_URL=https://serverless",
+    'CRDS_PATH=/grp/crds/roman/test/',
     'DD_ENV=ci',
     'DD_SERVICE=romancal',
 ]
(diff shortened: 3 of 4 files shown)


ddavis-stsci:
> If the intention of this PR is to modify the regression test runs the edits need to be made to the `JenkinsfileRT` and `JenkinsfileRT_dev` files. The current edits to the `.github/worksflows` files will only effect github actions runs.

Thanks for catching this. I must have overwritten these when I went back for the github workflow changes.

ddavis-stsci:
> ~This also breaks the crds github action which fails to fetch the pmap but oddly does not mark the job as failed: https://github.com/spacetelescope/romancal/actions/runs/6099274369/job/16551025595?pr=857#step:2:19~
> 
> EDIT: it appears that the crds step is failing on main as well. The roman crds server does not return the required json in either case.

I think this is more tricky. The local cache has a list of the pmaps (/grp/crds/roman/test/mappings/roman/*.pmap) but does not seem to have a indication of which is the operational context. So I think we'd need to get that from the server (https://roman-crds.stsci.edu/) and then use the local cache to pick up the files. So we need to initially set the CRDS server to https://roman-crds.stsci.edu/ and then to https://serverless for the run. I'm not familiar with the roman_ci.yml code and I think @zacharyburnett wrote this section? Maybe he can comment on the best way to do this?

WilliamJamieson:
> I think this is more tricky. The local cache has a list of the pmaps (/grp/crds/roman/test/mappings/roman/*.pmap) but does not seem to have a indication of which is the operational context. So I think we'd need to get that from the server ([roman-crds.stsci.edu](https://roman-crds.stsci.edu/)) and then use the local cache to pick up the files. So we need to initially set the CRDS server to [roman-crds.stsci.edu](https://roman-crds.stsci.edu/) and then to [serverless](https://serverless) for the run. I'm not familiar with the roman_ci.yml code and I think @zacharyburnett wrote this section? Maybe he can comment on the best way to do this?

Merging ddavis-stsci/romancal#1 into this branch should be the solution you described.

schlafly (approved):
Looks good to me, thanks!

zacharyburnett:
I apologize for the delay in response; I'm not sure this actually works. I don't think GitHub Actions has access to `/grp/`, and the test run shows that the cache was not found: https://github.com/spacetelescope/romancal/actions/runs/6113945200/job/16594486020#step:3:9
```
Cache not found for input keys: crds-
```

.github/workflows/roman_ci.yml
```diff
@@ -39,7 +40,7 @@ jobs:
       - id: path
         run: echo "path=${{ env.CRDS_PATH }}" >> $GITHUB_OUTPUT
       - id: server
-        run: echo "url=${{ env.CRDS_SERVER_URL }}" >> $GITHUB_OUTPUT
+        run: echo "url=${{ env.CRDS_LOCAL_URL }}" >> $GITHUB_OUTPUT
```
zacharyburnett, line 43: I may be incorrect here (I don't fully understand CRDS), but GitHub Actions doesn't see the shared drive at `/grp/` and doesn't have permission to make a new directory at the root, so this may not accomplish the desired result of using the local cache, and I do not see how it will be able to retrieve CRDS files.

PR #858, opened 2023-09-07
set CRDS_LOCKING_MODE=filelock in jenkins runs

braingram:
Jenkins runs are using xdist which can cause seemingly random errors when two simultaneous tests fetch the same reference file. Setting `CRDS_LOCKING_MODE=filelock` should allow the jenkins runs to lock the crds cache across the processes used for the xdist run to avoid these errors. More information about the `filelock` setting can be found in the [JWST CRDS docs](https://github.com/spacetelescope/crds/blob/456d62bb53d923fbd7d5779b8a993278c40d3002/documentation/crds_users_guide/source/environment.rst#file-based-locking)

A minimal example of the issue is as follows:
```python
import os
from stpipe import crds_client


def fetch():
    params = {
        'roman.meta.instrument.optical_element': 'F158',
        'roman.meta.instrument.name': 'WFI',
        'roman.meta.model_type': 'WfiScienceRaw',
        'roman.meta.telescope': 'ROMAN',
        'roman.meta.exposure.start_time': '2021-01-01T00:00:00.000',
        'roman.meta.instrument.detector': 'WFI01'
    }
    reference_file_type = 'linearity'
    observatory = 'roman'
    reference_name = crds_client.get_reference_file(params, reference_file_type, observatory)
    return os.path.getsize(reference_name)


def test_fetch_1():
    assert fetch() == 536872886


def test_fetch_2():
    assert fetch() == 536872886
```

Running the above with pytest xdist can (and seemingly often) result in a failure if `CRDS_LOCKING_MODE=filelock` is not set. The failure occurs because one test will start downloading the reference file (creating the file in the crds cache), the second test sees the file and assumes it's good (crds does not check that the file is valid see https://github.com/spacetelescope/crds/issues/930).

Regression tests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/341/

Tests finished with no errors.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -55,6 +55,7 @@ env_vars = [
     "CRDS_CONTEXT=roman_0051.pmap",
     "CRDS_SERVER_URL=https://roman-crds.stsci.edu",
     'CRDS_PATH=${WORKSPACE}/roman-crds-test-cache',
+    'CRDS_LOCKING_MODE=filelock',
     'DD_ENV=ci',
     'DD_SERVICE=romancal',
 ]
--- a/JenkinsfileRT_dev
+++ b/JenkinsfileRT_dev
@@ -55,6 +55,7 @@ env_vars = [
     "CRDS_CONTEXT=roman_0051.pmap",
     "CRDS_SERVER_URL=https://roman-crds-test.stsci.edu",
     'CRDS_PATH=${WORKSPACE}/roman-crds-test-cache',
+    'CRDS_LOCKING_MODE=filelock',
 ]
 if (env.ENV_VARS) {
     env_vars.addAll(MultiLineToArray(env.ENV_VARS))


braingram:
@ddavis-stsci this should hopefully fix the random regtest failures

braingram:
devdeps failure is unrelated and caused by numpy 2.0.

ddavis-stsci (commented):
At the moment I don't think that this will do anything since the use of the local CRDS cache has been enabled for the Jenkins runs. Once this is done we may not need this since JWST is not seeing the issues we have and they do not set CRDS_LOCKING_MODE. If these random errors continue then this would be a good next step.

braingram:
> At the moment I don't think that this will do anything since the use of the local CRDS cache has been enabled for the Jenkins runs. Once this is done we may not need this since JWST is not seeing the issues we have and they do not set CRDS_LOCKING_MODE. If these random errors continue then this would be a good next step.

Thanks for taking a look. Where is the local cache enabled for the jenkins run? My understanding of the `JenkinsfileRT` setting here:
https://github.com/spacetelescope/romancal/blob/191a25d650117ad35a3a353b3cf8005d2a96f571/JenkinsfileRT#L57
is that the crds files will be fetched to the workspace directory. The workspace directory is wiped for each run (at least that's how I'm understanding the `Wiping out workspace first.` message in the [log of a recent run](https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/341/consoleText)). This should mean that ever run re-fetches crds files. @zacharyburnett is this correct? This all differs significantly from JWST which could explain why they don't have similar random errors.

ddavis-stsci:
It is not yet enabled. You can see the PR at 
https://github.com/spacetelescope/romancal/pull/857
Jonathan is hopeful that this will fix the problem.

JWST runs many more regression tests but in general their input files 
are smaller, ~25-50MB as
opposed to our input files of almost 200MB and this is also true for the 
reference files. Our dark
reference files range from 832MB up to 2.6GB.

In general I'd like to make one change at a time so we can track what 
updates fix the issue.

ddavis-stsci:
Another note is that if we are using the local cache on /grp/crds/roman/test then file locking will have no effect since, 
```
The read-only nature of:
/grp/crds/cache*
prevents the use of locking for typical onsite users. None should be required.
```

braingram:
I did not realize [#857](https://github.com/spacetelescope/romancal/pull/857) was attempting to modify the jenkins runs. I'll close this PR and leave a comment there.

PR #871, opened 2023-09-18
fix caching and outputs

zacharyburnett:
This PR addresses some issues with caching and using the output variables in the GitHub Actions workflow:
1. the `hashFiles` method apparently only works in the workspace, so attempting to hash `/tmp/data/` silently failed
2. the variable call used the old `crds` job name instead of `data`, so also silently failed

I didn't catch this issue when reviewing #794, but it should be fixed here

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


ddavis-stsci (approved):
It looks like the devdeps failures are unrelated so LGTM.

PR #876, opened 2023-09-20
fix optional workflow triggers

zacharyburnett:
This PR applies the same fix as in `stdatamodels`, to resolve issues with triggering optional workflows.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/.github/workflows/roman_ci_cron.yaml
+++ b/.github/workflows/roman_ci_cron.yaml
@@ -8,8 +8,11 @@ on:
     # We also want this workflow triggered if the `Weekly CI` label is
     # added or present when PR is updated
     types:
-      - synchronize
+      - opened
+      - reopened
       - labeled
+      - unlabeled
+      - synchronize
   push:
     tags:
       - "*"
--- a/.github/workflows/tests_devdeps.yml
+++ b/.github/workflows/tests_devdeps.yml
@@ -8,8 +8,12 @@ on:
     tags:
       - '*'
   pull_request:
-    branches:
-      - master
+    types:
+      - opened
+      - reopened
+      - labeled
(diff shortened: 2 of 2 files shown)


ddavis-stsci (approved):
LGTM but I'm not the expert.
Question: for types labeled and unlabeled, Does that trigger the tests every time a label is changed?

zacharyburnett:
> Question: for types labeled and unlabeled, Does that trigger the tests every time a label is changed?

Yes, unfortunately every time a label is changed, the test will attempt to check if its label is present and then try to run again if so. This is to my knowledge a limitation of the trigger keywords (the workflow can't be configured to trigger on a specific label, only on the `labeled` / `unlabeled` events)

ddavis-stsci:
> > Question: for types labeled and unlabeled, Does that trigger the tests every time a label is changed?
> 
> Yes, unfortunately every time a label is changed, the test will attempt to check if its label is present and then try to run again if so. This is to my knowledge a limitation of the trigger keywords (the workflow can't be configured to trigger on a specific label, only on the `labeled` / `unlabeled` events)

I'm not sure that is a good idea. Folks forget labels often (well pretty much all the time) and I go and add them. I'm not sure that warrants rerunning the tests. I'm not sure what this tells us about the code.

zacharyburnett:
The tests that would be rerunning are the scheduled workflows, that have extra platforms (MacOS on more versions of Python) or the optional `devdeps` tests. The label has to be applied to the PR, and also the user would have to relabel the PR, for these tests to rerun (something that would happen on the next commit or rebase anyway). Since they are not needed in the majority of cases, I think it's a good trade off for the downside that they only rerun when both conditions are met.

PR #877, opened 2023-09-20
New ramp fitting default

schlafly:
This PR makes the new uneven ramp fitting algorithm the default.

**Checklist**
- [ ] updated relevant tests
- [X] added relevant label(s)


ddavis-stsci (approved):
LGTM

schlafly:
Thanks.  This is extra failing regression testing at the moment, because of a mismatch between read_pattern and the data arrays.  read_pattern implies that there are 8 resultants, each with 6 reads, while the data implies 6 resultants, with 8 reads each.  Pinging @tddesjardins to see if there's confusion there.  I'll hold off on merging this one at the moment.

schlafly:
Tyler placed an updated uncal file in /grp/roman/TEST_DATA/24Q1_B12/DMS/regression/r0000101001001001001_01101_0001_WFI01_uncal.asdf .  I can now successfully run the new ramp fitting on that file.  For fun, I took a look at the memory usage.
image
The sharp drop at ~9:33:50 is the end of the jump detection step.  When William merges the new jump detection code we should stay closer to the ramp fitting ~3 GB level at peak.

Since this now runs I am going to go ahead and merge it.

PR #911, opened 2023-09-28
rcal-681 Update reg tests for new algorithms

### LINKED ISSUE #909: romancal regression test updates
_Issue [RCAL-681](https://jira.stsci.edu/browse/RCAL-681) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Some of the recent changes to the software require corresponding updates to the regression tests.

1) the rampfit algorithm ols_as22 does not have the option to output the fit parameters so the fla --save_opt does not work (fails silently). This should be removed from the test_ramp_fitting module.

2) The input to test_wfi_flat_field is the _rampfit.asdf file and currently this should be the _assignwcs.asdf file, update the test input.

ddavis-stsci:
Resolves [RCAL-681](https://jira.stsci.edu/browse/rcal-681)

Closes #909 

This PR updates reg tests for new algorithms and removal of the optional rampfit output of fitopt since the new ols_cas22 method does not save these. 

**Checklist**
- [X] updated relevant tests
- [x] added relevant label(s)
All the regression tests now pass with the exception of the tweakreg tests, which need a new custom input file from INS.
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/390

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 general
 -------
 
+- Update regression tests with new data and update ramp fitting tests to use ols_cas22 [#911]
+
 - Fix bug with ``ModelContainer.get_crds_parameters`` being a property not a method [#846]
 
 - Fix random seed bug in PSF fitting methods [#862]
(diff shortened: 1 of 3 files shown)


schlafly (approved):
Looks good to me, thanks!

PR #919, opened 2023-10-02
TweakReg regression test fix.

mairanteodoro:
This PR implements the following items to fix TweakReg's (TR) regression test:

- fix TR's catalog metadata: a bug was preventing the catalog metadata `name` from being properly set;
- updated the files used by TweakReg's regression test:
   * input: `roman-pipeline/dev/WFI/image/r0000501001001001001_01101_0001_WFI02_cal_tweakreg.asdf`;
   * truth: `roman-pipeline/dev/truth/WFI/image/r0000501001001001001_01101_0001_WFI02_tweakreg.asdf`.

The truth file in Artifactory was created by processing the cal file through the `TweakRegStep`:
```python
from romancal.stpipe import RomanStep
from romancal.tweakreg.tweakreg_step import TweakRegStep

step = TweakRegStep()

args = [
    "romancal.step.TweakRegStep",
    "/full/path/to/association/file/asn_filename.json",
    "--output_file='tweakreg_truth_filename.asdf'",
    "--suffix='output'",
]
RomanStep.from_cmdline(args)
```
The association file (named `asn_filename.json` above) used in the creation of the truth file has the following content:
```json
{
       "asn_type": "None",
       "asn_rule": "DMS_ELPP_Base",
       "version_id": null,
       "code_version": "0.9.1.dev28+ge987cc9.d20230106",
       "degraded_status": "No known degraded exposures in association.",
       "program": "noprogram",
       "constraints": "No constraints",
       "asn_id": "a3001",
       "target": "none",
       "asn_pool": "test_pool_name",
       "products": [
           {
               "name": "files.asdf",
               "members": [
                   {
                       "expname": "r0000501001001001001_01101_0001_WFI02_cal_tweakreg.asdf",
                       "exptype": "science"
                   }
               ]
            }
       ]
}
```

The file used as input for the regression test, `r0000501001001001001_01101_0001_WFI02_cal.asdf`, was provided by @tddesjardins and has been processed through `assign_wcs`, `photom`, and `source_detection`. We added the `tweakreg` suffix before uploading it to Artifactory.

**Astrometric Utils unit tests**
This PR also adds parallax correction to the unit tests for the `astrometric_utils` package.

Since the coordinates in the catalog obtained from the VO API are now being updated for proper motion and parallax, we must consider that in the `test_astrometric_utils` suite.

**Regression test**
~All regression tests are passing: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/401/~
${\text{\color{red}The regression tests are failing to start and Zach and William are working on fixing it.}}$

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/tweakreg/tests/test_astrometric_utils.py
+++ b/romancal/tweakreg/tests/test_astrometric_utils.py
@@ -24,10 +24,158 @@ from romancal.tweakreg.astrometric_utils import (
     get_catalog,
 )
 
+ARAD = np.pi / 180.0
+
 
 class MockConnectionError:
     def __init__(self, *args, **kwargs):
         raise requests.exceptions.ConnectionError
+
+
+def get_earth_sun_orbit_elem(epoch):
+    amjd = (epoch - 2012.0) * 365.25 + 55927.0
+    imjd = int(amjd)
+    mjd = float(imjd)
+
+    time_argument = mjd - 51544.5
+    fmean_longitude = 280.460 + 0.9856474 * time_argument
+    fmean_anomaly = 357.528 + 0.9856003 * time_argument
+    iremainder = int(fmean_longitude / 360.0)
+    fmean_longitude = fmean_longitude - (iremainder * 360.0)
+    if fmean_longitude < 0:
+        fmean_longitude = fmean_longitude + 360.0
+    iremainder = int(fmean_anomaly / 360.0)
+    fmean_anomaly = fmean_anomaly - (iremainder * 360.0)
+    if fmean_anomaly < 0:
+        fmean_anomaly = fmean_anomaly + 360.0
+    ecliptic_longitude = (
+        fmean_longitude
+        + 1.915 * np.sin(fmean_anomaly * ARAD)
+        + 0.02 * np.sin(2.0 * fmean_anomaly * ARAD)
+    )
+    ecliptic_obliquity = 23.439 - 4.0 * 0.0000001 * time_argument
+
+    # radius vector (in AU)
+    # (approximation using an algorithm from USNO)
+    earth_sun_distance = (
+        1.00014
+        - 0.01671 * np.cos(fmean_anomaly * ARAD)
+        - 0.00014 * np.cos(2.0 * fmean_anomaly * ARAD)
+    )
+
+    earth_sun_distance = u.Quantity(earth_sun_distance, unit="AU")
+
+    return {
+        "earth_sun_distance": earth_sun_distance,
+        "ecliptic_longitude": ecliptic_longitude,
+        "ecliptic_obliquity": ecliptic_obliquity,
+    }
+
+
+def get_parallax_correction_mast(epoch, gaia_ref_epoch_coords):
+    orbit_elem = get_earth_sun_orbit_elem(epoch)
+    earth_sun_distance = orbit_elem["earth_sun_distance"]
+    ecliptic_longitude = orbit_elem["ecliptic_longitude"]
+    ecliptic_obliquity = orbit_elem["ecliptic_obliquity"]
+
+    # cartesian coordinates of the sun
+    xsun = earth_sun_distance * np.cos(ecliptic_longitude * ARAD)
+    ysun = (
+        earth_sun_distance
+        * np.cos(ecliptic_obliquity * ARAD)
+        * np.sin(ecliptic_longitude * ARAD)
+    )
+    zsun = (
+        earth_sun_distance
+        * np.sin(ecliptic_obliquity * ARAD)
+        * np.sin(ecliptic_longitude * ARAD)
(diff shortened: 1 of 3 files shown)


ddavis-stsci:
Is there some reason you are switching from WFI01 to WFI02? 
Does this mean that we need to retire the files for the older tests once this is in? 
The WFI02 file is also a custom file from INS (Tyler)?

mairanteodoro:
> Is there some reason you are switching from WFI01 to WFI02?

Yes; the reason is that seven files are throwing an error (`ValueError: Out of domain for acos`) when calculating the area of the spherical triangles during the `TweakRegStep` processing. The files are:

- `r0000501001001001001_01101_0001_WFI07_cal.asdf`;
- `r0000501001001001001_01101_0001_WFI16_cal.asdf`;
- `r0000501001001001001_01101_0001_WFI17_cal.asdf`;
- `r0000501001001001001_01101_0002_WFI01_cal.asdf`;
- `r0000501001001001001_01101_0002_WFI03_cal.asdf`;
- `r0000501001001001001_01101_0002_WFI09_cal.asdf`;
- `r0000501001001001001_01101_0002_WFI10_cal.asdf`.

If we can't process the files through `TweakRegStep`, we can't pass it to `ResampleStep`. Both @tddesjardins and @nden are aware of this issue.

For the resample regression test specifically, we are using SCA 02 because that allows us to test the alignment of the same SCA but from different exposures (the latest simulations have a small offset between exposures). SCA 01 can't be used for that since it can't be processed through `TweakRegStep`.


> Does this mean that we need to retire the files for the older tests once this is in?

I don't think so; this issue affects only TweakReg.

> The WFI02 file is also a custom file from INS (Tyler)?

Yes.

ddavis-stsci:
Do we have a timescale of when the "error (ValueError: Out of domain for acos)"  might be fixed? This clearly cannot go into the elpp until this is working. 
At that point we should revert the file to WFI01, or convert all the tests to use WFI02.

mairanteodoro:
> Do we have a timescale of when the "error (ValueError: Out of domain for acos)" might be fixed? This clearly cannot go into the elpp until this is working. At that point we should revert the file to WFI01, or convert all the tests to use WFI02.

The ticket for that work is https://jira.stsci.edu/browse/RCAL-688.

romancal/tweakreg/tests/test_astrometric_utils.py
```diff
@@ -481,39 +598,51 @@ def test_get_catalog_valid_parameters_but_no_sources_returned():
...
     )
 
     assert len(result) > 0
-    assert np.isclose(returned_ra, expected_new_ra, atol=1e-10, rtol=1e-9).all()
-    assert np.isclose(returned_dec, expected_new_dec, atol=1e-10, rtol=1e-9).all()
+    # choosing atol=0 corresponds to abs(a - b) / abs(b) <= rtol,
+    # where a is the returned value from the VO API and b is the expected
+    # value from applying the calculated PM and parallax corrections.
+    assert np.isclose(returned_ra, expected_ra, atol=0, rtol=1e-5).all()
+    assert np.isclose(returned_dec, expected_dec, atol=0, rtol=1e-5).all()
```
schlafly, line 645: An rtol of 1e-5 worries me?  For an object with a coordinate of something like 100, this is 1e-3 deg = 3.6" = 36 pixels?  Even for something with a coordinate of ~1, this is 0.36 pixels while we want to be doing more like 0.01 pixel.  Maybe just an atol of 0.001 and no rtol?

mairanteodoro: Changed `atol` to `1e-5` and `rtol=0`.

schlafly:
What is the actual level of agreement you're getting?

I said atol = 0.001 but I'm not sure what I was thinking; that's the 3.6" agreement we were talking about before.  You did atol=1e-5, which is 0.036" = 0.36 pixel, which is not great; we really want ~0.01 - 0.001 pix or something.

mairanteodoro:
> What is the actual level of agreement you're getting?
> 
> I said atol = 0.001 but I'm not sure what I was thinking; that's the 3.6" agreement we were talking about before. You did atol=1e-5, which is 0.036" = 0.36 pixel, which is not great; we really want ~0.01 - 0.001 pix or something.

We are using `atol=1e-8` (absolute coordinates difference, in deg) in the unit tests, which corresponds to an absolute difference in coordinates of $<$ 40 &mu;as.

mairanteodoro:
The tolerance for coordinates correction has been set to $2.8\times10\^{-9}$ deg, which corresponds to 10 $\mu$ as

schlafly (approved):
Thanks Mairan.  That looks good to me.

PR #920, opened 2023-10-02
update url for webbpsf data in github actions CI

braingram:
This PR updates the url used for fetching webbpsf data to the most recent one in the webbpsf docs.
https://webbpsf.readthedocs.io/en/latest/installation.html#installing-the-required-data-files

Based off the comments and changes in https://github.com/spacetelescope/webbpsf/pull/739
the shared url for the webbpsf data used in this PR should (hopefully) stay up-to-date for future webbpsf data releases.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/roman_ci.yml
+++ b/.github/workflows/roman_ci.yml
@@ -55,7 +55,7 @@ jobs:
       - id: data
         run: |
           echo "path=/tmp/data" >> $GITHUB_OUTPUT
-          echo "webbpsf_url=https://stsci.box.com/shared/static/n1fealx9q0m6sdnass6wnyfikvxtc0zz.gz" >> $GITHUB_OUTPUT
+          echo "webbpsf_url=https://stsci.box.com/shared/static/qxpiaxsjwo15ml6m4pkhtk36c9jgj70k.gz" >> $GITHUB_OUTPUT
       - run: |
           mkdir -p tmp/data/
           mkdir -p ${{ steps.data.outputs.path }}


ddavis-stsci:
Is there some reason that this file is not in artifactory? It is only 69MB in size. Also if box is down (as it was last night) the tests will just fail? 
Do we have instructions on how to update this file?

braingram:
> Is there some reason that this file is not in artifactory? It is only 69MB in size. Also if box is down (as it was last night) the tests will just fail? Do we have instructions on how to update this file?

That's a good question. @zacharyburnett do the github CI runs have access to artifactory?

At the moment yes if box is down the tests will fail as the `data` job (which fails if the file on box fails to download) is a prerequisite of many of the other CI jobs (including all that run the unit tests):
https://github.com/spacetelescope/romancal/blob/92b07ff213c52fcd29a35618db62e4479637c811/.github/workflows/roman_ci.yml#L81

I have been unable to find any documentation from the webbpsf project on when this file is updated. I believe the shared link for the data files included in this PR is intended to stay up-to-date based off the [release docs for webbpsf](https://github.com/BradleySappington/webbpsf/blob/develop/dev_utils/master_data_release.sh) which state:
```
Verify the shared link of webbpsf-data-latest.tar.gz is the same that exists in docs/installation.rst (“copy shared link” then “link settings”)
```

zacharyburnett:
> @zacharyburnett do the github CI runs have access to artifactory?

Not currently, but when we transition to on-prem runners they will.

zacharyburnett (approved):

bmorris3 (approved):
This URL will work going forward, and the data located at this URL will be in sync with the latest release of WebbPSF, so I'm approving this PR.

There's a further improvement that we should make to this action thanks to recent updates to the WebbPSF required data infrastructure from @bsappington. We can now also access that file from central store _without a remote download_, at `/grp/jwst/ote/webbpsf-data`, which may be preferable to avoid timeouts/relying on Box. I won't be back in-office until Oct 16 to make a PR to use central store rather than Box. If someone else wants to do it, you're welcome to 😄

braingram:
CI failures are unrelated to these changes and are also occurring on main:
https://github.com/spacetelescope/romancal/actions/runs/6394475030/job/17355965466#step:10:201

Issue #922, opened 2023-10-03
`test_astrometric_utils` failures on main branch


braingram:
11 tests suddenly started failing in `test_astrometric_utils.py`.
https://github.com/spacetelescope/romancal/actions/runs/6394475030/job/17355965466#step:10:201

@mairanteodoro any initial thoughts on why these might be failing?

```
FAILED romancal/tweakreg/tests/test_astrometric_utils.py::test_get_catalog_using_epoch[10-10-2000] - assert False
 +  where False = <built-in method all of numpy.ndarray object at 0x7fa17feabe70>()
 +    where <built-in method all of numpy.ndarray object at 0x7fa17feabe70> = array([False, False, False, False, False, False, False, False,  True,\n       False,  True, False, False, False, False,...False, False, False,\n       False, False, False, False, False, False, False, False,  True,\n       False, False, False]).all
 +      where array([False, False, False, False, False, False, False, False,  True,\n       False,  True, False, False, False, False,...False, False, False,\n       False, False, False, False, False, False, False, False,  True,\n       False, False, False]) = <function isclose at 0x7fa1ac1bbdb0>(array([ 9.99259107,  9.98845403,  9.98513531,  9.99349667, 10.00546396,\n        9.98691049, 10.01945872,  9.96919236, ...\n       10.08310588, 10.06293681, 10.01253871, 10.05945472,  9.90433175,\n       10.09693267, 10.00078586, 10.00124519]), array([ 9.99259109,  9.98845432,  9.98513556,  9.9934973 , 10.00546421,\n        9.98691121, 10.01945864,  9.96919267, ...\n       10.08310596, 10.06293726, 10.01253885, 10.0594541 ,  9.90433175,\n       10.09693307, 10.00078581, 10.00124548]), atol=1e-10, rtol=1e-09)
 +        where <function isclose at 0x7fa1ac1bbdb0> = np.isclose
FAILED romancal/tweakreg/tests/test_astrometric_utils.py::test_get_catalog_using_epoch[10-10-2010.3] - assert False
 +  where False = <built-in method all of numpy.ndarray object at 0x7fa17fc8fbd0>()
 +    where <built-in method all of numpy.ndarray object at 0x7fa17fc8fbd0> = array([ True, False, False, False, False, False, False, False,  True,\n       False,  True, False, False, False, False,...False, False, False,\n       False, False, False, False, False, False, False, False,  True,\n       False, False, False]).all
[...]

PR #910, opened 2023-09-27
force uninstall `deepdiff` to install the dev version (not dynamically versioned)

zacharyburnett:
This PR addresses the issue where, since `deepdiff` is not dynamically versioned (and thus does not have dev versions e.g. `6.5.0.dev12`), attempting to install the latest commit of `deepdiff` will fail if a wheel of `deepdiff` is already installed. As the Jenkins build workflow does not include support for the `--ignore-installed` option to `pip install`, this uninstalls the existing `deepdiff` installation to overwrite with the development version.

**Checklist**
- [x] ~added entry in `CHANGES.rst` under the corresponding subsection~
- [ ] updated relevant tests
- [x] ~updated relevant documentation~
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)
- [ ] ran regression tests, post a link to the Jenkins job below. [How to run regression tests on a PR](https://github.com/spacetelescope/romancal/wiki/Running-Regression-Tests-Against-PR-Branches)

PR #923, opened 2023-10-03
Change webbpsf data cache in github CI

braingram:
Based off changes in https://github.com/spacetelescope/romancal/pull/920 (which is approved by waiting to merge until https://github.com/spacetelescope/romancal/issues/922 can be fixed to allow the CI to run and the PR to pass branch protections).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
There are many failed jobs because of https://github.com/spacetelescope/romancal/issues/922 and the devdeps are failing due to numpy 2.0 deepdiff incompatibility (the devdeps job does not appear to be installing the dev version and appears related to https://github.com/spacetelescope/romancal/pull/910).

@zacharyburnett I requested a review on this PR even though it will stay draft until #922 can be fixed and #920 merged. Thanks for giving it a look!

.github/workflows/roman_ci.yml
```diff
@@ -91,8 +43,9 @@ jobs:
         DD_GIT_REPOSITORY_URL: ${{ github.repositoryUrl }}
         DD_GIT_COMMIT_SHA: ${{ github.sha }}
         DD_GIT_BRANCH: ${{ github.ref_name }}
-      cache-path: ${{ needs.data.outputs.data_path }}
-      cache-key: data-${{ needs.data.outputs.data_hash }}-${{ needs.data.outputs.crds_context }}
+      cache-path: ${{ needs.data.outputs.crds_path }}
```
zacharyburnett, line 46: shouldn't this be the generic data path?
```suggestion
      cache-path: ${{ needs.data.outputs.data_path }}
```

zacharyburnett, line 46: shouldn't this be the generic data path?
```suggestion
      cache-path: ${{ needs.data.outputs.data_path }}
```

.github/workflows/roman_ci.yml
```diff
@@ -21,55 +21,7 @@ concurrency:
...
-      - id: data_hash
-        run: echo "hash=${{ hashFiles( 'tmp/data' ) }}" >> $GITHUB_OUTPUT
-      - run: mv tmp/data/* ${{ steps.data.outputs.path }}
-      - uses: actions/cache@v3
-        with:
-          path: ${{ steps.data.outputs.path }}
-          key: data-${{ steps.data_hash.outputs.hash }}-${{ steps.crds_context.outputs.pmap }}
-      - id: webbpsf_path
-        run: echo "path=${{ steps.data.outputs.path }}/webbpsf-data" >> $GITHUB_OUTPUT
+    uses: ./.github/workflows/data.yml
```
zacharyburnett, line 24: I think this will attempt to run the `webbpsf-data` job every time, right? Then, if it is skipped, the subsequent jobs that need it will also be skipped

zacharyburnett: oh wait nevermind, I see now the `if: always() ...` line

.github/workflows/roman_ci.yml
braingram: I ended up removing `data_path` as it was the same as `crds_path` (both were `/tmp/data`) and the jobs were inconsistent with usage (`roman_ci` used `data_path`, `roman_ci_cron` used `crds_path` for the cache path).

What do you think about re-introducing `data_path` as something like `/tmp/data` and have `crds_path` be `/tmp/data/crds` (now in a `crds` subdirectory) and `webbpsf_path` be `/tmp/data/webbpsf` (same as before)? That way if more data is needed it can go into a separate subdirectory.

.github/workflows/roman_ci.yml
braingram: I should add a comment describing that as I'm not sure if that's a common way to do things.

.github/workflows/roman_ci.yml
zacharyburnett: yes that sounds good to me; also, we probably shouldn't be using the CRDS context in the cache key anyway, as CRDS files are only downloaded during tests not before; any downloaded CRDS files would not be saved

zacharyburnett:
made a PR at https://github.com/braingram/romancal/pull/1

.github/workflows/roman_ci.yml
braingram: Added a comment in 9cdab6831a7e3acc49222a36d3e49d0a74c6bed2

zacharyburnett:
It looks like the data cache is not being hit: https://github.com/spacetelescope/romancal/actions/runs/6408370917/job/17397252769#step:3:16
which doesn't entirely make sense, according to the logic in the docs: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

.github/workflows/roman_ci.yml
```diff
@@ -91,8 +41,11 @@ jobs:
         DD_GIT_REPOSITORY_URL: ${{ github.repositoryUrl }}
         DD_GIT_COMMIT_SHA: ${{ github.sha }}
         DD_GIT_BRANCH: ${{ github.ref_name }}
-      cache-path: ${{ needs.data.outputs.data_path }}
-      cache-key: data-${{ needs.data.outputs.data_hash }}-${{ needs.data.outputs.crds_context }}
+      cache-path: ${{ needs.data.outputs.path }}
+      cache-key: data-${{ needs.data.outputs.webbpsf_hash }}-${{ needs.data.outputs.crds_context }}
+      cache-restore-keys: |
+        data-${{ needs.data.outputs.webbpsf_hash }}-
+        data-
```
braingram, line 48: This will fall back to any cache that starts with `data-`. I don't think this is necessary (since the `webbpsf_hash` is read from the available hashes) and would open up the possibility that an incorrectly computed webbpsf hash would use other webbpsf data. For example if a hash `abc` was computed but no `data-abc` cache exists and a `data-123` cache exists having `data-` as a restore key would create the new cache with `data-abc-{pmap}` that would contain the `data-123` data.

zacharyburnett: alright, I'll remove that then
```suggestion
```

.github/workflows/data.yml
```diff
@@ -0,0 +1,104 @@
...
+          wget ${{ env.WEBBPSF_DATA_URL }} -O tmp/webbpsf-data.tar.gz
+          echo "hash=$( shasum tmp/webbpsf-data.tar.gz | cut -d ' ' -f 1 )" >> $GITHUB_OUTPUT
+      - id: cache_path
+        run: echo "path=${{ needs.path.outputs.path }}/webbpsf-data" >> $GITHUB_OUTPUT
+      - id: cache_download
+        name: check downloaded data against the existing cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.cache_path.outputs.path }}
+          key: data-${{ steps.download.outputs.hash }}-
```
braingram, line 75:
```suggestion
          key: data-${{ steps.download.outputs.hash }}
```
I wonder if the trailing hyphen in the generated cache is confusing the search.

zacharyburnett:
https://github.com/spacetelescope/romancal/actions/runs/6408979402/job/17399216721#step:3:10
```
Cache not found for input keys: data-639888f0376b13e17a71a27e1eff514f8c64b18e-roman_0051.pmap, webbpsf-639888f0376b13e17a71a27e1eff514f8c64b18e
```

.github/workflows/data.yml
```diff
@@ -0,0 +1,104 @@
...
+          mkdir -p tmp
+          wget ${{ env.WEBBPSF_DATA_URL }} -O tmp/webbpsf-data.tar.gz
+          echo "hash=$( shasum tmp/webbpsf-data.tar.gz | cut -d ' ' -f 1 )" >> $GITHUB_OUTPUT
+      - id: cache_path
+        run: echo "path=${{ needs.path.outputs.path }}/webbpsf-data" >> $GITHUB_OUTPUT
+      - id: cache_download
+        name: check downloaded data against the existing cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.cache_path.outputs.path }}
```
braingram, line 74:
```suggestion
          path: ${{ needs.path.outputs.path }}
```

I'm not sure if it matters that the cache path differs here from what is used for the ci jobs. Perhaps that explains the miss?

zacharyburnett: oh wow, good catch! That does matter

zacharyburnett:
That seems to have worked; this PR looks like it's ready to review now

zacharyburnett:
> The webbpsf hash does not appear to be propagated to the test jobs: https://github.com/spacetelescope/romancal/actions/runs/6409720129/job/17401548600?pr=923#step:3:4 shows:
> 
> ```
>     key: data--roman_0051.pmap
>     restore-keys: data-639888f0376b13e17a71a27e1eff514f8c64b18e-
> ```

fixed with 2c6ef24e26f92a79af0b43985bde91349471a960
https://github.com/spacetelescope/romancal/actions/runs/6410194510/job/17403071740#step:3:4
```
    key: data-639888f0376b13e17a71a27e1eff514f8c64b18e-roman_0051.pmap
```

braingram (commented):
I like it!

Just 2 minor edits for the use of `crds_path` instead of `path`.

I'd also like to see some of the combinations "exercised" once changes are finished including:
- removing the "update webbpsf data" label from this PR and confirming that the data is not downloaded
- removing the "webbpsf-{hash}" cache and triggering the "data.yml" (with a workflow_dispatch seems good enough)

It would also be great to see the CI green so the final combined cache can be generated but that will have to wait for changes outside the scope of this PR. 
@mairanteodoro  Does it make sense to delay this PR until those are done?

zacharyburnett:
removing the `update webbpsf data` label retriggered the `devdeps` and `scheduled` jobs, and they did skip downloading the data and were able to grab the existing cache https://github.com/spacetelescope/romancal/actions/runs/6410785985/job/17404827427#step:3:16

zacharyburnett:
I removed the `webbpsf-{hash}` cache, but it looks like we aren't able to manually start a `workflow_dispatch` if it's not present on the `main` branch

braingram:
Thanks for running the tests! All looks good to me.

braingram:
@zacharyburnett I tried adding the `update webbpsf data` label here as there no longer appears to be any `webbpsf-` caches and that portion of the workflow does not appear to be running. If you have a chance would you take a look?

zacharyburnett:
> @zacharyburnett I tried adding the `update webbpsf data` label here as there no longer appears to be any `webbpsf-` caches and that portion of the workflow does not appear to be running. If you have a chance would you take a look?

I'm not sure what the issue was, but I reran the workflow and it looks like the caches are generated again and the tests are running.

braingram:
@zacharyburnett should this get another review since we've both contributed? If not, I'm happy merging it as the CI appears green (except for the devdeps jobs which are failing due to unrelated issues fixed in: https://github.com/spacetelescope/romancal/pull/940)

zacharyburnett:
> @zacharyburnett should this get another review since we've both contributed? If not, I'm happy merging it as the CI appears green (except for the devdeps jobs which are failing due to unrelated issues fixed in: #940)

I think it looks good to merge, at least to me

braingram:
Note that I manually triggered the workflow after merging to generate a `webbpsf-` cache on the main branch.

PR #925, opened 2023-10-05
Include parallax correction in unit tests.

mairanteodoro:
This PR adds parallax correction to the unit tests.

Since the coordinates in the catalog obtained from the VO API are now being updated for both proper motion and parallax, we also need to take that into account in the `test_astrometric_utils` suite.

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)


mairanteodoro:
This fix has been added to https://github.com/spacetelescope/romancal/pull/919.

PR #933, opened 2023-10-11
Update Cas22 ramp fitting to integrate with the jump detection changes in stcal

WilliamJamieson:
This PR is intended to only make the necessary changes in `romancal` to support the Cas22 algorithm's interface changes introduced by spacetelescope/stcal#215 for the purposes of integrating jump detection directly into the the ramp fit itself. This PR is not intended to add support for the new jump detection algorithm, instead it is intended to resolve the minor interface changes introduced which break `romancal`. A follow on PR is planned to enable and test the new jump detection algorithm in `romancal`.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,9 @@ ramp_fitting
 
 - Make uneven ramp fitting the default [#877]
 
+- Update Ramp fitting code to support the ``stcal`` changes to the ramp fitting
+  interface which were necessary to support jump detection on uneven ramps [#933]
+
 refpix
 ------
 
--- a/romancal/ramp_fitting/ramp_fit_step.py
+++ b/romancal/ramp_fitting/ramp_fit_step.py
@@ -32,6 +32,9 @@ class RampFitStep(RomanStep):
         opt_name = string(default='')
         maximum_cores = option('none','quarter','half','all',default='none') # max number of processes to create
         suffix = string(default='rampfit')  # Default suffix of results
+        use_jump_detection = boolean(default=False) # Use jump detection during ramp fitting
+        threshold_intercept = float(default=None) # Override the intercept parameter for the threshold function in the jump detection algorithm.
+        threshold_constant = float(default=None) # Override the constant parameter for the threshold function in the jump detection algorithm.
     """  # noqa: E501
 
     weighting = "optimal"  # Only weighting allowed for OLS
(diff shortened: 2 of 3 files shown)


schlafly:
This PR looks good to me, but I'm somewhat leery of having a dependency against WilliamJamieson/stcal.git---e.g., if Mairan needs to make more changes to stcal for resample, it feels like it gets really complicated to have that right.  When do you think we'll be able to merge your changes into stcal?

WilliamJamieson:
> This PR looks good to me, but I'm somewhat leery of having a dependency against WilliamJamieson/stcal.git---e.g., if Mairan needs to make more changes to stcal for resample, it feels like it gets really complicated to have that right. When do you think we'll be able to merge your changes into stcal?

If you look at the commit history those changes are all in a single commit with a message indicating they are temporary. The reason for them is that this PR and spacetelescope/stcal#215 are interlinked due to the interface change. Since `romancal` already directly depends on `stcal`'s development version rather than a published stable version, merging spacetelescope/stcal#215 will immediately break `romancal` without the changes in this PR. However, the changes for this PR require the changes in spacetelescope/stcal#215 in order to function. As a consequence both PRs need to be merged in short order i.e. merge spacetelescope/stcal#215 -> remove the temporary commit -> merge this PR so that there are minimal disruptions to the actual functionality of `romancal`.

All this is to say, both this PR and spacetelescope/stcal#215 need to be fully reviewed and "ready" before either can be merged. The temporary commit is just acting so that these changes can be fully tested and reviewed as it places both codebases in the "final" configuration needed.

schlafly:
Sounds good, except for the dependency I'm happy to sign off on these changes.

schlafly:
Tests are failing because this points to stcal/main, and https://github.com/spacetelescope/stcal/pull/215 hasn't been merged yet.  I saw that tests were successful when this was pointed to the correct PR.  I'm approving this, with the understanding that we won't actually merge it until 215 is in.

schlafly (approved):
Looks good; tests failing due to unmerged stcal PR 215 are understood.  Good to go in when that is merged.

WilliamJamieson:
Note that due to tiny numerical differences the `ramp_fitting` regression test will start failing. However, @schlafly and I have reviewed and discussed the changes offline and determined that these are minor (due to a single unrealistic pixel in the regression data), and thus can be ok-ified after the next regression test run.

PR #941, opened 2023-10-13
use ndarray comparison options in wcs comparison

braingram:
When comparing 2 WCS objects in the deepdiff computed between the regtest result and truth file, use the array comparison options that are used for other arrays in the diff.

These changes will mean that if both the result and truth WCS objects return `nan` when evaluated on the bounding box (this is what's used to compare the two objects) that the WCS objects can be considered equal if both WCS objects return the same `nan`s.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

Running regression tests at: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%252FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/426/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ general
 
 - Use tolerance for more comparisons in ``compare_asdf`` [#917]
 
+- Use array comparison options (including ``nan`` equality) when
+  comparing ``WCS`` objects during ``compare_asdf`` [#941]
+
 ramp_fitting
 ------------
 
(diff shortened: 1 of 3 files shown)


schlafly (approved):
Looks good to me, thanks!

braingram:
Regression tests passed but artifacts failed to upload with `java.lang.RuntimeException: Failed uploading artifacts by spec` likely due to some plwishmaster/artifactory issue.

PR #953, opened 2023-10-25
RCAL-629: static preview generation script

zacharyburnett:
Resolves [RCAL-629](https://jira.stsci.edu/browse/RCAL-629)

add script for enabling static preview image generation from the command line, following what was discussed in the meetings:
- preview file
  - [x] `1080p x 1080p` PNG
  - [x] filename `<original>.png`
  - [ ] margins showing image info
  - [x] overlay showing orientation (using code from https://github.com/astropy/astropy/pull/15434)
- thumbnail file
  - [x] `300p x 300p` PNG
  - [x] filename `<original>_thumb.png`
- [x] gathering mean values from binned pixels
- [x] 90th percentile linear histogram stretch
- [x] using `afmhot` colormap

### Installation
```
pip install ".[sdp]"
```
(adds `stpreview` and `typer` to the environment)

### Usage
```shell
roman_static_preview preview /grp/roman/TEST_DATA/23Q4_B11/aligntest/r0000501001001001001_01101_0001_WFI01_cal.asdf
```
r0000501001001001001_01101_0001_WFI01_cal
```shell
roman_static_preview thumbnail /grp/roman/TEST_DATA/23Q4_B11/aligntest/r0000501001001001001_01101_0001_WFI01_cal.asdf
```
thumbnail

### Checklist
- [x] added relevant label(s)


zacharyburnett:
There's probably a better name for this script than `static_preview`

zacharyburnett:
@nden when you get a chance, could you point me to JWST sample data as well? Thanks!

zacharyburnett:
```shell
❯ roman_static_preview preview --help  
Usage: roman_static_preview preview [OPTIONS] INPUT [OUTPUT] [SHAPE]...

  create a preview image with a north arrow overlay indicating orientation

Arguments:
  INPUT       path to ASDF file with 2D image data  [required]
  [OUTPUT]    path to output image file
  [SHAPE]...  desired pixel resolution of output image  [default: 1080, 1080]

Options:
  --compass / --no-compass  whether to draw a north arrow on the image
                            [default: compass]
  --help                    Show this message and exit.
```

```shell
❯ roman_static_preview thumbnail --help
Usage: roman_static_preview thumbnail [OPTIONS] INPUT [OUTPUT] [SHAPE]...

Arguments:
  INPUT       path to ASDF file with 2D image data  [required]
  [OUTPUT]    path to output image file
  [SHAPE]...  desired pixel resolution of output image  [default: 300, 300]

Options:
  --compass / --no-compass  whether to draw a north arrow on the image
                            [default: no-compass]
  --help                    Show this message and exit.
```

ddavis-stsci:
I assume that this is to be run on ImageModels only? Is there a check for that?

zacharyburnett:
> I assume that this is to be run on ImageModels only? Is there a check for that?

Good point, is there a flag in `roman.meta` that indicates whether it's an image or not? If not I can write a check

ddavis-stsci:
We usually use something like
    if isinstance(model, ImageModel):

CHANGES.rst
```diff
@@ -56,8 +56,14 @@ refpix
 
 resample
 --------
+
 - Implement resampling step. [#787]
 
+scripts
+-------
+
+- added ``static_preview`` script to generate static previews of ASDF images [#953]
```
zacharyburnett, line 65:
```suggestion
- added ``roman_static_preview`` script to generate static previews of ASDF images [#953]
```

romancal/scripts/static_preview.py
```diff
@@ -0,0 +1,102 @@
...
+    if output is None:
+        output = Path.cwd()
+    if output.is_dir():
+        output = output / f"{input.stem}.png"
+
+    with asdf.open(input) as file:
+        if "image" not in file["roman"]["meta"]["model_type"].lower():
+            raise NotImplementedError(
+                f"\"{file['roman']['meta']['model_type']}\" model not supported"
+            )
```
zacharyburnett, line 46: added a check for image model here

romancal/scripts/static_preview.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if output is None:
+        output = Path.cwd()
+    if output.is_dir():
+        output = output / f"{input.stem}.png"
+
+    with asdf.open(input) as file:
+        model = file["roman"]["meta"]["model_type"]
+        if "image" not in model.lower():
+            raise NotImplementedError(f'"{model}" model not supported')
```
zacharyburnett, line 45: added a check for image model here

schlafly (commented):
This looks good to me.  I worry a little about the simple mean used by default in the downsampling, but I guess since we're "only" downsampling 10x10 or so for the thumbnails isolated cosmics blowing up a pixel or something aren't that big of a problem (and jump detection should have caught them).

Do we have preview requirements on L1 images?

schlafly:
> Do we have preview requirements on L1 images?

Answering my own question here, no, only on L2.  I'm a little surprised and it's not crazy to want to make these previews for L1 images (e.g., by using the last frame of the L1 image, or the difference between the last and first frames).  But out of scope for this PR, which looks good to me.

schlafly (approved):

nden:
We need to generate also previews for L3 images but this is due in later builds.

nden:
@schlafly @stscieisenhamer Is there a case when ramp_fitting will set a pixel value in the image to NaN? e.g. if the slope cannot be computed or some other condition? If that's a possibility the `downsample` function needs to use `np.nanmean`

schlafly:
That's a good point.  Jonathan's code didn't ever return NaNs, but William's update does, so we should use nanmean.  The L3s also use NaNs.

nden (approved):
LGTM

PR #958, opened 2023-10-27
Skip webbpsf tests by default

WilliamJamieson:
Currently the `romancal` testing runs tests which require the `webbpsf`'s input data by default. This is undesirable because it requires users to have to set up this data in order for the test suite to complete successfully.

This PR adds the `--webbpsf` marker which enables these tests. This means that the `--webbpsf` flag must be passed to `pytest` in order to run the tests in question. This means that by default the `webbpsf` marked tests are not run. However,
this PR does add this marker to every CI job both on GitHub and Jenkins, so that the CI will always run the `webbpsf` tests. This means we will know prior to merging any changes if these tests are broken while also removing the burden of requiring outside users to not have to worry about `webbpsf`. 

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/roman_ci.yml
+++ b/.github/workflows/roman_ci.yml
@@ -47,16 +47,16 @@ jobs:
       cache-key: data-${{ needs.data.outputs.webbpsf_hash }}-${{ needs.data.outputs.crds_context }}
       cache-restore-keys: webbpsf-${{ needs.data.outputs.webbpsf_hash }}
       envs: |
-        - linux: py39-oldestdeps-cov
+        - linux: py39-oldestdeps-webbpsf-cov
           pytest-results-summary: true
-        - linux: py39
+        - linux: py39-webbpsf
           pytest-results-summary: true
-        - linux: py310
+        - linux: py310-webbpsf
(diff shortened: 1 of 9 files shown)


romancal/conftest.py
```diff
@@ -84,3 +76,32 @@ def pytest_runtest_logstart(self, nodeid, location):
...
+
+    # ADD default skip markers
+    config.addinivalue_line(
+        "markers", "webbpsf: mark test as requiring webbpsf data to run"
+    )
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--webbpsf", action="store_true", default=False, help="run webbpsf tests"
```
WilliamJamieson, line 96: Technically, we could set `default` to be true if users happen to have the `WEBBPSF_PATH` environment variable set.

zacharyburnett (approved):
Looks good! Should the readme instructions also be updated? https://github.com/spacetelescope/romancal/blob/13c316ee3a3296cce3ed66a4b292fc2159678be6/README.md?plain=1#L188

PR #959, opened 2023-10-27
Generate L3 exposure times from resampled exposure time images

schlafly:
Currently we generate L3 exposure times as sums of the times of the input exposures.  This is awkward since one can imagine having two different level 3 images:
1. 4 non-overlapping images that together cover the L3 image
2. 4 completely overlapping images that cover the L3 image

The current code gives an exposure time for both of these L3 images that is 4x the L2 image exposure time.  But in case (1) each individual pixel only sees 1x exposure time, so it's weird to say that its exposure time is 4x nominal.

This PR changes that computation so that case (1) gets 1x and case (2) gets 4x, reflecting the actual times the typical pixel sees.  It does this by generating a resampled exposure time image that reflects the amount of time each part of the output image sees from input images, and then computing some simple statistics on that exposure time image.

We could consider faster resampling modes to produce these images, but just reusing the existing resampling infrastructure was expedient.  This does add another resampling step, though, and so makes things ~25% slower.

Resolves [RCAL-696](https://jira.stsci.edu/browse/RCAL-696)

Regression tests are failing.  This is partially expected; the exposure time in the metadata changes intentionally because of this PR, and local runs of the regression test confirm that that change gets picked up and identified as a 'regression.'  I also get complaints that the data, context, err, var_rnoise, var_poisson are different.  That's also occurring on main, though, and isn't related to this PR.  It bears further investigation.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


romancal/resample/resample.py
```diff
@@ -401,20 +412,82 @@ def resample_variance_array(self, name, output_model):
...
+            exptime_tot += resampled_exptime.value
+
+        return exptime_tot
+
+    def update_exposure_times(self, output_model, exptime_tot):
         """Update exposure time metadata (in-place)."""
-        total_exposure_time = 0.0
+        m = exptime_tot > 0
+        if np.any(m):
+            total_exposure_time = np.median(exptime_tot[m])
```
ddavis-stsci, line 470: I'm not sure we should label the median exposure time as the total exposure time. 
Would is make sense to add an attribute median_exposure_time?

ddavis-stsci (commented):
Besides my small nit it looks good.

ddavis-stsci (approved):

romancal/resample/resample.py
schlafly: Yeah, changing this is part of the L3 metadata updates; I was just filling up columns.  OTOH, the exposure time being computed is the total exposure time for the median pixel with at least some exposure time, so it's not totally crazy---but I agree that we should update the metadata.  Do you think that should be part of this PR?

ddavis-stsci: I don't think this is urgent. 

I'm not to worried about that we can define it this way. I'm more concerned that some automated process will look for total_exposure_time and generate dubious results. 

We should probably have a larger discussion with others, INS, SE?, archive?, before making a decision and finalizing the attributes.

ddavis-stsci: I think this would need a larger discussion with the other components, INS, SE?, Archive? 
So I'd go ahead and merge this and write a follow up ticket.

PR #964, opened 2023-10-30
RCAL-670: additional check for data quality test.

mairanteodoro:
This PR adds a check for the existence of `NaN`s and zeros in the variance arrays, which is indicative of pixels that retained their initialization value (these pixels were not updated by the variance propagation process).

**Note**
We could not run the regression test due to an issue that's being investigated in [SPB-1688](https://jira.stsci.edu/browse/SPB-1688).

https://jira.stsci.edu/browse/RCAL-670

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


romancal/regtest/test_resample.py
```diff
@@ -106,9 +107,45 @@ def test_resample_single_file(rtdata, ignore_asdf_paths):
...
+                )
+            }"""  # noqa: E501
+    )
+    assert all(
+        np.sum(~np.isnan(getattr(resample_out, x)))
+        for x in ["var_poisson", "var_rnoise"]
+    )
+
+    step.log.info(
+        f"""DMS343 MSG: Are there NaNs or zeros in the data and variance arrays? :\
```
schlafly, line 127: It looks like this is only checking the variance arrays, which I think is appropriate.  Let's delete 'data and' and add some verbiage so that it more obviously connects to the requirement text.
```suggestion
        f"""DMS343 MSG: Are there NaNs or zeros in the variance arrays, indicating poor data quality? :\
```

schlafly (approved):
LGTM, thanks!

PR #977, opened 2023-11-06
Protect imports in script better

WilliamJamieson:
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/JP-nnnn)

Closes #976

This PR protects the external dependencies by making them local to a function. This way they are not imported immediately.

**Checklist**
- [ ] added entry in `CHANGES.rst` under the corresponding subsection
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)
- [ ] ran regression tests, post a link to the Jenkins job below. [How to run regression tests on a PR](https://github.com/spacetelescope/romancal/wiki/Running-Regression-Tests-Against-PR-Branches)

PR #982, opened 2023-11-08
fix import in script for use by typer

zacharyburnett:
This PR addresses an issue from https://github.com/spacetelescope/romancal/pull/977 where Typer doesn't have access to the imported `Annotated`

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/romancal/scripts/static_preview.py
+++ b/romancal/scripts/static_preview.py
@@ -1,14 +1,13 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Optional
+from typing import Optional
 
 import asdf
 import numpy
 
-if TYPE_CHECKING:
-    from typing_extensions import Annotated
-
 
 def command():
+    from typing_extensions import Annotated
+
     try:
         import typer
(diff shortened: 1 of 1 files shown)


romancal/scripts/static_preview.py
```diff
@@ -1,14 +1,13 @@
...
 
 import asdf
 import numpy
 
-if TYPE_CHECKING:
-    from typing_extensions import Annotated
-
 
 def command():
+    from typing_extensions import Annotated
```
WilliamJamieson, line 9: If this is supposed to be available at runtime then you must include this as one of the listed dependencies.

WilliamJamieson (changes requested):

romancal/scripts/static_preview.py
WilliamJamieson: Its probably fine, especially in a try block. I'm just thinking about what would happen if typer dropped the dependency, which it probably will do eventually.

WilliamJamieson (approved):

PR #983, opened 2023-11-08
replace `pyupgrade` `isort` and `black` with `ruff format`

zacharyburnett:
This PR replaces the `pre-commit` checks for `pyupgrade`, `isort`, and `black` with  `ruff format`

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -166,36 +166,23 @@ report = { exclude_lines = [
 [tool.bandit]
 skips = ["B101", "B307", "B404", "B603"]
 
-[tool.isort]
-profile = "black"
-filter_files = true
-line_length = 88
-
-[tool.black]
-line-length = 88
-force-exclude = '''
-^/(
-  (
-      \.eggs
-    | \.git
-    | \.pytest_cache
-    | \.tox
-  )/
-)
-'''
-
 [tool.ruff]
 line-length = 88
 exclude = [
-    'jdocs',
-    '.tox',
     '.eggs',
+    '.git',
+    '.pytest_cache',
+    '.tox',
+    'jdocs',
     'build',
 ]
 ignore = [
     'E741', # ambiguous variable name
 ]
 extend-select = ['NPY']
+extend-include = ["*.ipynb"]
+
+[tool.ruff.lint.isort]
 
 [tool.ruff.extend-per-file-ignores]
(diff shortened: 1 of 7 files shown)


WilliamJamieson:
In order to actually make this a replacement for all the claimed tools, the rule sets for `isort`, etc have to be enabled in the `pyproject.toml` file. Those are not part of the default rule set.

schlafly (approved):
This looks fine to me, and I'm probably in favor of fewer moving pieces, which seems to be the goal of this PR.

But I do think we want to ~settle on some choice of code formatting, etc., and not mess with it very often.  Does this represent that choice?

zacharyburnett:
> This looks fine to me, and I'm probably in favor of fewer moving pieces, which seems to be the goal of this PR.
> 
> But I do think we want to ~settle on some choice of code formatting, etc., and not mess with it very often. Does this represent that choice?

We can configure `ruff format` a bit more than we can Black, namely line length, quote character, indent style, etc.: https://docs.astral.sh/ruff/formatter/#configuration

pyproject.toml
```diff
@@ -166,36 +166,23 @@ report = { exclude_lines = [
...
+    'jdocs',
     'build',
 ]
 ignore = [
     'E741', # ambiguous variable name
 ]
 extend-select = ['NPY']
+extend-include = ["*.ipynb"]
+
+[tool.ruff.lint.isort]
```
WilliamJamieson, line 185: This still is not selecting the rules for ruff to replace the pre-commit tooling this PR claims to be replacing. I'll open a PR against your branch fixing these issues.

WilliamJamieson (changes requested):

schlafly:
I guess it looks like a reviewed this a long time ago and thought it was okay but that I wanted to believe that this would be a stable set of formatting thing for a long time.  Do we think that?  Are we using this in other projects or are we leading the way here?  Broadly, having a common formatting scheme is good but I don't think whatever we have been doing has been causing a lot of problems?

zacharyburnett:
I don't remember re-requesting a review for this like it says I did, nor for #1522; perhaps this was triggered by a change in automation on GitHub's side. I'll take this back into draft and work more on consolidating rules

zacharyburnett:
closing as this change has already been implemented

PR #984, opened 2023-11-08
Source detection+PSF docs updates

bmorris3:
This PR adds narrative docs for users and devs following the introduction of PSF fitting in Source Detection Step (#794, #841). This is a starting point, happy to iterate as reviewers see fit.

Resolves [RCAL-708](https://jira.stsci.edu/browse/RCAL-708)

### DIFF
--- a/docs/roman/source_detection/arguments.rst
+++ b/docs/roman/source_detection/arguments.rst
@@ -1,11 +1,13 @@
 Arguments
 =========
+
 The source detection fitting step has several arguments. These can be specified
 by the user by passing them to the step in a Python session, or setting them
 in a parameter file.
 
 * ``--kernel_fwhm``: A parameter for DAOStarFinder: size of Gaussian kernel in
-                     pixels. Default is 2.0.
+                     pixels. By default the FWHM is assumed to be the diffraction
+                     limited PSF, given the filter used for this observation.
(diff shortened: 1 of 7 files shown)


docs/roman/source_detection/arguments.rst
```diff
@@ -42,7 +44,7 @@ in a parameter file.
 * ``--bkg_boxsize``: If using `calc_threshold_img` size of box in pixels for
                      2D background / threshold images and if using
                      calc_threshold_2d, the size of the box used when detecting
-                     sources. Default is 3.
+                     sources. Default is 9.
```
schlafly, line 47: Good catch!  9 is pretty tiny too.  I think we use 256 for the equivalent of this in other surveys!

schlafly (approved):
Looks good to me, thanks!

bmorris3:
I won't update the changelog for this entry until https://github.com/spacetelescope/romancal/pull/987/ is merged (see https://github.com/spacetelescope/romancal/pull/987/files#diff-ff3c479edefad986d2fe6fe7ead575a46b086e3bbcf0ccc86d85efc4a4c63c79R75-R81).

docs/roman/source_detection/arguments.rst
bmorris3: This number will still need to be adjusted if there's a crowding, or the source is very bright, etc.

PR #988, opened 2023-11-13
fix issue where the path of the WebbPSF cache was not the same as the restore cache in the subsequent workflow

zacharyburnett:
This PR addresses an issue where the cached WebbPSF data was not actually restored:
https://github.com/spacetelescope/romancal/actions/runs/6829635768/job/18623576605?pr=980#step:3:10

This was because the paths were different; GitHub Actions caches must have the same key AND path to match an `actions/cache` call

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/.github/workflows/data.yml
+++ b/.github/workflows/data.yml
@@ -29,10 +29,10 @@ jobs:
       path: ${{ steps.path.outputs.path }}
     steps:
       - id: path
-        run: echo "path=${{ env.DATA_PATH }}/webbpsf-data" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


schlafly (approved):
LGTM, thanks!

.github/workflows/data.yml
```diff
@@ -45,12 +45,12 @@ jobs:
       - id: cache_check
         uses: actions/cache@v3
         with:
-          path: ${{ needs.webbpsf_path.outputs.path }}
+          path: ${{ needs.path.outputs.path }}
```
zacharyburnett, line 48: this was the offending line, for future reference (and other repos' workflows that may have the same problem)

bmorris3 (approved):
Thank you thank you!

PR #991, opened 2023-11-13
RCAL-699: Add regression testing for new jump detection and make new jump detection default to on

### LINKED ISSUE #957: Implement Regression testing for Jump detection
_Issue [RCAL-699](https://jira.stsci.edu/browse/RCAL-699) was created on JIRA by [William Jamieson](https://jira.stsci.edu/secure/ViewProfile.jspa?name=wjamieson):_

The Jump detection algorithm's authors have provided several of their simulations that were used to develop the algorithm. These should be turned into a set of regression tests for romancal.

WilliamJamieson:
Resolves [RCAL-699](https://jira.stsci.edu/browse/RCAL-699)

Closes #957

This PR enables jump detection within the Cas22 ramp fit by default and adds regression tests for it.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -216,9 +221,17 @@ def ols_cas22(self, input_model, readnoise_model, gain_model):
...
         var_poisson = output.variances[..., Variance.poisson_var]
         err = np.sqrt(var_poisson + var_rnoise)
+        dq = output.dq.astype(np.uint32)
+
+        ramp_dq = np.zeros(dq.shape[1:], dtype=np.uint32)
+        # Turn the jump detection DQ into a ramp DQ
+        for i in range(dq.shape[1]):
+            for j in range(dq.shape[2]):
+                if np.any(dq[:, i, j] == dqflags.group["JUMP_DET"]):
+                    ramp_dq[i, j] = dqflags.group["JUMP_DET"]
```
schlafly, line 231: I was imagining ramp_dq |= np.any(dq, axis=0) * dqflags.group['JUMP_DET'].
I hadn't appreciated that the new ramp fitting didn't include any other DQ flags.  We should add at least SATURATION and DONOTUSE or something back, but I'll review what needs to happen there and we shouldn't make it part of this PR.

schlafly (commented):
Thanks.  I left a minor nit that you may ignore.  I hadn't appreciated that we weren't setting any pixel dq flags before this; I will review separately what DQ flags we should try to bring over.  I could imagine more generally something like
```
dq |= ramp_dq
pixeldq = np.any(dq, axis=0)
```
but there may be things we don't want to bring over there.  I think we do want to record in some way when we have run the new jump detection; I think the least surprising would be cal_steps['JUMP'] = 'COMPLETE', even for this combined step.

WilliamJamieson:
This will very likely break several regression tests; in particular, the `wfi_pipeline` tests. However, as the regression tests are currently failing I cannot determine what needs to be "fixed".

Fixes will be needed because the jump detection step itself will not be run by default now.

WilliamJamieson:
> I think the least surprising would be cal_steps['JUMP'] = 'COMPLETE', even for this combined step.

I worry about confusion related to this. Maybe we should add a new "step" to `cal_steps` such as `RAMP_JUMP` or something like that so its distinguishable from the stand alone jump detection step.

romancal/regtest/scripts/create_jump_data.py
```diff
@@ -0,0 +1,148 @@
+import pickle
```
WilliamJamieson, line 1: I am not sure where to put this file, its just what I wrote in order to create all the regression test data. I thought it might want to within `romancal` so that if they need to be regenerated someone else would be able to do it.

ddavis-stsci (changes requested):
This needs to be updated. The basic input data for the regression tests are provided by INS. The tests for the combined jump/rampfitting Step should be the output of the pipeline using the INS input files run through the dark subtraction step.

schlafly:
> This needs to be updated. The basic input data for the regression tests are provided by INS. The tests for the combined jump/rampfitting Step should be the output of the pipeline using the INS input files run through the dark subtraction step.

The plan is to add William's new regression tests on top of the existing ones in test_wfi_pipeline (which I notice now still need decoration).  The new regression tests that William has added are additional to the ones in test_wfi_pipeline which I see as adequate to demonstrate the step.  Do you disagree?

ddavis-stsci:
That's fine but these are not regression tests. These are basically unit 
tests and not connected to the rest of the software.
If we want this in the tree as the default method this step needs to 
include the updated test_wfi_pipeline to show that it works
with our data.

ddavis-stsci (approved):

WilliamJamieson:
> That's fine but these are not regression tests.

I disagree. These definitely are regression tests. By definition regression testing is checking that the output of a piece of software does not change relative to some software change; that is, a software change does not induce a "regression" in the expected behavior of the program. The tests I believe @ddavis-stsci is referring to are ones which run ramp fitting on a given input and compare it to a known "good" static output. The isolated output of ramp fitting is being regressed against a fixed known good output.

Note that this is effectively what all the regression tests do for each of the pipeline steps. The only difference being that the input data for the test is chosen to be the known good output data of the previous test or provided by INS (typically for the first step in a given pipeline). In fact, the test in question can easily be updated to follow this pattern; however, I saw little point in following this trend in this case. The data I used to generate the test data in this case is directly derived from the data provided to me from the algorithm's designers and so was what I regressed my work against to show that the implementation in Romancal actually matches what the algorithm designers expected. Simply working against the output of the previous step introduces potentially large issues in verification that the implementation is correct as it lacks controls on whether or not the input data has had issues introduced by the previous step. The approach I took makes it possible to run the test under the circumstances the algorithm designers worked. 

> These are basically unit tests and not connected to the rest of the software.

If the above statement is referencing the fact the tests in question are isolated to the ramp fitting code and so therefore do not involve other parts of the pipeline, then this is an integration test as it integrates multiple parts of the program and checks the behavior is as expected. Which in the proposed scenario is checked via a regression test. I also want to point out that in this case, this sort of testing is in fact being done by the wfi pipeline tests.

PR #1009, opened 2023-11-17
Update resample test to use new files.

schlafly:
This PR addresses updates the resample regression test to use file names from the newest set of files.  It also takes advantage of the asn_from_list module to jettison some home-made asn-making code.


ddavis-stsci (approved):
Looks good. I added the no-change-log-needed label.

PR #1012, opened 2023-11-17
Point tweakreg to new files.  Move the asn file into artifactory.

schlafly:
This updates the tweakreg regression test to use new files.  It also removes the association file creation logic in favor of adding an association file to artifactory.

### DIFF
--- a/romancal/regtest/test_tweakreg.py
+++ b/romancal/regtest/test_tweakreg.py
@@ -1,5 +1,3 @@
-from io import StringIO
-
 import pytest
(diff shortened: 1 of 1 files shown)


ddavis-stsci (approved):
FYI: the suffix for tweakreg should be _tweakreg.asdf. The suffix _tweakregstep is the default when stpipe does not receive rules for constructing the suffix. 
We can fix this later.

schlafly:
Thanks, we'll fix that later.  Merging.

PR #1013, opened 2023-11-18
Update photutils dependency to released version.

schlafly:
This PR updates the photutils dependency to version 1.10.0 instead of a github development version.


ddavis-stsci (approved):

pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
     'gwcs >=0.18.1',
     'jsonschema >=4.8',
     'numpy >=1.22',
-    'photutils @ git+https://github.com/astropy/photutils.git',
+    'photutils >=1.9.0',
```
WilliamJamieson, line 22: The tests are failing, @bmorris3 does this require us to wait for a photutils release. If so can one happen? If not then the `romancal.lib.tests.test_psf` tests will need to be xfailed.

WilliamJamieson (changes requested):

pyproject.toml
schlafly: Larry is making the release this afternoon.

WilliamJamieson (approved):

schlafly:
Regression tests are failing only their usual bits, merging.

PR #1017, opened 2023-11-20
Turn on the apodized FFT interpolation

WilliamJamieson:
The original Goddard code effectively had this turned off, so in order to replicate those results this was turned off in `romancal` on purpose too even though was part of the Goddard code. This was reported to Goddard as a potential bug, and was confirmed to be a bug in the reference implementation by Goddard.

This PR enables the apodized FFT interpolation by default

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -129,6 +129,8 @@ refpix
 
 - Update cal_step, add suffix and add to the exposure pipeline [#890]
 
+- Enable apodized FFT interpolation by default. [#1017]
+
 resample
 --------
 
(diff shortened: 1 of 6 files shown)


schlafly:
This looks good to me.  It may also go through the regression tests fine since they have zeroed reference pixel channels, in which case we could go ahead and just merge it.  Did you run the regression tests?

WilliamJamieson:
This in fact passes the regression tests currently because amp33 is set to zero in those, see: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/597/

schlafly:
Great.  This is simple, approved by Goddard, and passes the regression tests; approving.

schlafly (approved):

PR #1024, opened 2023-11-21
Fix the pollution of stdout and stderr by ddtrace

### LINKED ISSUE #943: investigate ddtrace 2.0 logging output polluting pytest logs
This run (prior to ddtrace 2.0) contains expected output from the pytest run:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/397/pipeline/215

As seen in this later run (with ddtrace 2.0):
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/401/pipeline/215

The pytest output is polluted by hundreds of lines of log messages from `ddtrace`. I am not pasting the log messages in here because I'm not sure if this logging output contains private information.

WilliamJamieson:
Closes #943

When `pytest --ddtrace` was run, like in the Jenkins CLI, the logs were getting polluted by all sorts of output from `ddtrace` This was due to two effects.

1. `pytest`'s `log_cli_level` was not set, so `pytest` was sending all the logs it did not internally generate (it will still capture and display logs as part of a failure/error message) to stdout. This has now been set to the recommended "info" level, so only logs more serious then "info" will get this treatment.
2. `ddtrace` was encountering an error as part of its injection of datadog into python every time a package was imported after this test: https://github.com/spacetelescope/romancal/blob/29cbeda93779f02122f275ac6edb9932191d0b38/romancal/lib/tests/test_suffix.py#L17-L26 was run. Indeed, I could reproduce this behavior anywhere in the test suite if I added a test (or to a test) a call which eventually resulted in https://github.com/spacetelescope/romancal/blob/29cbeda93779f02122f275ac6edb9932191d0b38/romancal/stpipe/utilities.py#L57-L91 being called. I determined that this was caused by the imported modules state of python being polluted with a module that `ddtrace` could not inject datadog logging into. In this case, `load_local_pkg` searches the file system for a particular python file and then forces python to load and then execute that file as if it was a module. `ddtrace` is unable to do its thing in this case, which resulted in a module existing that it could not log. This caused it to send an error to stderr every time a new module, which was not previously imported, was imported.

This pull request fixes both of these issues by first setting the `log_cli_level` in the `pytest` configuration, and then refactoring the `load_local_pkg` to instead use Python's built-in `pkgutils` to walk the sub-packages of an already imported Python package. This walking cases the import of all the sub-packages via the normal Python import mechanism, which allows `ddtrace` to do its thing.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -87,7 +87,7 @@ bc0.conda_packages = [
 bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
     "pip install -e .[test]",
-    "pip install pytest-xdist pytest-sugar ddtrace",
+    "pip install pytest-xdist ddtrace",
     'echo "CRDS_CONTEXT = $(crds list --contexts $CRDS_CONTEXT --mappings | grep pmap)"',
 ]
 bc0.build_cmds = bc0.build_cmds + PipInject(env.OVERRIDE_REQUIREMENTS)
@@ -98,6 +98,7 @@ bc0.test_cmds = [
     "pytest -r sxf -n auto --bigdata --slow --webbpsf \
     --cov --cov-report=xml:coverage.xml \
     --ddtrace \
+    --color=no \
     --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope \
     --env=${artifactoryenv} ${pytest_args}",
     "curl -Os https://uploader.codecov.io/latest/linux/codecov",
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,8 @@ general
 - Use array comparison options (including ``nan`` equality) when
   comparing ``WCS`` objects during ``compare_asdf`` [#941]
 
+- Fix dynamic importing issue with the ``ddtrace`` package. [#1024]
+
 ramp_fitting
 ------------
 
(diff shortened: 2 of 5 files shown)


stscieisenhamer (approved):

schlafly:
I'm running a regtest with these changes here.  https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/491/ .

JenkinsfileRT
```diff
@@ -87,7 +87,7 @@ bc0.conda_packages = [
 bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
     "pip install -e .[test]",
-    "pip install pytest-xdist pytest-sugar ddtrace",
+    "pip install pytest-xdist ddtrace",
```
WilliamJamieson, line 90: These Jenkins workflow changes disable fancy symbols and coloring of the pytest output on Jenkins. These normally look great (and are useful when running locally) in a terminal because its inserting escaped unicode characters and color codes. However, it pollutes log output stored by Jenkins because all those escape sequences show up in that. Thus explicitly turning them off will make that log much easier to  read.

WilliamJamieson:
Sample Jenkins output before these changes:
Screenshot 2023-11-21 at 5 06 21 PM


After these changes:
Screenshot 2023-11-21 at 5 05 21 PM

It is even more readable on the "blue" interface: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/activity

schlafly (approved):
LGTM, thanks!

PR #1044, opened 2023-12-11
RCAL-724: Add additional information to `compare_asdf` diff reports.

### LINKED ISSUE #1027: compare_asdf function should report the filenames in its logging
_Issue [RCAL-724](https://jira.stsci.edu/browse/RCAL-724) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

The utility `compare_asdf` does not report what the filenames, or model names, are that are being compared. Reporting the input is an important bit of information that should be part of its output. Primary usecase is in tests to know what files have differences without looking through code to find what was being compared.

WilliamJamieson:
Resolves [RCAL-724](https://jira.stsci.edu/browse/RCAL-724)

Closes #1027

Currently the `compare_asdf` diff reports only return the deep diff results which indicate any differences between two asdf files. Why this is very useful, it does not tell the entire story; namely, it does not indicate what the diff was of. Issue #1027 indicates that it would be useful to know both the files being compared and the model names for the `roman_datamodels` included within those files (if applicable).

This PR adds a small header string to the start of the diff report that indicates the files input into `compare_asdf` and attempts to determine the `roman_datamodels` model described by the file (returning `Not a model` if the file does not describe a `roman_datamodel`), returning both for the "result" and "truth" inputs to `compare_asdf`. This is just a header stuck on top of the diff report string already being output, so it will not change any of our current regression tests or other reports. It will simply insert clarifying information into the report format.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 0.13.1 (unreleased)
 ===================
 
+general
+-------
+
+- Updated the ``compare_asdf`` diff reports to include descriptive information
+  about what is being compared. [#1044]
+
 dq_init
 -------
 
--- a/romancal/regtest/regtestdata.py
+++ b/romancal/regtest/regtestdata.py
@@ -6,6 +6,7 @@ import sys
 from difflib import unified_diff
 from glob import glob as _sys_glob
 from pathlib import Path
+from textwrap import dedent
 
 import asdf
 import astropy.table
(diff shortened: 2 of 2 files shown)


WilliamJamieson:
The diff reports will go from something like:
```
{'arrays_differ': {"root['roman']['data']": {'abs_diff': <Quantity 1.9999995 electron / s>,
                                             'n_diffs': 0,
                                             'worst_abs_diff': {'index': (0, 0),
                                                                'value': <Quantity 0.0002 electron / s>},
                                             'worst_fractional_diff': {'index': (0,
                                                                                 0),
                                                                       'value': <Quantity inf>}}}}
```
to
```
Diff report for:
    result file: /private/var/folders/wr/y4mz2y553jn8wwblt_9wtn240002qx/T/pytest-of-wjamieson/pytest-178/test_compare_asdf_large_0/test0.asdf
        model type: ImageModel
    truth file: /private/var/folders/wr/y4mz2y553jn8wwblt_9wtn240002qx/T/pytest-of-wjamieson/pytest-178/test_compare_asdf_large_0/test1.asdf
        model type: ImageModel

{'arrays_differ': {"root['roman']['data']": {'abs_diff': <Quantity 1.9999995 electron / s>,
                                             'n_diffs': 0,
                                             'worst_abs_diff': {'index': (0, 0),
                                                                'value': <Quantity 0.0002 electron / s>},
                                             'worst_fractional_diff': {'index': (0,
                                                                                 0),
                                                                       'value': <Quantity inf>}}}}
```

stscieisenhamer:
Yep, that is what was missing. The only issue is the terminology: "result" and "truth" are tied to our testing methodology. However, presumably, this is a general tool and I don't think this terminology will make sense to a general user who is just diffing files/models/asdftree.

I don't have any particular preference as to what a more general term is, `fitsdiff` uses "A" and "B" which are then used throughout the rest of the report. I don't recall if `compare_asdf` results are amenable to that type of labeling. May be sufficient to say "input 'filename/modelrepr/asdftree' compared against 'filename/modelrepr/asdftree'

schlafly (approved):
Looks good to me, thanks!

WilliamJamieson:
Regression tests pass: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/596/

PR #1046, opened 2023-12-11
add documentation for static preview script

zacharyburnett:
Resolves [RCAL-731](https://jira.stsci.edu/browse/RCAL-731)

This PR adds documentation for `roman_static_preview`

**Checklist**
- [ ] added entry in `CHANGES.rst` under the corresponding subsection
- [ ] ~updated relevant tests~
- [x] updated relevant documentation
- [ ] updated relevant milestone(s)
- [x] added relevant label(s)
- [ ] ~ran regression tests, post a link to the Jenkins job below. [How to run regression tests on a PR](https://github.com/spacetelescope/romancal/wiki/Running-Regression-Tests-Against-PR-Branches)~

PR #1049, opened 2023-12-15
add instructions to clone to install Git LFS in the repository

zacharyburnett:
#1046 added large image files that went into Git LFS. In order to use LFS, the git hooks must be installed to the repository (`git lfs install`). This PR adds that step to the cloning instructions in the README. 

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ To install your own copy of the code into that environment, you first need to fo
     cd <where you want to put the repo>
     git clone https://github.com/spacetelescope/romancal
     cd romancal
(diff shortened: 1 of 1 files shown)


ddavis-stsci (changes requested):
I (and Nadia?) think we need to reevaluate why we need git-lsf. This is just for some small ~48K images so it seems to be overkill to me. 

My error, the largest is 890K. I'm not sure we need the full resolution image for the documentation and a smaller one should be fine. We can also save space using a jpg file.

PR #1058, opened 2024-01-08
Add unit tests and outlier module refactoring.

mairanteodoro:
Resolves [RCAL-502](https://jira.stsci.edu/browse/RCAL-502)

This PR adds unit tests to the newly added outlier detection pipeline module.
The regression test for the outlier detection step will be added soon.

**Regression tests results**
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/516/

**Checklist**
- [X] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/romancal/outlier_detection/outlier_detection_step.py
+++ b/romancal/outlier_detection/outlier_detection_step.py
@@ -1,6 +1,5 @@
 """Public common step definition for OutlierDetection processing."""
 
-import contextlib
 import os
 from functools import partial
 
@@ -55,19 +54,46 @@ class OutlierDetectionStep(RomanStep):
 
         self.input_models = input_models
         self.input_container = isinstance(self.input_models, ModelContainer)
-        # Setup output path naming if associations are involved.
-        asn_id = None
-        with contextlib.suppress(AttributeError, KeyError):
-            asn_id = self.input_models.meta.asn_table.asn_id
-        if asn_id is None:
-            asn_id = self.search_attr("asn_id")
-        if asn_id is not None:
-            # TODO: revisit this once we have a better idea of the associations
-            _make_output_path = self.search_attr("_make_output_path", parent_first=True)
+        self.skip = False
 
+        # make sure input is a ModelContainer
+        if self.input_container:
+            return self._run_process()
+
+        # if input is not a ModelContainer, log warning message and raise TypeError
+        self.log.warning(
+            "Input is not a ModelContainer. Outlier detection step will be SKIPPED."
+        )
+        raise TypeError
+
+    def _run_process(self):
+        # check conditions for valid input
+        self.valid_input = len(self.input_models) >= 2 and all(
+            model.meta.exposure.type == "WFI_IMAGE" for model in self.input_models
+        )
+
+        if self.valid_input:
+            return self._run_outlier_detection()
+
+        # if input is not valid then log a warning message and
+        # skip outlier detection step
+        self.log.warning(
+            "Input contains either < 2 elements or the wrong exposure type \
+            (i.e. meta.exposure.type != 'WFI_IMAGE')."
+        )
+        for model in self.input_models:
+            model.meta.cal_step.outlier_detection = "SKIPPED"
+        self.skip = True
+        return self.input_models
+
+    def _run_outlier_detection(self):
+        # Setup output path naming if associations are involved.
+        asn_id = self.input_models.asn_table.get("asn_id", None)
+        if asn_id is not None:
+            _make_output_path = self.search_attr("_make_output_path", parent_first=True)
             self._make_output_path = partial(_make_output_path, asn_id=asn_id)
 
-        # Setup outlier detection parameters
+        detection_step = outlier_detection.OutlierDetection
         pars = {
             "weight_type": self.weight_type,
             "pixfrac": self.pixfrac,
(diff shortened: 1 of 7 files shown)


romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,42 +1,358 @@
...
+    # The sky rotation expects values in deg.
+    # This should be removed when models work with quantities.
+    model = (
+        (Scale(0.1 / 3600) & Scale(0.1 / 3600))
+        | SphericalToCartesian(wrap_lon_at=180)
+        | rot
+        | CartesianToSpherical(wrap_lon_at=360)
+    )
+    model.name = "v23tosky"
+    return model
```
nden, line 60: Why the copy/duplicaiton of this function? Can you call the one in [pointing](https://github.com/spacetelescope/romancal/blob/main/romancal/assign_wcs/pointing.py#L6)?

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,42 +1,358 @@
...
+        l2 = maker_utils.mk_level2_image(shape=(100, 100))
+        l2_im = rdm.ImageModel(l2)
+        _create_wcs(l2_im)
+        l2_im.meta.wcsinfo.vparity = -1
+        return l2_im
+
+    return _base_image
+
+
+def create_asn_file(tmp_path, members_mapping=None):
```
nden, line 133: This looks like another [code duplicate](https://github.com/spacetelescope/romancal/blob/main/romancal/tweakreg/tests/test_tweakreg.py#L88). Make it a fixture ?

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -55,19 +54,46 @@ def process(self, input_models):
...
+
+        # make sure input is a ModelContainer
+        if self.input_container:
+            return self._run_process()
+
+        # if input is not a ModelContainer, log warning message and raise TypeError
+        self.log.warning(
+            "Input is not a ModelContainer. Outlier detection step will be SKIPPED."
+        )
+        raise TypeError
```
nden, line 67: I don't think raising an error is desirable in this case. It may be better to issue a warning and skip the step.

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -55,19 +54,46 @@ def process(self, input_models):
...
+
+    def _run_process(self):
+        # check conditions for valid input
+        self.valid_input = len(self.input_models) >= 2 and all(
```
nden, line 71: Is there a need for `valid_input` to be an instance variable, i.e. is `self` necessary here?

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,42 +1,358 @@
...
+    assert exec_info.type == TypeError
+
+
+def test_outlier_skips_step_on_invalid_number_of_elements_in_input(base_image):
+    img = base_image()
+
+    step = OutlierDetectionStep()
+    step.input_models = ModelContainer([img])
+
+    step._run_process()
```
nden, line 206: Should `step.process()` be called here to exercise the actual interface?

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,42 +1,358 @@
...
+    img_1.meta.filename = "img_1.asdf"
+    img_1.save(tmp_path / "img_1.asdf")
+    img_2 = base_image()
+    img_1.meta.filename = "img_2.asdf"
+    img_2.save(tmp_path / "img_2.asdf")
+
+    asn_filepath = create_asn_file(tmp_path)
+
+    step = OutlierDetectionStep()
+    step.input_models = ModelContainer(asn_filepath)
```
nden, line 254: Is `outlier_detection` supposed to run with associations as input?
If so the handling of the association should be done within the `process` function.

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -87,81 +113,36 @@ def process(self, input_models):
...
+                #  crf_file = os.path.basename(crf_path)
+                #  outlr_path = crf_path.replace(crf_file, outlr_file)
+                outlr_file = model.meta.filename.replace("cal", "outlier_i2d")
+                blot_path = crf_path.replace("crf", "blot")
+                median_path = blot_path.replace("blot", "median")
+
+                for fle in [outlr_file, blot_path, median_path]:
+                    if os.path.isfile(fle):
+                        os.remove(fle)
+                        self.log.debug(f"    {fle}")
```
nden, line 146: Does it make sense to put the code dealing with cleaning of generated files in a separate function?

schlafly (changes requested):
I may have missed it, but do any of these tests test the ability to detect outliers?  I think you're always sending all-zero images and we should never detect an outlier; in principle we could test that.  It would also be good to set a particular pixel to be high and see if it gets detected, or something?

You also have pretty good control here so I could imagine picking a value for the ivar of the image and the snr parameter and checking that when inserting an outlying pixel that it gets identified at the right snr.  On the other hand, that flows through drizzle & coadding and so might be a bit harder to predict than I'm implying here.  But at least showing that a really obvious outlier gets flagged and that nothing gets flagged when all pixels are the same or consistent with noise or something seems like a test we should definitely have in the outlier detection unit tests?

mairanteodoro:
> I may have missed it, but do any of these tests test the ability to detect outliers?

Yes, I agree that we should have a test for the detection functionality. That will be added after I have worked on [RCAL-725](https://jira.stsci.edu/browse/RCAL-725), which deals with the actual outlier detection functionality.

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -305,6 +133,9 @@ def test_outlier_init_default_parameters(pars, base_image):
     assert step.make_output_path == pars["make_output_path"]
     assert step.resample_suffix == f"_outlier_{pars['resample_suffix']}.asdf"
 
+    # clean up
+    [x.unlink() for x in Path(os.getcwd()).glob("*.asdf")]
```
schlafly, line 137: Does this and its relatives feel dangerous?

schlafly (commented):
Okay, fine with me, but let's wait for Nadia's comments.  And for my two cents, the most important tests are the ones that test something about the core outlier detection algorithm, and not the API around it.  Though of course both are valuable.

mairanteodoro:
> Okay, fine with me, but let's wait for Nadia's comments. And for my two cents, the most important tests are the ones that test something about the core outlier detection algorithm, and not the API around it. Though of course both are valuable.

Agreed.

nden (approved):
LGTM
Is a regression test in the plans as well, possibly as part of the hlp pipeline?

mairanteodoro:
> LGTM Is a regression test in the plans as well, possibly as part of the hlp pipeline?

Yes, I will implement the regression test soon after I'm done with RCAL-725, if that's OK.

schlafly (approved):

PR #1065, opened 2024-01-18
Updates to Romancal to support the auto generated roman_datamodels models.

WilliamJamieson:
This PR makes all the fixes to `romancal` to support the autogenerated `roman_datamodels` models, see spacetelescope/roman_datamodels#303.

Note there is one regression test failing, but it appears that it should be failing on main but currently is not; namely, https://github.com/spacetelescope/romancal/blob/1c2c7d26168779e1cc3f9a0b3fe49c6017eb1a86/romancal/regtest/test_wfi_flat_field.py#L38-L69. It should be failing because even on main, the model produced by the `romancal` step is entirely different both in datamodel type and shapes of all the major data arrays.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


romancal/datamodels/tests/test_datamodels.py
```diff
@@ -85,15 +84,15 @@ def _setup_list_of_l2_files(n, obj_type, tmp_path):
                 / f"test_model_container_input_as_list_of_filepaths_{i:02}.asdf"
             )
             # create an ASDF file with an L2 model
-            utils.mk_level2_image(filepath=filepath, shape=(100, 100))
+            rdm.WfiImageModel.make_default(shape=(8, 100, 100), filepath=filepath)
```
schlafly, line 87: Not sure it is meaningful, but the addition of the 8 here surprises me; I expect WfiImageModels to be 2 dimensionsal.

romancal/dq_init/dq_init_step.py
```diff
@@ -44,24 +43,25 @@ def process(self, input):
...
-                    )
+                    output_model[key] = value.astype(output_model[key].dtype)
                 else:
-                    input_ramp[key] = input_model.__getattr__(key)
+                    output_model[key] = value
+
+            # Remove the "resultantdq" attribute as this is accidentally moved and breaks
+            # the regression tests
+            if "resultantdq" in output_model.model_extra:
+                del output_model.model_extra["resultantdq"]
```
schlafly, line 63: We would make a separate ticket for this one?

romancal/flatfield/flat_field.py
```diff
@@ -123,7 +123,10 @@ def apply_flat_field(science, flat):
         science["var_flat"] = np.zeros(shape=science.data.shape, dtype=np.float32)
         science.var_flat = science.data**2 / flat_data_squared * flat_err**2
 
-    science.err = np.sqrt(science.var_poisson + science.var_rnoise + science.var_flat)
+    err = np.sqrt(science.var_poisson + science.var_rnoise + science.var_flat)
+    science.err = u.Quantity(
+        value=err.value, unit=u.Unit(err.unit.to_string()), dtype=err.dtype
+    )
```
schlafly, line 129: Are the units for some reason not propagating sensibly here?

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -271,10 +273,8 @@ def create_image_model(input_model, image_info):
 
     # Create output datamodel
     # ... and add all keys from input
-    meta = {}
-    meta.update(input_model.meta)
+    meta = input_model.meta.model_dump()
     meta["cal_step"]["ramp_fit"] = "INCOMPLETE"
-    meta["photometry"] = maker_utils.mk_photometry()
```
schlafly, line 277: I think I infer that WfiImageModel.make_default(...) is now doing this for us?

schlafly (commented):
Thanks William.  This looks good to me.  I left some comments inline.  Most of this is pretty straightforward translation (good!), though there were a few more surprising things.  I think the ~ugliest one is instantiating image models with an unused (?) dimension for the number of resultants.  Maybe also L3 models down the line.  It may also be worth discussing whether there are alternatives to constructing full models to pull out source detection and resample sections.

romancal/datamodels/tests/test_datamodels.py
WilliamJamieson: This is something we should talk about. The issue is there are 3 dimensional fields such as `amp33` and `border_ref_pix_left`. So it needs a third dimension to properly flush those values out when guessing.

Unfortunately, the `make_default` function by its nature is not as "intelligent" about what shapes are what, so it requires that it has enough dimensions to fill out every array of the model.

romancal/dq_init/dq_init_step.py
WilliamJamieson: Probably

romancal/flatfield/flat_field.py
WilliamJamieson: Correct. One is getting units that should be "right" but which astropy is indicating are not equivalent to the units stored in the validator. It is something I need to investigate in more detail

romancal/ramp_fitting/ramp_fit_step.py
WilliamJamieson: Correct.

romancal/resample/resample.py
```diff
@@ -186,7 +189,9 @@ def resample_many_to_many(self):
         """
         for exposure in self.input_models.models_grouped:
             output_model = self.blank_output
-            output_model.meta["resample"] = mk_resample()
+            output_model.meta[
+                "resample"
+            ] = datamodels.WfiMosaicModel.make_default().meta.resample
```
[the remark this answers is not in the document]
WilliamJamieson: This is a product of "getting it working". I think this is not needed, but I'll need to do some testing.

romancal/resample/resample.py
```diff
@@ -331,17 +340,19 @@ def resample_many_to_one(self):
...
-        img.meta.background["level"] = 0
-        img.meta.background["subtracted"] = True
+        # Should background be added to wfi_image's meta as an optional feature?
+        class Background(BaseDataModel):
+            schema_uri: ClassVar[None] = None
+
+            level: float
+            subtracted: bool
+
+        img.meta.background = Background(level=0, subtracted=False)
```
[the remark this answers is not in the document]
WilliamJamieson: I think `background` should have its own entry in the schemas, but it currently does not. Whenever we introduce that, this complexity can be removed.

romancal/skymatch/tests/test_skymatch.py
```diff
@@ -211,12 +211,12 @@ def test_skymatch(wfi_rate, skymethod, subtract, skystat, match_down):
...
-        assert im.meta.background.method == skymethod
-        assert im.meta.background.subtracted == subtract
+        assert im.meta.background["method"] == skymethod
+        assert im.meta.background["subtracted"] == subtract
 
         # test computed/measured sky values if level is set:
-        if not np.isclose(im.meta.background.level.value, 0):
-            assert abs(im.meta.background.level.value - rlev) < 0.01
+        if not np.isclose(im.meta.background["level"].value, 0):
+            assert abs(im.meta.background["level"].value - rlev) < 0.01
```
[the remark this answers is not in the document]
WilliamJamieson: Alright, so this goes back to my above comment. I attempted to make `background` act like it was a data model, but somewhere in skymatch it is being attached as a `dict` meaning it stays a `dict` (so no dot access). I did not trace this back into skymatch.

romancal/source_detection/source_detection_step.py
```diff
@@ -193,7 +192,9 @@ def process(self, input):
 
             # tweakreg_catalog_name will be saved to the final output file,
             # while tweakreg_catalog is intended to be deleted by TweakRegStep
-            input_model.meta["source_detection"] = maker_utils.mk_source_detection()
+            input_model.meta["source_detection"] = rdm.WfiImageModel.make_default(
+                _shrink=True
+            ).meta.source_detection
```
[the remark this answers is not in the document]
WilliamJamieson: Correct `_shrink=True` limits the image arrays created to `(8, ...)`. Since this is only there to set `source_detection` (which I think might be redundant) and `source_detection` contains no arrays, this is just to make things cheaper.

PR #1066, opened 2024-01-18
fix compare_asdf model difference

braingram:
As noted https://github.com/spacetelescope/romancal/pull/1065#issue-2088562393
`compare_asdf` fails to identify differences if the `DataModel` type differs. This PR updates `compare_asdf` to identify these differences and adds a test for the fix.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


schlafly:
This looks good to me.  @ddavis-stsci , are we fine starting to break the regression tests now for a bit as schema changes, ... start to go in?

ddavis-stsci:
> This looks good to me. @ddavis-stsci , are we fine starting to break the regression tests now for a bit as schema changes, ... start to go in?

The regression tests are already broken (gwcs required) so sure go ahead.

PR #1070, opened 2024-01-24
Make remove_trends more robust against outliers / CRs.

schlafly:
This PR solves an issue where information in the border pixels in L1 files can lead to weird amp discontinuities in reference pixel corrected data.  This is not a PR that we should merge, but it illustrates what we need to do to improve the robustness of the reference pixel correction algorithm.

The trend removal algorithm is conceptually simple; it fits a linear function to the reference pixels in amplifier as a function of readout time and subtracts it.  This algorithm nevertheless has two important issues.
1. It uses the pixel value being zero as a signal for the pixel being invalid and requiring exclusion from the fit.
2. It is not a robust fit and therefore can pull the whole amp around if there are serious outliers.

The use of zero as a signal seems okay but is badly exacerbated by the current CRDS read noise reference files which have read noises artificially low (much less than 1 DN).  This means that the images have no read noise, so only Poisson noise from signal pixels contributes to the noise.  The reference pixels have no signal, so they are just constants + CRs.  The remove_trends step sees the pixels after a constant has been removed; this makes the reference pixels zeros + CRs.  This means that almost all of the reference pixels are ignored in the current Roman L1 files and _only_ the ones corrupted by CRs are actually seen.  Oops!  This PR "fixes" that issue by artificially leaving on a signal of 1e-9 DN for pixels that are originally non-zero but are zeroed after constant subtraction.  That's a dumb hack and a better approach would check the "offset" file correctly to see if the offset was also zero.

This could also be fixed by fixing the CRDS reference files, though some change along the lines of this PR is still needed since with real read noises of ~5 DN we will still be throwing away meaningful numbers of reference pixels for no reason with the current != 0 mask.

Note that for most amps there are only reference pixels at the beginning and end, and so the fit could basically be taking a median of the beginning and end reference pixels and fitting a line to those two points.  That would be fairly robust and would do what is needed.

This PR also has a little change to the linear fitting to make it robust to outliers.  This is not as important as the change to include pixels with values equal to zero, but is also a good idea.

The Goddard code has a hook that imagines calling an external routine to identify and remove cosmic rays.  We need that to eventually exist.  Currently I think CRs in the border pixels will still have some impact blurring out and affecting science pixels, just not nearly as badly as they previously had.

### DIFF
--- a/romancal/refpix/data.py
+++ b/romancal/refpix/data.py
@@ -12,7 +12,7 @@ from enum import IntEnum
 
 import numpy as np
 from astropy import units as u
-from scipy import fft
+from scipy import fft, stats, optimize
 
 
 class Const(IntEnum):
@@ -224,6 +224,13 @@ class StandardView(BaseView):
 
         # Apply the offset to the data (in place)
         data -= offset
+        m = (data == 0) & (offset != 0)
+        data[m] = 1e-9  # hack!  But remove_trends uses data = 0 as a signal
+        # for missing data, so data whose offset subtracts exactly is treated
+        # as missing rather than as truly zero, which is bad.  This perturbs
+        # that slightly.
+        # it would be better to have the frame_not_zero checks look at
+        # the offsets explicitly.
 
         self.offset = offset.reshape(rows, columns)
 
@@ -361,6 +368,9 @@ class ChannelView(BaseView):
         t_ref = t[REF_ROWS, :]
         not_zero = self.data != 0
 
+        def chi(param, datax, datay, sdev):
+            return (datay - param[1] - param[0] * datax) / sdev
+
         # fit needs to be done for each channel and frame separately
         for chan_data, chan_not_zero in zip(self.data, not_zero):
             for frame_data, frame_not_zero in zip(chan_data, chan_not_zero):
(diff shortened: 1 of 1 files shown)


schlafly:
Pinging @WilliamJamieson and @rcosenti-stsci as potentially interested.

schlafly:
Maybe @tddesjardins as well.

romancal/refpix/data.py
```diff
@@ -224,6 +224,13 @@ def remove_offset(self) -> StandardView:
 
         # Apply the offset to the data (in place)
         data -= offset
+        m = (data == 0) & (offset != 0)
+        data[m] = 1e-9  # hack!  But remove_trends uses data = 0 as a signal
+        # for missing data, so data whose offset subtracts exactly is treated
+        # as missing rather than as truly zero, which is bad.  This perturbs
+        # that slightly.
+        # it would be better to have the frame_not_zero checks look at
+        # the offsets explicitly.
```
WilliamJamieson, line 233:
```suggestion
        # HACK!
        # `remove_trends` uses data = 0 as a signal for missing data, so
        #    data whose offset subtracts exactly is treated as missing rather
        #    than as truly zero, which is bad.  This perturbs that slightly.
        # It would be better to have the frame_not_zero checks look at
        #    the offsets explicitly.
        data[m] = 1e-9
```
This cleans up the comment a bit. It makes it make more sense to me.

romancal/refpix/data.py
```diff
@@ -375,8 +385,19 @@ def remove_trends(self) -> ChannelView:
                 if x_vals.size < 1 or y_vals.size < 1:
                     continue
 
-                # Perform the fit using a 1st order polynomial, (i.e. linear fit)
-                m, b = np.polyfit(x_vals, y_vals, 1)
+                # # Perform the fit using a 1st order polynomial, (i.e. linear fit)
+                # m, b = np.polyfit(x_vals, y_vals, 1)
+
```
WilliamJamieson, line 390: Remove the commented out code and add a small explanation of what is going on here.

WilliamJamieson (changes requested):
Some basic comments.

Also, it looks like the unit tests need to be updated slightly as they are now failing.

It maybe worth removing the "test against Goddard code" related tests before this PR is merged, since this is a divergence from that code base and we now of regression testing in place to realistically test this code. Let me know if you want me to open a PR doing that work (we planned on doing that once solid regression testing was put in place).

schlafly:
Closing this as RTB pursues alternative approaches and as we have adopted romanisim to get rid of this issue in our simulated files.

PR #1072, opened 2024-01-25
Provide a default WCS for ramp fitting.

schlafly:
Units test started failing when we made the WCS a required argument, because the unit tests relied on being able to construct ImageModels from default RampModels.  The latter don't have default WCSes, but the former require them, so the code to construct the one from the other needs to add a default WCS.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


nden (approved):
This will do it.

PR #1084, opened 2024-01-30
Consolidate regression tests into fewer files.

schlafly:
This PR consolidates the regression tests to use fewer files and adds a script to create all of the regression test files from a directory of input files.

Changes to the regression tests:
- Point ramp fitting tests to WFI01 and darkcurrent subtracted files.  We had these pointing to weird files to deal with a caching issue and a shape issue in the run-up to B12; we don't need to continue doing that.
- Point test_wfi_dq_init, test_wfi_flat_field, test_wfi_saturation tests to the one fiducial spectroscopic regression test file; these formerly used a second file but I'm not sure what the motivation for that originally was.
- Rename the munged L1 file with a different time to be clear that this is a file we are making with a weird time rather than a file from RTB.
- Move some test_wfi_photom files to use the nominal imaging L1 file rather than the grism file.

Then there is a big new script that generates all of the regtest files from the input files, which following the above changes are:

```
# r0000101001001001001_01101_0001_WFI01 - default for most steps
# r0000201001001001001_01101_0001_WFI01 - equivalent for spectroscopic data
# r0000101001001001001_01101_0002_WFI01 - a second resample exposure, only cal step needed
# r0000101001001001001_01101_0003_WFI01 - for ramp fitting; truncated image
# r0000201001001001001_01101_0003_WFI01 - for ramp fitting; truncated spectroscopic
#                                         we need only darkcurrent & ramp fit for these
#
# r00r1601001001001001_01101_0001_WFI01 - special 16 resultant file, imaging, only need cal file
# r10r1601001001001001_01101_0001_WFI01 - special 16 resultant file, spectroscopy, only need cal file
# roman_dark_WFI01_IMAGE_STRESS_TEST_16_MA_TABLE_998_D1 - special dark for 16 resultant file
```

That script became a fairly messy mixture of python and shell script.  I could make it all python which might be better, though it will still have a bunch of pipeline invocations and copying that is nicer in shell.  One runs this script by:
- copying the input files above from central store/roman/eschlafly/regtestinput into a local directory
- making an output directory (e.g., data)
- running make_regtestdata.sh data

I think this would be uncontroversial except for the fact that it involves changing the names of some files.  These are presumably embedded in past requirement tickets and may be annoying to change.

The regtest directory generated by this script is 76 GB.  It may be possible to be a bit more efficient, as I have not been careful with checking that each of the intermediate steps that get saved for the processing of the "primary" imaging and spectroscopic files is actually used.  But this is probably pretty close to the minimum the current tests need.

I note that the test_wfi_flat_field/test_flat_field_grism_step is testing the flat field applied to grism data.  It's a bit annoying because a normal pipeline run doesn't run the flat fielding on grism data, and so these files aren't ordinarily created.  The regtest runs the step on an L1 file, which the flat field step shouldn't be able to process in the first place, but presumably since the step just ends up getting skipped this doesn't cause problems.  I have left in that behavior, but this test doesn't have any SOC test decoration and so I think I'd rather just delete this test and the corresponding lines in the script to generate its file.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/scripts/make_regtestdata.sh
+++ b/romancal/scripts/make_regtestdata.sh
@@ -1,0 +1,161 @@
+#!/bin/bash
+
+# takes one argument: the directory into which to start putting the regtest files.
+
+# input files needed:
+# r0000101001001001001_01101_0001_WFI01 - default for most steps
+# r0000201001001001001_01101_0001_WFI01 - equivalent for spectroscopic data
+# r0000101001001001001_01101_0002_WFI01 - a second resample exposure, only cal step needed
+# r0000101001001001001_01101_0003_WFI01 - for ramp fitting; truncated image
+# r0000201001001001001_01101_0003_WFI01 - for ramp fitting; truncated spectroscopic
+#                                         we need only darkcurrent & ramp fit for these
+#
+# Not yet found / looked at these files.
+# r00r1601001001001001_01101_0001_WFI01 - special 16 resultant file, imaging, only need cal file
+# r10r1601001001001001_01101_0001_WFI01 - special 16 resultant file, spectroscopy, only need cal file
+# roman_dark_WFI01_IMAGE_STRESS_TEST_16_MA_TABLE_998_D1 - special dark for 16 resultant file
+
+outdir=$1
+
+# set up the directory structure
+mkdir -p $outdir/roman-pipeline/dev/WFI/image
+mkdir -p $outdir/roman-pipeline/dev/truth/WFI/image
+mkdir -p $outdir/roman-pipeline/dev/WFI/grism
+mkdir -p $outdir/roman-pipeline/dev/truth/WFI/grism
+
+
+# most regtests; run the pipeline and save a lot of results.
+for fn in r0000101001001001001_01101_0001_WFI01 r0000201001001001001_01101_0001_WFI01
+do
+    echo "Running pipeline on ${fn}..."
+    strun roman_elp ${fn}_uncal.asdf --steps.dq_init.save_results True --steps.saturation.save_results True --steps.linearity.save_results True --steps.dark_current.save_results True --steps.rampfit.save_results True --steps.assign_wcs.save_results True --steps.photom.save_results True --steps.refpix.save_results True --steps.flatfield.save_results True --steps.assign_wcs.save_results True
+    [[ ${fn} = r00002* ]] && dirname="grism" || dirname="image"
+    cp ${fn}_*.asdf $outdir/roman-pipeline/dev/WFI/$dirname/
+    cp ${fn}_*.asdf $outdir/roman-pipeline/dev/truth/WFI/$dirname/
+
+    # uncal file doesn't belong in truth.
+    rm $outdir/roman-pipeline/dev/truth/WFI/image/${fn}_uncal.asdf
+done
+
+
+# truncated files for ramp fit regtests
+for fn in r0000101001001001001_01101_0003_WFI01 r0000201001001001001_01101_0003_WFI01
+do
+    echo "Running pipeline on ${fn}..."
+    strun roman_elp ${fn}_uncal.asdf --steps.dark_current.save_results True --steps.rampfit.save_results True
+    [[ ${fn} = r00002* ]] && dirname="grism" || dirname="image"
+    cp ${fn}_darkcurrent.asdf $outdir/roman-pipeline/dev/WFI/$dirname/
+    cp ${fn}_rampfit.asdf $outdir/roman-pipeline/dev/truth/WFI/$dirname/
+done
+
+
+# second imaging exposure
+strun roman_elp r0000101001001001001_01101_0002_WFI01_uncal.asdf
+cp r0000101001001001001_01101_0002_WFI01_cal.asdf $outdir/roman-pipeline/dev/WFI/image/
+
+
+# resample regtest; needs r0000101001001001001_01101_000{1,2}_WFI01_cal.asdf
+# builds the appropriate asn file and calls strun with it
+echo "Creating regtest files for resample..."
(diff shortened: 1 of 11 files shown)


schlafly:
@nden , @ddavis-stsci , this is the regtest generation script I've been working on the past few days, together with a handful of file name changes to romancal to try to get us to use as few files as possible.  This would make it easy to blow away the current regtest directory and regenerate all of the files from the L1s whenever we wanted.

romancal/scripts/make_regtestdata.sh
```diff
@@ -0,0 +1,160 @@
...
+
+
+# resample regtest; needs r0000101001001001001_01101_000{1,2}_WFI01_cal.asdf
+# builds the appropriate asn file and calls strun with it
+echo "Creating regtest files for resample..."
+asn_from_list r0000101001001001001_01101_0001_WFI01_cal.asdf r0000101001001001001_01101_0002_WFI01_cal.asdf -o mosaic_asn.json --product-name mosaic
+strun romancal.step.ResampleStep mosaic_asn.json --rotation=0 --output_file=mosaic.asdf
+cp mosaic_asn.json $outdir/roman-pipeline/dev/WFI/image/
+cp mosaic_resamplestep.asdf $outdir/roman-pipeline/dev/truth/WFI/image/
+
```
ddavis-stsci, line 63: mosaic_resamplestep.asdf should be a valid Roman filename

romancal/scripts/make_regtestdata.sh
```diff
@@ -0,0 +1,160 @@
...
+cp ${basename}_changetime_flat.asdf $outdir/roman-pipeline/dev/truth/WFI/image
+
+
+echo "Creating regtest files for tweakreg tests..."
+# tweakreg regtest; call tweakreg on a cal file.  Make appropriate asn file.
+asn_from_list r0000101001001001001_01101_0001_WFI01_cal.asdf -o tweakreg_asn.json --product-name tweakreg
+strun romancal.step.TweakRegStep tweakreg_asn.json
+cp tweakreg_asn.json $outdir/roman-pipeline/dev/WFI/image/
+cp r0000101001001001001_01101_0001_WFI01_tweakregstep.asdf $outdir/roman-pipeline/dev/truth/WFI/image/
+
```
ddavis-stsci, line 92: This should be _tweakreg.asdf

romancal/scripts/make_regtestdata.sh
```diff
@@ -0,0 +1,160 @@
...
+wcsinfo = model.meta.wcsinfo
+if wcsinfo.ra_ref >= 350.0:
+    delta[0] *= -1.0
+if wcsinfo.dec_ref >= 80.0:
+    delta[1] *= -1.0
+wcsinfo.ra_ref += delta[0]
+wcsinfo.dec_ref += delta[1]
+model = AssignWcsStep.call(model)
+model.to_asdf(f'${basename}_cal_repoint.asdf')"
+    [[ ${basename} = r00002* ]] && dirname="grism" || dirname="image"
```
ddavis-stsci, line 151: Isn't the above just the code from test_wfi_pipeline?

ddavis-stsci (commented):
Overall the script looks fine. There does seem to be a lot of unnecessary copies of large files. 

Most of the file name changes are fine as long as INS will produce files with those names. 

We should mostly try to use valid roman file names. For the cases where we have renamed file (e.g. ALL_SATURATED) it would be nice if INS would produce these files as part of the files they produce.

For r0000201001001001001_01101_0001_WFI17,  This is from JE so I'm not sure why this was selected over r0000201001001001001_01101_0001_WFI01. I'd check with him before changing the file. 

I'm also a bit uncomfortable about having this in the scripts directory that is public. I'm not sure that a general user would ever need this.

schlafly:
The ramp fitting regression test files (e.g. WFI17) were driven by a weird problem we were having with plwishmaster seeming to cache the incorrect files.  It was complaining about file mismatches after Jonathan had just explicitly pushed the correct files.  It was surmounted by throwing some new files in.  This is an opportunity to move away from that.

Yes, I did copy the code from test_wfi_pipeline into the regtest script for the pointing tweak; I didn't see a robust way around that.  My understanding is that the test_wfi_pipeline code does some tweaks to the pointing on the fly and then compares the tweaked WCS with the regression test file.  So I needed to make the same tweaks when I generate the truth file for the regression test.

Maybe rather than the current _ALL_SATURATED endings I'll introduce a new product r99999 that we make that will go with r00001 and r00002 from RTB.  My preference is to keep this kind of metadata stuff in our hands and let them focus on making good images.  The ugly _ALL_SATURATED choice has the benefit that it's clear what the file is from looking at it; an r99999 scheme will lose that and we'll need a lookup table to see which file is which.  But that's fine.

My motivation for bundling this with the main code is mostly so that other developers who want to add regression files can also add the associated scripts to this file, and so that others can run it more easily.

schlafly:
This PR, in conjunction with the romanisim script here:
https://github.com/spacetelescope/romanisim/pull/102/files#diff-0a1a564121f526266b39935128c307e905898e6170d6bf3f591027a8360dea97
makes a complete set of regtest files from romanisim input.

I've added one conceptually unrelated change to tweakreg to use mid_time instead of target.proper_motion_epoch because I couldn't bring myself to populate the latter in the L1 files for these kinds of images.

nden (commented):
Overall it LGTM. I think the real test is tunning it and passing tests.
Are we going to have to update test tickets with artifacts and test description for next build?

schlafly:
I have a set of files generated by this script in /grp/roman/eschlafly/regtest-20240313 that passed tests with this PR if you want to take a look.

nden (approved):

ddavis-stsci (approved):

PR #1085, opened 2024-01-30
Rcal-622 Add the ability to copy resultantdq from SDF to a rampmodel

### LINKED ISSUE #814: Update dqinit to look for an incoming dq array in the level 1 files
_Issue [RCAL-622](https://jira.stsci.edu/browse/RCAL-622) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

When the science raw schema is updated (rad-132) to include possible data drops in constructing the level 1 file the dq array is needed and will be added as an optional product. It may, in rare cases, be needed and will contain location of missing data, "DROPOUT", values.

Update the dq init step to detect this and use any values that come from SDF.

ddavis-stsci:
Resolves [RCAL-622](https://jira.stsci.edu/browse/RCAL-622)

Closes #814 

This PR adds the ability to copy resultantdq from a SDF science raw model to the new rampmodel created by dq_init.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/570/
 The failures are all from requiring wcs and renaming the roman_datamodel which is being addressed in the coming rad and datamodels release.


schlafly (approved):
This looks good to me.  It looks to me separate from this PR, but my recollection is that our guidance to SDF is that they should not include a resultantdq array when there are no flagged pixels, which we expect to be the case for >99% of real images.  Is that also what you're thinking?

ddavis-stsci:
That is correct, the resultantdq should be present only if there are data drops in the telemetry.

PR #1086, opened 2024-01-31
Update tweakwcs dependency

schlafly:
tweakwcs 0.8.6 includes an update to the image alignment step that dramatically reduces memory usage and improves speed for aligning 18 detector Roman arrays to Gaia.  This updates the dependency of romancal to require that version.

Passing regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/587/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     'stcal >=1.5.1',
     #'stcal @ git+https://github.com/spacetelescope/stcal.git@main',
     'stpipe >=0.5.0',
-    'tweakwcs >=0.8.0',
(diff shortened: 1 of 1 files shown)


schlafly:
@ddavis-stsci , this updates us to require the tweakwcs version that does the faster / lower memory source matching.  As you pointed out on the call, this is a little optional, in that we'll by default pull in the latest version, and the earlier versions work and are just a bit explosive if you try to run them with the full array.  But it made sense to me to bump the minimum version.

ddavis-stsci (approved):
LGTM

PR #1088, opened 2024-02-02
Use dark slope in ramp fitting

schlafly:
Presently the ramp fitting step operates on dark-subtracted ramps, but assumes that the Poisson noise in those ramps can be well estimated by the square root of the rate.  The dark current contribution to the rate should be part of that Poisson noise, but is omitted because the step operates on dark-subtracted ramps.

This PR adds the dark current portion of the dark reference file (dark_slope) back into the ramp, performs the ramp fitting, and then removes the contribution from the slopes.  This is related to
https://innerspace.stsci.edu/display/RI/Ramp+Fitting+-+v1
.  It has the slight benefit of being more correct (we're choosing weights better), but being more convoluted (we're adding and subtracting dark current.

We'll also use more memory than previously---the darks are rather big, and now we need to have them in memory during ramp fitting, which is when we have peak memory usage.

This changes the variances slightly and is expected to break regression tests.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,6 +70,8 @@ ramp_fitting
 
 - Add default WCS when constructing image model from ramp model [#1072]
 
+- Account for Poisson noise from dark current when fitting ramps.
+
 resample
 --------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,7 +70,7 @@ ramp_fitting
 
 - Add default WCS when constructing image model from ramp model [#1072]
 
-- Account for Poisson noise from dark current when fitting ramps.
+- Account for Poisson noise from dark current when fitting ramps. [#1087]
 
 resample
 --------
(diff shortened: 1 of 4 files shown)


schlafly:
@tddesjardins , do you want to review this use of dark_slope during ramp fitting?  This isn't quite what was requested in https://innerspace.stsci.edu/display/RI/Ramp+Fitting+-+v1 (which would require a API change to stcal), but I think it's in some way more correct.

schlafly:
@t-brandt, this is the PR with the dark_slope implementation we talked about; I'd love to get this "blessed" by RTB so that I can merge.

tddesjardins:
LGTM; RTB approves!

ddavis-stsci (approved):
The code look fine to me.

PR #1092, opened 2024-02-07
RCAL-726: allow ModelContainer as input to outlier detection step.

mairanteodoro:
Resolves [RCAL-726](https://jira.stsci.edu/browse/RCAL-726)

This PR changes the outlier detection step to allow a `ModelContainer` to be passed in as input;

**Regression tests**
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/651/
N.B.: The failures reported for the forked branch are the same as the ones reported for the main branch.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


ddavis-stsci:
Well it run but I'm not quite sure what it is doing. It seems to be saving the results in files? e.g. 2024-02-07 12:44:01,658 - ```
stpipe.HighLevelPipeline.outlier_detection - INFO - Exposure r0000101001001001001_01101_0002_WFI01_outlier_i2d.asdf saved to file
2024-02-07 12:44:02,471 - stpipe.HighLevelPipeline.outlier_detection - INFO - Computing median
2024-02-07 12:44:38,759 - stpipe.HighLevelPipeline.outlier_detection - INFO - Saved model in r0000101001001001001_01101_0002_WFI01_median.asdf
```
and when running the pipeline we should only get the final output file, which at the moment is the poorly named test_sky_asn_highlevelpipeline.asdf
It is not clear to me what is being passed from outlier_detection to resample.

**Comment by mairanteodoro** [comment]
> [...] It is not clear to me what is being passed from outlier_detection to resample.

TL;DR: a `ModelContainer` object is being passed from outlier_detection to resample.

Long answer:
On line 105 of `outlier_detection_step.py`:
```python
step = detection_step(self.input_models, **pars)  # <=== we instantiate the class with a ModelContainer
```

Then we run the `do_detection` method that will call resample:
```
resamp = resample.ResampleData(
    self.input_models, single=True, blendheaders=False, **pars  # <=== still a ModelContainer
)
drizzled_models = resamp.do_drizzle()
```

ddavis-stsci:
Why do you need to instantiate a model? The pipeline should be passing 
you a model container.

ddavis-stsci:
Oops,  never mind I misinterpreted what you were doing.

nden (changes requested):
In addition to the comments, running outlier_detection on the command line with an association file fails with an error in resample. Is this supposed to work?

romancal/pipeline/highlevel_pipeline.py
```diff
@@ -57,11 +57,11 @@ def process(self, input):
...
-            # result = self.outlier_detection(input)
-            self.skymatch.suffix = "i2d"
-            result = self.resample(input)
+            self.outlier_detection.suffix = "outlier_detection"
+            result = self.outlier_detection(result)
+            self.resample.suffix = "i2d"
+            result = self.resample(result)
             if input_filename:
-                result.meta.filename = input_filename
+                result.meta.basic.filename = input_filename
```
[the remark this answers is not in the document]
mairanteodoro: This change was reverted as it should not be included in this PR.

mairanteodoro:
> In addition to the comments, running outlier_detection on the command line with an association file fails with an error in resample. Is this supposed to work?

Could you please provide the command that you used?

mairanteodoro:
> In addition to the comments, running outlier_detection on the command line with an association file fails with an error in resample. Is this supposed to work?

Current code was successfully executed with the following:

```python
from romancal.stpipe import RomanStep

output_foldername = "custom/path/where/files/will/be/saved/to"

input_filename = "test_asn.json"

args = [
    "--disable-crds-steppars",
    "--save_results=True",
    f"--output_dir={output_foldername}",
    # "--suffix=elp",
    "romancal.step.OutlierDetectionStep",
    str(input_filename),
]

RomanStep.from_cmdline(args)
```

ddavis-stsci (approved):
It now takes a ModelContainer as input and outputs a ModelContainer as expected so I think this is ready to merge. 

```
type(mc1)
<class 'romancal.datamodels.container.ModelContainer'>
result_outl = OutlierDetectionStep.call(mc1)
...
2024-03-12 15:10:47,635 - stpipe.OutlierDetectionStep - INFO - New pixels flagged as outliers: 657432 (3.93%)
2024-03-12 15:10:47,994 - stpipe.OutlierDetectionStep - INFO - Step OutlierDetectionStep done
type(result_outl)
<class 'romancal.datamodels.container.ModelContainer'>
```

schlafly:
@mairanteodoro , would you mind resolving the merge conflict here.  There's something slightly non-trivial in outlier_detection.py.

romancal/outlier_detection/outlier_detection.py
```diff
@@ -84,10 +83,13 @@ def do_detection(self):
         if pars["resample_data"]:
             # Start by creating resampled/mosaic images for
             # each group of exposures
+            # in_memory = pars["in_memory"]
+            # pars["in_memory"] = True
             resamp = resample.ResampleData(
                 self.input_models, single=True, blendheaders=False, **pars
             )
             drizzled_models = resamp.do_drizzle()
+            # pars["in_memory"] = in_memory
```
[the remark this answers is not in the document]
mairanteodoro: Thanks for catching that!

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -52,16 +52,18 @@ class OutlierDetectionStep(RomanStep):
...
-        self.input_models = input_models
-        self.input_container = isinstance(self.input_models, ModelContainer)
         self.skip = False
 
+        try:
+            self.input_models = ModelContainer(input_models)
+        except TypeError:
+            self.log.warning("Input cannot be parsed into a ModelContainer.")
+            self.skip = True
+            return
```
nden, line 62: This needs to return the input unchanged as the expectation is that the step returns the input if it cannot run so that the next step can still run in the pipeline.

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,86 +1,130 @@
...
 )
 def test_outlier_raises_error_on_invalid_input_models(input_models):
-    """Test that OutlierDetection raises an error if input is not a ModelContainer."""
+    """Test that OutlierDetection raises an error if input cannot be parsed into a ModelContainer."""
 
     step = OutlierDetectionStep()
 
-    step.process(input_models)
+    with pytest.raises(Exception) as exec_info:
+        step(input_models)
```
nden, line 53: The code issues a warning and the test looks for an AttributeError (while actually a TypeError is raised).
This is a problem with the code and the test. 
Parsing invalid input into a `ModelContainer` raises a TypeError which is caught and a warning issued. So this test should be looking for and catching a warning.

The AttributeError which you see when you run the current code comes from `finalize_results`, not from parsing the input into a `ModelContainer`  and is raised because the input is not returned in that case (see my previous comment).

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,86 +1,130 @@
...
+    Test if the outlier detection step is skipped when the exposure type is different from WFI image.
+    """
     img_1 = base_image()
     img_1.meta.exposure.type = "WFI_PRISM"
     img_2 = base_image()
     img_2.meta.exposure.type = "WFI_PRISM"
 
     step = OutlierDetectionStep()
-    step.process(ModelContainer([img_1, img_2]))
+    step(ModelContainer([img_1, img_2]))
```
nden, line 81:
```suggestion
    step.call(ModelContainer([img_1, img_2]))
```

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -1,86 +1,130 @@
...
-    asn_filepath = create_mock_asn_file(tmp_path)
+    mc = ModelContainer([img_1, img_2])
 
     step = OutlierDetectionStep()
-    # set output dir for all files created by the step
-    step.output_dir = tmp_path
     # make sure resample does not save file to disk
     step.in_memory = True
-    step.process(ModelContainer(asn_filepath))
+    step(mc)
```
nden, line 127:
```suggestion
    step.call(mc)
```

nden (changes requested):
Looks mostly good. Some inline comments

romancal/outlier_detection/outlier_detection_step.py
mairanteodoro: If `input_models` can't be parsed into a `ModelContainer`, that means it is invalid, in which case `finalize_results` will error out when trying to set metadata to invalid input when running an individual step:

https://github.com/spacetelescope/romancal/blob/879af02b536c354001df13198ce5a49af4637f52/romancal/stpipe/core.py#L66-L70

nden: I think there's contradiction in your comment because `None` is also not a parsable input for ModelContainer and currently the code returns `None`. 
All I am asking is to make the different return statements consistent. If you do so the test won't be looking for a generic Exception to be raised (which can be anything) but for catching the Warning that is issued when the TypeError is raised.

I will dismiss my review. Feel free to merge after considering the suggestions for the wording of the warnings.

mairanteodoro: `OutlierDetectionStep` now returns the same input if it is not parsable into a `ModelContainer`. Thanks, @nden!

PR #1093, opened 2024-02-07
Updates to change pass to pass_number

WilliamJamieson:
This PR makes the minor fixes needed to support the changes from spacetelescope/rad#360 and spacetelescope/roman_datamodels#306. These are to avoid a name collision with the `pass` syntax word in python.

Note that the regression test files will have to be updated because they use the word `pass` instead of `pass_number` and so will not validate when being read.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


PaulHuwe:
This change will not be made. Closing.

PR #1098, opened 2024-02-09
Make copies of border pixel arrays when building L2 image.

schlafly:
romancal produces larger L2 files than one would expect given the sizes of the arrays in rad; typical L2 file sizes are ~1.1 GB but one would estimate ~400 MB from the different planes.  The issue turns out to be that the border pixels are views into the larger full L1 image rather than copies just of the side 4 pixels.  asdf is saving the full array, rather than just the small subset representing the view, leading to the large files.

Changing these to copies addresses the issue.  After this change, file sizes reduce to ~400 MB.

**Checklist**
- [X] ran regression tests, https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/599/


schlafly:
@ddavis-stsci , we could conceivably include this.  This makes the L2 files we make 3x smaller but doesn't affect any regression tests in my run.  It addresses @wlim-ipac 's issue from the call yesterday.

ddavis-stsci (approved):
Looks clean to me.

WilliamJamieson (approved):

PR #1108, opened 2024-02-21
Find and fix all the tmpdir usages

WilliamJamieson:
[`pytest` recommends creating temporary directories with the `tmp_path` fixture instead of the legacy `tmpdir`](https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures). To check for use of the legacy method one can run with
```
pytest -p no:legacypath
```
This PR used that command to discover all uses of `tmpdir` and correct them. It does not add this as a permanent check because the `pytest_asdf` plugin that is built into the `asdf` package relies on this legacy system. This was fixed by asdf_format/asdf#1759, but has not been released yet.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -1,4 +1,22 @@
+import os
+from contextlib import contextmanager
+
 import pytest
+
+# chdir contextmanager was added with Python 3.11
+try:
+    from contextlib import chdir
+except ImportError:
+
+    @contextmanager
+    def chdir(path):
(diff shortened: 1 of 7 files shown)


schlafly:
LGTM.

nden (approved):
Thank @WilliamJamieson

Issue #1115, opened 2024-02-27
Use tmp_path instead of tmpdir pytest fixture


jdavies-st:
Replace all instances `tmpdir` and `tmpdir_factory` pytest fixtures with `tmp_path` and `tmp_path_factory` fixtures, respectively.

Then in pyproject.toml:

```toml
[tool.pytest.ini_options]
addopts = [
    "-p no:legacypath",
]
```

which will prevent it from being used in the future.  This is will make the code more robust for use of `pathlib.Path` instances in the runtime code.

nden:
Thanks! Fixed by #1108

braingram:
I'm going to reopen this to track the addition of `no:legacypath` to `pyproject.toml`. This is currently held up by the `pytest-asdf` plugin (which has been fixed in main but not yet released).

braingram:
This is also blocked by ddtrace (which uses the legacy path as part of it's plugin):
https://github.com/spacetelescope/romancal/actions/runs/8071483455/job/22051208643?pr=1117#step:10:215

braingram:
This is also blocked by `ci_watson` using `tmpdir`
https://github.com/spacetelescope/ci_watson/blob/2fc02e8312310753c88ff0a3c4bf16cb545205e8/ci_watson/plugin.py#L70
and by uses of `tmp_factory` in the romancal regtests (see this run that disabled ddtrace for the regtests and shows 26 errors):
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/613/pipeline/220

braingram:
A PR is open to update the `tmpdir` usage in `ci_watson`: https://github.com/spacetelescope/ci_watson/pull/64

braingram:
https://github.com/spacetelescope/romancal/pull/1117
enabled this option for unit tests.

enabling it for regression tests is blocked by:
https://github.com/DataDog/dd-trace-py/issues/8533

I vote for leaving this open to track the issue, once ddtrace is updated we can enable the regression tests and close the issue.

emolter:
I happened to run across [this line](https://github.com/spacetelescope/romancal/blob/a7c3a5d430dfae954482ee80ac0c7089c401548e/romancal/tests/base_classes.py#L32) that seems to still point to _jail; looks to me like this should be changed before merging.

braingram:
[...]

PR #1117, opened 2024-02-27
remove `tmpdir_factory` usage, enable no:legacypath as a tox factor

braingram:
Incompatibility between ddtrace (and ci_watson) and the `-p no:legacypath` option for pytest means that the `-p no:legacypath` option cannot be "globally" enabled.

This PR instead enables the option as a tox factor and uses the factor for a single pytest suite run `py310-webbpsf-nolegacypath`. This will allow romancal to keep the current asdf lower pin, not wait until ddtrace and ci_watson are updated and still reap most of the benefits of the `-p no:legacypath` option as all tests run during the `py310-webbpsf-nolegacypath` run will be checked for legacy path usage.

Regtests were run with no errors `-p no:legacypath` (and ddtrace disabled):
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/616/pipeline/220
The run used c843a2a88c972fc0253b02d7b0942a007a044142 which has since been removed from this PR to allow the regtests to continue to use ddtrace.

Once ddtrace is updated, the regtests should also be updated to use `-p no:legacypath`.

As the changes in this PR will not enforce `-p no:legacypath` for the regtests https://github.com/spacetelescope/romancal/issues/1115 will be updated once this PR is merged.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/conftest.py
+++ b/romancal/conftest.py
@@ -43,8 +43,19 @@ def slow(request):
     return request.config.getoption("--slow")
 
 
+@pytest.fixture(scope="function")
+def function_jail(tmp_path):
+    """Perform test in a pristine temporary working directory."""
+    old_dir = os.getcwd()
+    os.chdir(tmp_path)
+    try:
+        yield str(tmp_path)
+    finally:
+        os.chdir(old_dir)
+
+
 @pytest.fixture(scope="module")
-def jail(request, tmpdir_factory):
+def module_jail(request, tmp_path_factory):
     """Run test in a pristine temporary working directory, scoped to module.
 
     This fixture is the same as _jail in ci_watson, but scoped to module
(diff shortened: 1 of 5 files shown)


romancal/regtest/conftest.py
```diff
@@ -191,12 +191,12 @@ def _rtdata_fixture_implementation(artifactory_repos, envopt, request):
 
 
 @pytest.fixture(scope="function")
-def rtdata(artifactory_repos, envopt, request, _jail):
+def rtdata(artifactory_repos, envopt, request, function_jail):
```
braingram, line 194: This change replaces the use of the `_jail` fixture from `ci_watson` (which uses the legacy path). Given the leading underscore in this fixture it doesn't appear to be intended for public use. The replacement `function_jail` is named to help distinguish it from `jail` (which is module scoped and this PR renamed to `module_jail`).

romancal/conftest.py
```diff
@@ -41,8 +41,19 @@ def slow(request):
...
+    old_dir = os.getcwd()
+    os.chdir(tmp_path)
+    try:
+        yield str(tmp_path)
+    finally:
+        os.chdir(old_dir)
+
+
 @pytest.fixture(scope="module")
-def jail(request, tmpdir_factory):
```
braingram, line 45: This `tmpdir_factory` usage was missed in the earlier PR likely because it is only used in regtests.

schlafly (approved):
Looks good to me, thanks!

PR #1120, opened 2024-02-29
RCAL-751 Create FluxStep to apply the flux correction to Level 2 data.

stscieisenhamer:
Resolves [RCAL-751](https://jira.stsci.edu/browse/RCAL-751)

This PR implements the flux correction to convert from native Level 2 electrons/s units to MJy/sr.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)


stscieisenhamer:
Of the functional items still needed, the main one is ensuring file naming is correct if step results are saved. There is also documentation and a regression test needed. However, the main code is ready for initial review.

The main design point that needs to be considered is that, currently, there is no optimization concerning memory management. I believe there are still larger discussions on the best solutions for this, so I have left the code dead-simple: All is presumed in-memory and the input model(s) are modified in-place.

romancal/flux/flux_step.py
```diff
@@ -0,0 +1,114 @@
...
+        return
+
+    # Apply the correction
+    # Assignments into the model are done through `_instance` to avoid
+    # validation errors on the units.
+    log.debug("Flux correction being applied")
+    c_mj = model.meta.photometry.conversion_megajanskys
+    model._instance["data"] = model.data * c_mj
+    for variance in VARIANCES:
+        model._instance[variance] = getattr(model, variance) * c_mj**2
```
schlafly, line 114: The 'err' column also needs to get multiplied by c_mj.

The model._instance[field] is to avoid validation, which we expect will not be needed if we loosen the set of allowed units in the wfi-image schema?  It will be challenging for us to save these at present?

romancal/flux/tests/test_flux_step.py
```diff
@@ -0,0 +1,117 @@
...
+from romancal.flux import FluxStep
+
+
+@pytest.mark.parametrize(
+    "attr, factor",
+    [
+        ("data", 1),
+        ("var_rnoise", 2),
+        ("var_poisson", 2),
+        ("var_flat", 2),
```
schlafly, line 18: Add: ('err', 1)

schlafly (commented):
This looks good.  Have you verified that if you plug it into the HLP and feed it an association that it does the right thing, calibrating the images and then later steps seeing those calibrated images?

The step probably needs to be registered here:
https://github.com/spacetelescope/romancal/blob/78532b92555a8701f82744ecf0d3b1711f7e7a83/romancal/step.py#L16
and get some short docs.

I'm still hung up by the fact that I think the one time I used the HLP the model container contained a list of file names rather than a list of in-memory models, and so used this branch
https://github.com/spacetelescope/romancal/blob/main/romancal/datamodels/container.py#L211
where it's not clear to me that anything would get saved.  Maybe not part of this PR, but it would be good if you could figure out how to call the pipeline in a way such that this step would work.

I imagine that the sky subtraction & outlier detection steps are in a similar situation, but I am not sure.

romancal/flux/flux_step.py
schlafly: Oh, and I haven't paid attention to whether conversion_megajanskys carries units or not; if it's just a scalar, we want to update the units.

stscieisenhamer: The `_instance` use here was really because of a bug in roman_datamodels that has been resolved by PR#322. After pulling roman_datamodels:main, the use of `_instance` is removed.

`conversion_magajanskys` does carry units. Oddly errors are not raised at this point, though I suspect there will be when models are serialized.  Regardless, the aforementioned widening of allowed units will resolve.

stscieisenhamer:
HLP has some implementation issues. In its current form, it can only operate if all is in-memory and the steps work on the data in-place, and that is just to get around the syntax issues; see how `result` and `input` relate from step-to-step.

Workin' it...

stscieisenhamer:
Status: Main work is complete, I believe.  Two outstanding issues to be discussed:

Unit change of Level 2: The presumption here is still `e/s`. Should I update to `DN/s` or wait until the related `rad` changes are merged first?

HLP: Between RCAL-777 and other structure/memory-related discussions, I have done minimal updates and xfailed the related test. I would suggest disconnecting any further HLP work from this PR and leave it for other issues + PR's.

romancal/flux/flux_step.py
```diff
@@ -0,0 +1,114 @@
...
+        log.debug("Input data units are %s", model.data.unit)
+        return
+
+    # Apply the correction
+    # Assignments into the model are done through `_instance` to avoid
+    # validation errors on the units.
+    log.debug("Flux correction being applied")
+    c_mj = model.meta.photometry.conversion_megajanskys
+    for data in DATA:
+        model[data] = model[data] * c_mj
```
schlafly, line 112: And, sorry, I'm confused all the sudden.  It looks to me like c_mj has units of MJy/sr, and model[data] should have units of something like e/s, so are the units of the new model_data e/s * MJy/sr?  Where are the old units being stripped?

docs/roman/flux/main.rst
```diff
@@ -0,0 +1,40 @@
...
+
+Flux Application
+----------------
+If the input data is in :math:`electrons / second`, the flux scale factor, as
+found in the meta ``meta.photometry.conversion_megajanskys``, is simply
+multiplied to the ``data`` and ``err`` arrays. The square of the scale factor is
+multiplied to all the variance arrays. The resultant units is in :math:`MJy/sr`.
+
+.. math::
+   DATA =* SCALE
```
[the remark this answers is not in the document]
stscieisenhamer: WTF on the docs...will fix that but not sure where that came from.

About syntax: Yea, I was going down with the headcold thing...will fix.

romancal/flux/flux_step.py
schlafly: The L3 images need to be in MJy / sr.

conversion_megajansky "should" be in (MJy / sr) / (e / s), so that the (e / s) cancels out in that product.  But it looks like we've chosen to define it as the number of MJy/sr producing 1 count per second, eliding the denominator.  I think the easiest thing would be to define c_mj = conversion_megajansky / (1.0 * e / s).  That will need to change to c_mj = conversion_megajansky / (1.0 * DN / s) when the L2 unit changes go in, but that doesn't sound bad.

We will also need to work with RTB to make sure that the photometry tables are consistent with the unit change.  In principle I think we need new files with all of the zero points modified by a factor of the average gain, though in practice I don't think we're presently looking very hard at the overall zeropoints and so if we didn't do this we wouldn't notice.  @tddesjardins

stscieisenhamer:
I have left units at e/s, just for test/validation to pass. However, if merged before the unit change is in, all that needs to [change is this line](https://github.com/spacetelescope/romancal/blob/896845d15666a84e551bc56042514b32c0ef1c86/romancal/flux/flux_step.py#L18).

schlafly:
This looks good to me.  Dave indicates in the rcal-777 discussion that you should be able to get the hlp test to pass with an invocation along the lines of
`strun --disable-crds-steppar roman_hlp mosaic_asn.json --verbose`
does that not work for you?

schlafly (approved):
LGTM, thanks!

stscieisenhamer:
Errors are due to the unit change for Level 2 from e/s to DN/s. This should resolve with the recent rad/rdm merges.

PR #1125, opened 2024-03-04
Improve PSF fitting

schlafly:
This PR makes a few minor changes to the PSF fitting to improve the astrometric agreement.  These are:
- When sqrt_n_psfs = 1, choose the central PSF to be in the middle of the detector.  Previously we chose one of the corners in this case.
- Ignore the NO_LIN_CORR flag.  Currently ~half of all pixels in the CRDS reference files are flagged with NO_LIN_CORR and one can get bad results if one masks everything with this flag.  When these files are updated (post TVAC?), I expect we can drop this flag.
- Fix the ~sign of the local background correction.  We had been accidentally turning off the local background correction by keeping it when None was passed in and setting it to a default when a specific choice was sent in; oops.
- Add a minimum source separation to avoid detecting too much junk in the wings of bright stars.
- Make the peak brightness larger to avoid throwing away the bright Gaia stars.

With these changes we can do pretty dark good fits to the stars in the simulator, in all filters.

image

Before this PR, agreement was ~0.02 pix, now it is ~0.005 pix.  We need ~1 mas to pass SOC-493.  This is significantly better than that, though these plots are with an artificial grid of sources rather than real Gaia sources, and only use one SCA rather than the full array.

Regression tests are here
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/625/
but "fail" because the catalog of sources is much smaller.  This is due to the new minimum source separation which prevents from finding large numbers of sources in the wings of bright stars.  We will need to update the regtest files to resolve this issue.

**Checklist**
- [x] added relevant label(s)
- [X] ran regression tests, https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/625/


bmorris3 (approved):
This looks great. Thanks!

PR #1131, opened 2024-03-08
Rad-155

nden:
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR makes the changes needed for RAD-155. It depends on

https://github.com/spacetelescope/rad/pull/390
https://github.com/spacetelescope/roman_datamodels/pull/328

Regression tests are run here
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/635/
The one failure is expected and needs an updated truth file.

**Checklist**
- [x] added entry in `CHANGES.rst` under the corresponding subsection
- [x] updated relevant tests
- [ ] updated relevant documentation
- [x] updated relevant milestone(s)
- [x] added relevant label(s)
- [x] ran regression tests, post a link to the Jenkins job below. [How to run regression tests on a PR](https://github.com/spacetelescope/romancal/wiki/Running-Regression-Tests-Against-PR-Branches)

PR #1139, opened 2024-03-12
Fix some dangling references to basic.filename and basic.telescope.

schlafly:
The change in the L3 files to move some keywords from meta.basic.X to meta.X left some bits of the L3 pipeline in a broken state.  This is intended to resolve those.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


nden:
This, and others,  I believe are fixed in #1131

PR #1144, opened 2024-03-14
[SCSB-145] require Python 3.10

zacharyburnett:
resolves [SCSB-145](https://jira.stsci.edu/browse/SCSB-145)

propagate Astropy's deprecation of Python 3.9 to downstream packages

> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

PR #1145, opened 2024-03-15
remove unused test code

braingram:
This PR removes the unused sdp-related test fixtures. Including
- `--sdp-pool` and `--standard-pool` options
- `sdp-pool` and `standard-pool` fixtures
- `SDPPoolsSource` and `BaseRomanTest` test classes
- and the commented out `raw_from_asn`

Regression tests show only unrelated error:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/652/tests

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
nden I requested your review as these files appear to be copies of files in jwst. I think they might be in use there in [test_against_standard](https://github.com/spacetelescope/jwst/blob/77c4966a836e03f34939d3de6ba805bc0934a7d6/jwst/regtest/test_associations_standards.py#L112). That may mean that the files removed in this PR might eventually be used by romancal (or that their lack of use is a bug). Thanks and let me know if more details are helpful.

nden (approved):
These were copied from JWST anticipating we'll have to use them for testing generating associations but it looks like we'll have other types of associations and may not need them. I'm fine removing them. @ddavis-stsci Do you agree?

ddavis-stsci:
Removing them is probably fine. It doesn't look like we'll have 
realistic pool files for a while.
Also if I recall they depend on the ci_watson plugin. I'm not sure how 
that will transition to
GA.

If we need them later we'll just put them back.

braingram:
Thanks! The py39 failures are due to `roman_datamodels` dropping python 3.9 and are addressed in: https://github.com/spacetelescope/romancal/pull/1144

PR #1146, opened 2024-03-18
RCAL-725: fix outlier detection step issues.

mairanteodoro:
Resolves [RCAL-725](https://jira.stsci.edu/browse/RCAL-725)

This PR fixes a bug with the outlier detection step in which the algorithm was still detecting outliers even if the input models were identical. Additionally, the median image was not being properly created, which was being caused by wrongly setting the `single` parameter in the call to resample.

A unit test for identical input datamodels was added.

**Regression tests**
No errors/failures were reported:
- https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/655/

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/outlier_detection/outlier_detection.py
+++ b/romancal/outlier_detection/outlier_detection.py
@@ -84,7 +84,7 @@ class OutlierDetection:
             # Start by creating resampled/mosaic images for
             # each group of exposures
             resamp = resample.ResampleData(
-                self.input_models, single=True, blendheaders=False, **pars
+                self.input_models, single=False, blendheaders=False, **pars
             )
             drizzled_models = resamp.do_drizzle()
 
--- a/romancal/outlier_detection/outlier_detection.py
+++ b/romancal/outlier_detection/outlier_detection.py
@@ -90,16 +90,21 @@ class OutlierDetection:
 
         else:
             # for non-dithered data, the resampled image is just the original image
-            drizzled_models = self.input_models
-            for i in range(len(self.input_models)):
-                drizzled_models[i].weight = build_driz_weight(
+            drizzled_models = ModelContainer()
+            for i, model in enumerate(self.input_models):
+                model["weight"] = build_driz_weight(
                     self.input_models[i],
                     weight_type="ivm",
                     good_bits=pars["good_bits"],
                 )
+                drizzled_models.append(model)
 
         # Initialize intermediate products used in the outlier detection
-        median_model = rdm.open(drizzled_models[0]).copy()
+        median_model = (
+            rdm.open(drizzled_models[0]).copy()
+            if isinstance(drizzled_models[0], str)
+            else drizzled_models[0].copy()
+        )
 
         # Perform median combination on set of drizzled mosaics
         median_model.data = Quantity(
@@ -230,6 +235,8 @@ class OutlierDetection:
 
         log.info("Blotting median")
         for model in self.input_models:
+            # TODO: fix dtype conflict
+            model.dq = model.dq.astype("uint32")
             blotted_median = model.copy()
 
             # clean out extra data not related to blot result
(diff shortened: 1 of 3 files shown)


romancal/outlier_detection/outlier_detection.py
```diff
@@ -84,22 +84,27 @@ def do_detection(self):
...
-            for i in range(len(self.input_models)):
-                drizzled_models[i].weight = build_driz_weight(
+            drizzled_models = ModelContainer()
+            for i, model in enumerate(self.input_models):
+                model["weight"] = build_driz_weight(
                     self.input_models[i],
                     weight_type="ivm",
                     good_bits=pars["good_bits"],
                 )
+                drizzled_models.append(model)
```
schlafly, line 100: Just because I'm not seeing it... can you elaborate on the choice of directly iterating on self.input_models vs. making a new model container for them?

romancal/outlier_detection/outlier_detection.py
```diff
@@ -230,6 +235,8 @@ def blot_median(self, median_model):
 
         log.info("Blotting median")
         for model in self.input_models:
+            # TODO: fix dtype conflict
+            model.dq = model.dq.astype("uint32")
```
schlafly, line 239: Do you intend to have a # TODO here?

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -293,3 +293,67 @@ def test_outlier_do_detection_find_outliers(tmp_path, base_image, clean_up_after
     assert all(o == i for i, o in zip(outliers_input_coords, outliers_output_coords))
 
     clean_up_after_test("*.asdf")
+
+
+def test_outlier_do_detection_do_not_find_outliers_in_identical_images(
```
schlafly, line 298: Just trying to understand, this looks like a reasonable but conceptually unrelated test to the main single = True bug fix above.  Is that the case?

mairanteodoro: Yes, that's correct. In this case, we're passing copies of the same image to the method and asserting that no pixel is flagged as outlier after processing.

romancal/outlier_detection/outlier_detection.py
mairanteodoro: It seems that the romanisim files that I used do not suffer from the `int32`/`uint32` issue, so I removed the `TODO` label. Thanks for the catch!

romancal/outlier_detection/outlier_detection.py
mairanteodoro: Thanks, Eddie!
When I used the `ModelContainer`, I intended to test something else not related directly with the ticket; it should not be there. What I really wanted to do is to clean up the `for` block to look like this:
``` python
drizzled_models = self.input_models
for model in drizzled_models:
    model.weight = build_driz_weight(
        model,
        weight_type="ivm",
        good_bits=pars["good_bits"],
     )
```

braingram:
@mairanteodoro I have a question about a change in this PR.

Passing `single=False` results in `ResampleData` using `many_to_one` (instead of `many_to_many`) when it is called prior to `create_median` during `outlier_detection`. During the `hlp` regtest (where 3 files are input, each with a different "group id") this results in a single image (the resampled result of all 3 combined groups) being passed to `create_median`. This might be true for all calls to outlier detection so that it's always taking the median of a single combined image. This is different than what is done for jwst (which uses `single=True`) where each group produces a separate image (so for 3 input groups, 3 drizzled images would be produced and the median taken over these 3 drizzled images).

Is this the intended result for these changes and if so, the median could be completely removed as it's only a median over a single image.

nden:
It's unclear to me which build this went into. @mairanteodoro please add the milestone to all PRs in the future.

PR #1147, opened 2024-03-18
RCAL-620: allow ModelContainer to work with context manager.

mairanteodoro:
Resolves [RCAL-620](https://jira.stsci.edu/browse/RCAL-620)

This PR implements two magic methods (`.__enter__()` and `.__exit__()`) that allows `ModelContainer` to work properly with a context manager.

Two unit tests have been added to test the new implementation.

**Regression test**
No errors/failures reported:
~https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/661/~
~https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/683/~
~https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/690/~
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/691/

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/datamodels/container.py
+++ b/romancal/datamodels/container.py
@@ -93,6 +93,7 @@ class ModelContainer(Sequence):
         asn_file = "/path/to/asn_file.json"
         container = ModelContainer(asn_file)
 
+
     In any of the cases above, the content of each file in a `ModelContainer` can
     be accessed by iterating over its elements. For example, to print out the filename
     of each file, we can run:
@@ -101,6 +102,14 @@ class ModelContainer(Sequence):
 
         for model in container:
             print(model.meta.filename)
+
+
+    Additionally, `ModelContainer` can be used with context manager:
+
+    .. code-block:: python
+
+        with ModelContainer(asn_file) as asn:
+            # do stuff
 
 
     Notes
@@ -225,6 +234,14 @@ class ModelContainer(Sequence):
             if not isinstance(model, rdm.DataModel) and self._return_open:
                 model = rdm.open(model, memmap=self._memmap)
             yield model
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is not None:
+            print(f"\nAn exception occurred in ModelContainer: {exc_val}")
+            print(f"\nTraceback:\n{exc_tb}")
 
     def insert(self, index, model):
         if isinstance(model, rdm.DataModel):
--- a/romancal/datamodels/container.py
+++ b/romancal/datamodels/container.py
@@ -239,9 +239,8 @@ class ModelContainer(Sequence):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if exc_type is not None:
-            print(f"\nAn exception occurred in ModelContainer: {exc_val}")
-            print(f"\nTraceback:\n{exc_tb}")
+        # exceptions will be propagated out of the context
+        return False
 
     def insert(self, index, model):
         if isinstance(model, rdm.DataModel):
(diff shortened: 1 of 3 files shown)


romancal/datamodels/tests/test_datamodels.py
```diff
@@ -504,3 +510,81 @@ def test_make_file_with_index(filename, idx, expected_filename_result, tmp_path)
...
+        result_filenames.extend(
+            tmp_path / datamodel.meta.filename for datamodel in asn1
+        )
+
+    expected_filenames = [filename1, filename2]
+    expected_filenames.sort()
+    result_filenames.sort()
+
+    assert all(e == r for e, r in zip(expected_filenames, result_filenames))
+    assert type(asn1) is ModelContainer
```
ddavis-stsci, line 547: Can the above code basically just be
with ModelContainer(asn_filepath, return_open=False, save_open=False ) as asn1:
    assert type(asn1) is ModelContainer

What does the rest of the code above do wrt the context managment?

romancal/datamodels/tests/test_datamodels.py
```diff
@@ -10,6 +10,12 @@
 from romancal.datamodels.container import ModelContainer, make_file_with_index
 
 
+@pytest.fixture
+def mock_print(mocker):
+    # Mocking the print function to capture outputs for assertions
+    return mocker.patch("builtins.print")
+
```
ddavis-stsci, line 17: This seems to be the only place where mocker is used. Why do these unit tests need mocker?

romancal/datamodels/container.py
```diff
@@ -226,6 +235,14 @@ def __iter__(self):
                 model = rdm.open(model, memmap=self._memmap)
             yield model
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is not None:
+            print(f"\nAn exception occurred in ModelContainer: {exc_val}")
+            print(f"\nTraceback:\n{exc_tb}")
```
schlafly, line 244: I haven't thought enough about exception handling and context managers.  Do you expect the traceback to not get printed without these three lines?

mairanteodoro: By default (i.e. in the case where we don't add any logic to the method), if an exception occurs inside the context manager, the `.__exit__()` method is called and the traceback will be printed. The implemented code was just a placeholder for handling exceptions; I refactored it to simply `return False` so that, if needed, exceptions can be handled outside the `.__exit__()` method for the moment.

romancal/datamodels/tests/test_datamodels.py
mairanteodoro: We needed mocker for mocking the print statement in the newly added unit test for the `.__exit__()` method. However, since the `.__exit__()` method has been refactored, `mocker` is not needed anymore and has been removed.

schlafly (commented):
This looks reasonable to me.  Thinking out loud for a second, usually when one has a context manager one uses it to clean up open files automatically when the with block ends.  We're not doing that here.  I think that would look like
```
for model in self._models:
    if isinstance(model, rdm.DataModel):
        model.close()
```
but it's unclear to me if that's desired in this context.  @ddavis-stsci , should we be cleaning up the models in this case or is it fine to leave them open and let the garbage collector clean them up later?

ddavis-stsci:
It is usually good practice to have the exit method do some of the 
cleanup. We could rely on the
garbage collector but that can be a bit wonky at times.

ddavis-stsci (approved):
LGTM

PR #707, opened 2023-05-31
Add regression test for TweakReg.

mairanteodoro:
Resolves [RCAL-489](https://jira.stsci.edu/browse/RCAL-489)

This PR implements the regression test for `TweakReg`.

The regression test uses one of the latest `WFISim` simulated images located under `/grp/roman/TEST_DATA/23Q3_B10/DMS/SOURCE_DETECTION`, and it is currently passing:

- https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/253/.

**The regression test is executed with an association file as input**.

For now, we only use one SCA file (`r0000401001001001001_01101_0001_WFI01_cal.asdf`) that has been processed during its creation via `WFSIm`. Three additional `romancal` steps were performed on the simulated file:
- `assign_wcs`;
- `photom`;
- `source_detection`.

Both [input](https://bytesalad.stsci.edu:443/artifactory/roman-pipeline/dev/WFI/image/r0000401001001001001_01101_0001_WFI01_cal_tweakreg.asdf) and [truth](https://bytesalad.stsci.edu:443/artifactory/roman-pipeline/dev/truth/WFI/image/r0000401001001001001_01101_0001_WFI01_cal_twkreg_proc.asdf) files have been uploaded to Artifactory.

The DMS number was found using the following connections:

```mermaid
graph TD
    A[https://jira.stsci.edu/browse/RCAL-489] -- epic --> B[https://jira.stsci.edu/browse/RCAL-428]
    B[https://jira.stsci.edu/browse/RCAL-428] -- parent link --> C[https://jira.stsci.edu/browse/RDM-3193]
    C[https://jira.stsci.edu/browse/RDM-3193] -- parent link -->  D[https://jira.stsci.edu/browse/RDM-3072]
    D[https://jira.stsci.edu/browse/RDM-3072] -- issue links/relates to --> E[https://jira.stsci.edu/browse/RSOCREQ-113]
    E[https://jira.stsci.edu/browse/RSOCREQ-113] -- issue links/is a parent of --> F[https://jira.stsci.edu/browse/RSUBREQ-946]
    F[https://jira.stsci.edu/browse/RSUBREQ-946] -- DOORs Requirement ID --> G(DMS280)
```

Links to related tickets:
- https://jira.stsci.edu/browse/RCAL-489
- https://jira.stsci.edu/browse/RCAL-428
- https://jira.stsci.edu/browse/RDM-3193
- https://jira.stsci.edu/browse/RDM-3072
[...]

PR #1160, opened 2024-03-25
RCAL-800: Set flux step status for each input

stscieisenhamer:
Resolves [RCAL-800](https://jira.stsci.edu/browse/RCAL-800)

Related PRs:
- [roman_datamodel#332](https://github.com/spacetelescope/roman_datamodels/pull/332)
- [rad#395](https://github.com/spacetelescope/rad/pull/395)

This PR sets the step completion status for each input model to FluxStep.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,11 @@ dq_init
 - Allow ``dq_init`` to pass through keys not defined in ``RampModel``
   schema [#1151]
 
+flux
+----
+
+- Set flux step status for each input. [#1160]
+
 resample
 --------
 
--- a/romancal/flux/flux_step.py
+++ b/romancal/flux/flux_step.py
@@ -70,11 +70,8 @@ class FluxStep(RomanStep):
             )
 
         for model in input_models:
-            try:
-                apply_flux_correction(model)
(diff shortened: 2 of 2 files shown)


ddavis-stsci (approved):
LGTM

schlafly (commented):
This looks fine to me.  Do you think there are cases where we want the pipeline to continue when we have failed to flux calibrate?  My sense is that if we try to flux calibrate and fail something has gone badly wrong and we don't want to just continue and mark as skipped, but maybe I'm missing some case where that would be the desired behavior?

schlafly:
That sounds like a good approach to me.

schlafly (approved):
This looks good to me.  Presumably it will break the regression tests, though, and we'll need to do an okify; are you able to do that?

stscieisenhamer:
In principle I can and will attempt it.

stscieisenhamer:
[Regtest #707](https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/707/)

schlafly:
Looks good, yup, go ahead and update and okify.

PR #1161, opened 2024-03-25
RCAL-775: initial commit of tessel matching code

perrygreenfield:
This allows finding what tessels (defined in a table of tessels) overlap a given image (either by supplied ra, dec corners for the image, or a GWCS instance for the image.

There are some clearly open issues to deal with that I'm not sure. 

- Currently I have a subset of the tessel table in the test directory. 
- Does it need to be in a data directory?
- Currently it uses an environmental variable to locate the tessel table
- For the test file to find the subset table, it modifies os.environ before importing the `tessel_match` but flake8 doesn't like that. What is the right way to do that.
- The current module has some plotting utilities for visualizing the overlaps. Should these go in a separate module?
- For now I use tessel as the name for these things. This may need to change. Is there a better initial name?


romancal/tessel_match/tessel_match.py
```diff
@@ -0,0 +1,291 @@
...
+    # # Organize corners into two ra, dec lists
+    # imra = np.array([point[0] for point in image_corners])
+    # imdec = np.array([point[1] for point in image_corners])
+    vec_im_corners = image_coords_to_vec(image_corners)
+    # Approximate center of image by averaging corner vectors
+    im_center = normalize_vector(vec_im_corners.mean(axis=1))
+    # Compute difference vector between image centerand tessel centers
+    diff = vec_centers - im_center
+    dist = np.sqrt((diff**2).sum(axis=1)) * 180 / np.pi
+    match = np.where(dist < 0.5)
```
schlafly, line 127: Is 0.5 here a magic constant?  Would it make sense to pick something derived from the max linear size of the skycells or image?

romancal/tessel_match/tessel_match.py
```diff
@@ -0,0 +1,291 @@
...
+    pvec_im_corners = np.zeros((5, 3))
+    pvec_im_corners[:4] = vec_im_corners.transpose()
+    pvec_im_corners[4] = pvec_im_corners[0]
+    impoly = sgp.SingleSphericalPolygon(pvec_im_corners, im_center)
+    realmatch = []
+    for i in range(ncandidates):
+        # print(i)
+        cellpoints = points[:, :, i].transpose()
+        cellcenter = mcenters[i]
+        cellpoly = sgp.SingleSphericalPolygon(cellpoints, cellcenter)
```
schlafly, line 153: Mostly as a note to myself: for tangent plane projections, I think the edges of the skycells are "small circles" rather than "great circles" and so they are not exactly described by a spherical polygon.  The image itself has some other non-trivial border shape.  This will be a good approximation but not exact.

romancal/tessel_match/tessel_match.py
```diff
@@ -0,0 +1,291 @@
...
+import asdf
+import gwcs.wcs as wcs
+import numpy as np
+import spherical_geometry.polygon as sgp
+import spherical_geometry.vector as sgv
+from matplotlib import pyplot as plt
+
+RAD_TO_ARCSEC = 180.0 / np.pi * 3600.0
+
+plt.ion()
```
schlafly, line 19: It feels like the matplotlib dependencies should be limited to the plotting functions?  Presently matplotlib is only an optional dependency of romancal, which is included if you also want to be able to build the docs.

schlafly (commented):
This looks good to me.

On the algorithm side, the biggest question to me is whether we should include some kind of margin?  Right now I understand 'overlapping' as defined as whether the great circle polygons through the corners overlap, but neither the skycells nor the images will exactly match the great circle polygons.  We should probably add a margin to one or both of the skycells or the images to make sure we never miss a case when there is overlap.  I don't immediately see a good way to do this in the code; I could imagine something like "expand_corners" that takes the original corners, computes the middle, the differences between the corners and the middle, and adds a little bit more in that direction to each corner vector before renormalizing and returning.  I suspect 1% will be more than enough.  I think including too many matches is largely harmless; we'll read in images we don't have to but we'll find they don't intersect the image and won't do anything.  Missing overlaps also won't touch many pixels but feels bad.

On the other hand, the nominal tangent plane skycells are probably ~concave compared to the spherical polygons that are ~convex, so maybe that already covers the margin there.  I don't have an intuition for whether real image edges are closer to concave or convex.

I feel like we should not make matplotlib a requirement to import this file.

I wasn't able to figure out where find_closest_tangent_point was used in the main code.  I guess it's only needed for plotting?  It wasn't clear to me what it was adding on top of find_tessel_matches.

romancal/tessel_match/tessel_match.py
perrygreenfield: No, 0.5 isn't magic and it is certainly way larger than it needs to be. Picking an optimum value may need experimentation or as you describe. It doesn't cost a lot to be a little conservative on it.

perrygreenfield: As for adding some buffer region around the given corners, I had the same thought. I wouldn't put that in the table, but have the code do that (something like the equivalent to parallel sides to the existing sides but bumped out some fixed amount. It's not trivial, but not a big deal to implement on the fly.

romancal/tessel_match/tessel_match.py
perrygreenfield: I suspected as much, but referring to the previous discussion, it would be simpler just to pad the sides by the worst case.

romancal/tessel_match/tessel_match.py
perrygreenfield: I am inclined to move that stuff into a separate module. I do think it has some use early on for diagnosing problems and getting a good feel for what is going on. And it should be optional I agree.

romancal/tessel_match/tessel_match.py
mcara: Why not add extra vertices to the `SphericalPolygon` instead of padding?

perrygreenfield: Then the question becomes how many do you need and that may still leave (for convex borders) edge (ahem) cases that happen rarely. I heard from someone that spherical_geometry speed decreases rapidly with many vertices. I'll try to remember who that was.

perrygreenfield: It also makes the patch table much bigger. The current on based on Eddie's original is already 1.2 GB in size.

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+    mra2 = ptab[match]["ra_corn2"]
+    mra3 = ptab[match]["ra_corn3"]
+    mra4 = ptab[match]["ra_corn4"]
+    mdec1 = ptab[match]["dec_corn1"]
+    mdec2 = ptab[match]["dec_corn2"]
+    mdec3 = ptab[match]["dec_corn3"]
+    mdec4 = ptab[match]["dec_corn4"]
+    mcenters = vec_centers[match]
+    mra = np.vstack([mra1, mra2, mra3, mra4, mra1])
+    mdec = np.vstack([mdec1, mdec2, mdec3, mdec4, mdec1])
```
mcara, line 141: Are `mra*` and `mdec*` scalars or vectors? Why do you use `vstack` here? `vstack` applied to a list of scalars results in a 2D array. The documentation on `lonlat_to_vector` arguments says:
```
lon, lat : scalars or 1-D arrays
```
It may work with higher dimensionality arrays but this is an undocumented feature.

mcara:
It is difficult to review this PR without a description of these tables, how patches are defined, tessels, etc., and how they are provided. I think it needs some documentation on expected format of input data.

perrygreenfield:
@mcara regarding the table, some of the details may change, but really the code only depends on ra, dec components for determining whether a patch is touched by the supplied image (in effect). I think that part is fairly straightforward. There are other table columns that will need to be used by other programs (e.g., drizzle) for constructing output images and corresponding WCS models for those data. And information about the meaning for those columns will need to be developed. I started an outline for this at https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=RI&title=Skymap+tessellation Can you see this page since I think access is limited (if not let me know and I will request access for you). The outline is in the comments near the end.

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+import numpy as np
+import spherical_geometry.polygon as sgp
+import spherical_geometry.vector as sgv
+from matplotlib import pyplot as plt
+
+RAD_TO_ARCSEC = 180.0 / np.pi * 3600.0
+
+plt.ion()
+
+PATCH_TABLE = None
```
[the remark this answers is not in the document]
perrygreenfield: The use of a global is something I would also like to avoid, and I'm sure this can be done better. Though note that the idea here is avoiding reloading the table multiple times if multiple images are to be processed. This global could be encapsulated in higher level function that works on an association, or as a class attribute.

romancal/patch_match/patch_match.py
mcara: > Did you mean "applied to a list of 1-D arrays"?

No. I did not mean that. I assumed `mra1` is a scalar => `mra` is a 2D array (why do you need a 2D array, assuming `mra1` is a scalar). I used "assuming" and that is why I started my question with _"Are `mra*` and` mdec*` scalars or vectors?"_

mcara, line 141:
```suggestion
    mra = np.array([ptab[match][f"ra_corn{i:1d}"] for i in range(1, 5)])
    mdec = np.array([ptab[match][f"dec_corn{i:1d}"] for i in range(1, 5)])
    mcenters = vec_centers[match]
```

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+    mdec1 = ptab[match]["dec_corn1"]
+    mdec2 = ptab[match]["dec_corn2"]
+    mdec3 = ptab[match]["dec_corn3"]
+    mdec4 = ptab[match]["dec_corn4"]
+    mcenters = vec_centers[match]
+    mra = np.vstack([mra1, mra2, mra3, mra4, mra1])
+    mdec = np.vstack([mdec1, mdec2, mdec3, mdec4, mdec1])
+    points = np.array(sgv.lonlat_to_vector(mra, mdec))
+    # Create polygon for supplied image_corners
+    pvec_im_corners = np.zeros((5, 3))
```
mcara, line 144:
```suggestion
    pvec_im_corners = np.empty((5, 3))
```

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+        if impoly.intersects_poly(cellpoly):
+            # print(f"candidate {i} intersects")
+            realmatch.append(i)
+    return match[0][realmatch], match[0]
+
+
+def get_corners(patch):
+    """
+    Construct a the vertex coordinates for a patch from a patch definition suitable
+    for plotting the defined region (x coordinates, y coordinates).
```
mcara, line 163: this is not clear at all. it seems that this function returns Cartesian coordinates of vertex vectors `(x, y, z)` but the comment talks about plotting "the defined region" (x coordinates, y coordinates). What happened to the `z` coordinate?

Also, I would have expected this function to return the value of the variable `corners` - not Cartesian coordinates. So, maybe rename it to `get_cartesian_corners`.

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    p = patch
```
mcara, line 165: Just use `patch`.

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,291 @@
...
+            p["dec_corn1"],
+            p["dec_corn2"],
+            p["dec_corn3"],
+            p["dec_corn4"],
+            p["dec_corn1"],
+        ),
+        (p["ra_corn1"], p["ra_corn2"], p["ra_corn3"], p["ra_corn4"], p["ra_corn1"]),
+    )
+    corners = np.array(corners)
+    vec_corners = sgv.lonlat_to_vector(corners[0], corners[1])
```
mcara, line 177:
```suggestion
    ra = np.array([patch[f"ra_corn{(i % 4) + 1:1d}"] for i in range(5)])
    dec = np.array([patch["dec_corn{(i % 4) + 1:1d}"] for i in range(5)])
    vec_corners = sgv.lonlat_to_vector(ra, dec)
```

mcara: or, maybe even better, since this is used frequently in the code:

Make this a global constant at the top of the module:
```python
_CORNER_LABELS = ('1', '2', '3', '4', '1')
```
and then change this code as:
```suggestion
    rd = [(patch["ra_corn" + lb], patch["dec_corn" + lb]) for lb in _CORNER_LABELS]
    vec_corners = sgv.lonlat_to_vector(*zip(*rd))
```

romancal/patch_match/patch_match.py
perrygreenfield: I'll await the comment of others on this change. I was tempted to do something like this, but things like this make the code more obscure. If there were many more such like this, sure this change make sense. So I wonder what others think on this issue.

romancal/patch_match/patch_match.py
perrygreenfield: Is there a real substantial difference in this?

romancal/patch_match/patch_match.py
perrygreenfield: I changed the x, y to ra, dec, and changed the function to the name suggested.

romancal/patch_match/patch_match.py
perrygreenfield: Done.

romancal/patch_match/patch_match.py
mcara: it doesn't look complicated and it is just 2 lines vs 12.

romancal/patch_match/patch_match.py
mcara: Substantial or not, it's a waste and why not fix it in a next commit?

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,299 @@
...
+        (
+            patch["ra_corn1"],
+            patch["ra_corn2"],
+            patch["ra_corn3"],
+            patch["ra_corn4"],
+            patch["ra_corn1"],
+        ),
+    )
+    corners = np.array(corners)
+    vec_corners = sgv.lonlat_to_vector(corners[0], corners[1])
```
mcara, line 185: Even if you do not like my suggestion with list comprehension and `zip`, my not create two separate lists `ra_corners` and `dec_corners` instead of punching them into a single array and then extracting rows?

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,299 @@
...
+            if tp == patch_tp:
+                patch_tp_id.append(i)
+    return closest_tangent_point, patch_tp_id
+
+
+def normalize_vector(vec):
+    """
+    Normalize a 3d vector to have length 1. Only works on 1d arrays.
+    """
+    return vec / np.sqrt((vec**2).sum())
```
mcara, line 230: Why reinvent this? Just use https://github.com/spacetelescope/spherical_geometry/blob/5de7fc79d8e3cb4f1c0a9338acf06b350d17b15b/spherical_geometry/vector.py#L129-L140

It also works on Nx3 vectors.

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,299 @@
...
+    ultra-precision). This also breaks down numerically very near the
+    poles.
+    """
+    # First compute the tangent plane axis vectors.
+    x_axis = normalize_vector(np.cross([0, 0, 1], tangent_point_vec))
+    y_axis = normalize_vector(
+        np.array([0, 0, 1])
+        - np.array(tangent_point_vec)
+        * np.dot(np.array([0, 0, 1]), np.array(tangent_point_vec))
+    )
```
mcara, line 247:
```suggestion
    z1 = np.array([0.0, 0.0, 1.0])
    y_axis = normalize_vector(
        z1 - np.array(tangent_point_vec) * tangent_point_vec[2]
    )
```

mcara: OR
```suggestion
    y_axis = normalize_vector(
        np.subtract([0.0, 0.0, 1.0], np.array(tangent_point_vec) * tangent_point_vec[2])
    )
```

romancal/patch_match/patch_match.py
```diff
@@ -0,0 +1,299 @@
...
+def veccoords_to_tangent_plane(vertices, tangent_point_vec):
+    """
+    Convert the spherical geometry vectors to tangent plane coordinates
+    in arcseconds. This algorithm is not precise, but should be good
+    enough for now (and besides, the goal here is visualizaion, not
+    ultra-precision). This also breaks down numerically very near the
+    poles.
+    """
+    # First compute the tangent plane axis vectors.
+    x_axis = normalize_vector(np.cross([0, 0, 1], tangent_point_vec))
```
mcara, line 242: consider using https://github.com/spacetelescope/spherical_geometry/blob/5de7fc79d8e3cb4f1c0a9338acf06b350d17b15b/spherical_geometry/great_circle_arc.py#L64-L79

it is likely more accurate due to quad math library

ddavis-stsci (approved):
For my test images the matching code works as expected.

nden (approved):
I tested the code using functions in gwcs and  it works correctly.

schlafly (approved):
Looks good to me, thanks Perry!

romancal/patch_match/test/test_patch_match.py
```diff
@@ -0,0 +1,193 @@
...
+import numpy as np
+import pytest
+import spherical_geometry.vector as sgv
+
+import romancal.patch_match.patch_match as pm
+
+mpath = inspect.stack()[0][1]
+pm.load_patch_table(os.path.join(os.path.split(mpath)[0], "patches_subset.asdf"))
+rotate = sgv.rotate_around
+absindex = 925050
```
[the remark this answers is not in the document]
perrygreenfield: Thanks. Perhaps I was inadvertently reading the full table locally. I'll fix the code to deal with working with a subset table.

braingram:
I rebased this to pull in the changes in main and to re-trigger the docs build (which failed due to a failure to connect to http://matplotlib.org which is likely temporary). Prior to the rebase all tests except the docs were passing.

PR #1170, opened 2024-04-08
RCAL-786: update default DQ flags in romancal

mairanteodoro:
Resolves [RCAL-786](https://jira.stsci.edu/browse/RCAL-786)

This PR updates the default DQ flags used in `resample_step`, `outlier_detection_step`, and `skymatch_step` to also include the `DEAD` flag. The default is now `DO_NOT_USE+NON_SCIENCE+DEAD`.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ general
 
 - Update the high level pipeline to use updates in Outlier_detection and tweakreg [#1143]
 
+- Update the default DQ flags to '``~DO_NOT_USE+NON_SCIENCE+DEAD``' in
+  ``resample_step``, ``outlier_detection_step``, and ``skymatch_step``. [#1170]
+
 documentation
 -------------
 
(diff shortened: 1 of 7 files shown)


schlafly:
In our discussion at the last tag up, we decided that we should follow the Webb precedent and require that things like DEAD also have DO_NOT_USE set in the mask files.  We need only be responsible for "new" reasons to set a pixel as DO_NOT_USE.  The only one of those of which I am aware is when a pixel is too saturated to such that no ramp can be fit.  We address that case here:
https://github.com/spacetelescope/romancal/blob/main/romancal/ramp_fitting/ramp_fit_step.py#L367-L376

So I think we can close this DR without any new development work.  I don't think we need to add DEAD to this list, since it should imply DO_NOT_USE in the mask files.  i.e., I think we can just close this PR---sorry!

PR #1171, opened 2024-04-09
Update HLP regtest to use existing L1/L2 files

schlafly:
This PR updates the HLP regtest to use existing L1/L2 files rather than add new ones.  This reduces the number of files we need to track.

It also adds the corresponding L3 image creation to the regtest file creation script.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below. https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/705/


ddavis-stsci (approved):
LGTM

PR #1181, opened 2024-04-12
RCAL-795: fix incorrect weight maps returned by resampling

mairanteodoro:
Resolves [RCAL-795](https://jira.stsci.edu/browse/RCAL-795)

This PR fixes a bug in the resample step in which the output weight and context arrays were being reset to zero every time a new image was added to the drizzle object.

**Regression tests**
No issues due to the changes in this PR.
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/706/

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,8 @@ resample
 
 - Updated Level 3 ``cal_step`` attribute creation. [#1165]
 
+- Fix bug that prevented properly update of the resampled output weight and context arrays. [#1181]
+
 outlier_detection
 -----------------
 
--- a/romancal/resample/gwcs_drizzle.py
+++ b/romancal/resample/gwcs_drizzle.py
@@ -84,7 +84,7 @@ class GWCSDrizzle:
 
         self.outsci = product.data
         self.outwcs = outwcs or product.meta.wcs
-        self.outwht = np.zeros(self.outsci.shape, dtype=np.float32)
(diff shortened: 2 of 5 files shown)


schlafly:
Thanks, that looks right to me.  Can you add an image with an example weight map for some coadded dithered data so we can visually see that the new weight maps look sensible?  e.g., large where all the images overlap, smaller on the edges, some zero regions where no image touches?  Thanks for catching the context images too; it would be good to show something like context & 2**i for the different images to see that those now work too.

mairanteodoro:
The image below shows three dithered mock images used in `test_resample.test_resampledata_do_drizzle_default_single_exposure_weight_array`.

Screenshot 2024-04-15 at 4 58 55 PM

The images have all the same size (~11.2"x11.2") and are shifted by 5". The image ID is displayed in white on the top left corner of each image.

The colors indicate the total weight for each pixel, whereas the black points indicate the contribution of the images at that position as stored in the context image. I haven't added bad pixels to the images, so for each image the weight map is a matrix of ones.

The results show that both the weight array and the context image are being updated properly now.

schlafly (approved):
Great, thanks Mairan.  That looks good.  This will presumably need a regtest update since it will change the weight and context arrays.  Are you able to make that update?

PR #1182, opened 2024-04-16
RCAL-799: Populate the Level 3 wcsinfo

stscieisenhamer:
Resolves [RCAL-700](https://jira.stsci.edu/browse/RCAL-799)

This PR fills in the Level 3 wcsinfo meta

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/romancal/assign_wcs/utils.py
+++ b/romancal/assign_wcs/utils.py
@@ -358,7 +358,7 @@ def compute_fiducial(wcslist, bounding_box=None):
     return fiducial
 
 
-def create_footprint(wcs, shape=None):
+def create_footprint(wcs, shape=None, center=True):
     """Calculate sky footprint
 
     Parameters
@@ -368,6 +368,9 @@ def create_footprint(wcs, shape=None):
 
     shape : n-tuple or None
        Shape to use if wcs has no defined shape.
+
+    center : bool
+        If True use the center of the pixel, otherwise use the corner.
 
     Returns
     -------
@@ -381,7 +384,7 @@ def create_footprint(wcs, shape=None):
 
     # footprint is an array of shape (2, 4) - i.e. 4 values for RA and 4 values for
     # Dec - as we are interested only in the footprint on the sky
-    footprint = wcs.footprint(bbox, center=True, axis_type="spatial").T
+    footprint = wcs.footprint(bbox, center=center, axis_type="spatial").T
     # take only imaging footprint
     footprint = footprint[:2, :]
 
--- a/romancal/assign_wcs/utils.py
+++ b/romancal/assign_wcs/utils.py
@@ -358,6 +358,59 @@ def compute_fiducial(wcslist, bounding_box=None):
     return fiducial
 
 
+def create_footprint(wcs, shape=None):
+    """Calculate sky footprint
+
+    Parameters
+    ----------
+    wcs : `gwcs.WCS`
+        The WCS information to get the footprint from
+
+    shape : n-tuple or None
+       Shape to use if wcs has no defined shape.
+
+    Returns
+    -------
+    footprint : `numpy.ndarray`
+        The footprint.
+    """
+    bbox = wcs.bounding_box
+
+    if bbox is None:
+        bbox = wcs_bbox_from_shape(shape)
+
+    # footprint is an array of shape (2, 4) - i.e. 4 values for RA and 4 values for
+    # Dec - as we are interested only in the footprint on the sky
+    footprint = wcs.footprint(bbox, center=True, axis_type="spatial").T
+    # take only imaging footprint
+    footprint = footprint[:2, :]
+
+    # Make sure RA values are all positive
+    negative_ind = footprint[0] < 0
+    if negative_ind.any():
+        footprint[0][negative_ind] = 360 + footprint[0][negative_ind]
+
+    footprint = footprint.T
+    return footprint
+
+
+def create_s_region(footprint):
+    """Create the properly formatted string to fill the S_REGION FITS keyword
+
+    Parameters
+    ----------
+    footprint : `numpy.ndarray`
+        The footprint to be represented.
+
+    Returns
+    -------
+    s_region : str
+        The formatted string that can be directly assigned to the FITS S_REGION keyword.
+    """
+    s_region = "POLYGON ICRS " + " ".join([str(x) for x in footprint.ravel()]) + " "
+    return s_region
+
+
 def add_s_region(model):
     """
     Calculate the detector's footprint using ``WCS.footprint`` and save it in the
@@ -372,29 +425,11 @@ def add_s_region(model):
     -------
     A formatted string representing the detector's footprint
     """
-
-    bbox = model.meta.wcs.bounding_box
-
-    if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
-
-    # footprint is an array of shape (2, 4) - i.e. 4 values for RA and 4 values for
-    # Dec - as we are interested only in the footprint on the sky
-    footprint = model.meta.wcs.footprint(bbox, center=True, axis_type="spatial").T
-    # take only imaging footprint
-    footprint = footprint[:2, :]
-
-    # Make sure RA values are all positive
-    negative_ind = footprint[0] < 0
-    if negative_ind.any():
-        footprint[0][negative_ind] = 360 + footprint[0][negative_ind]
-
-    footprint = footprint.T
-    update_s_region_keyword(model, footprint)
+    update_s_region_keyword(model, create_footprint(model.meta.wcs, shape=model.shape))
 
 
 def update_s_region_keyword(model, footprint):
-    s_region = "POLYGON ICRS " + " ".join([str(x) for x in footprint.ravel()]) + " "
+    s_region = create_s_region(footprint)
     log.info(f"S_REGION VALUES: {s_region}")
     if "nan" in s_region:
         # do not update s_region if there are NaNs.
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -751,7 +751,7 @@ def gwcs_into_l3(model, wcs):
     l3_wcsinfo.y_ref = -transform['crpix2'].offset.value
     l3_wcsinfo.pixel_scale = (transform['cdelt1'].factor.value + transform['cdelt2'].factor.value) / 2.0
 
-    world_center = wcs(*[v / 2. for v in model.shape[::-1]])
+    world_center = wcs(*[(v - 1) / 2. for v in model.shape[::-1]])
     l3_wcsinfo.ra_center = world_center[0]
     l3_wcsinfo.dec_center = world_center[1]
     l3_wcsinfo.pixel_scale_local = compute_scale(wcs, world_center)
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -2,6 +2,7 @@ import logging
 import math
 from typing import List
 
+from astropy.coordinates import SkyCoord
 import numpy as np
 from astropy import units as u
 from drizzle import cdrizzle, util
@@ -748,6 +749,7 @@ def gwcs_into_l3(model, wcs):
     l3_wcsinfo.ra_center = world_center[0]
     l3_wcsinfo.dec_center = world_center[1]
     l3_wcsinfo.pixel_scale_local = compute_scale(wcs, world_center)
+    l3_wcsinfo.orientat_local = calc_pa(wcs, *world_center)
 
     try:
         footprint = utils.create_footprint(wcs, model.shape)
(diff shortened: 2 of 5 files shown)


stscieisenhamer:
Getting this draft pr out there because I think I have too simple of math the scale and orientation calculations. Better minds need to review.

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
+    l3_wcsinfo.x_ref = -transform.offset_0.value  # cdelt1
+    l3_wcsinfo.y_ref = -transform.offset_1.value  # cdelt2
+    l3_wcsinfo.pixel_scale = (transform.factor_3.value + transform.factor_4.value) / 2.0
+
+    world_center = wcs(*[v / 2.0 for v in model.shape])
+    world_center_plus = wcs(*[(v / 2.0) + 1 for v in model.shape])
+    world_delta = [w_plus - w for w_plus, w in zip(world_center_plus, world_center)]
+    l3_wcsinfo.ra_center = world_center[0]
+    l3_wcsinfo.dec_center = world_center[1]
+    l3_wcsinfo.pixel_scale_local = (world_delta[0] + world_delta[1]) / 2.0
```
schlafly, line 754: world_center_plus = wcs(center + 1, center + 1), so it corresponds to sqrt(2) pixels difference.  I start getting confused about what this is doing if things aren't aligned neatly with ra/dec.  These are also on the sphere, but we want pixel_scale to be in physical coordinates (i.e., if we're near dec 90, we still want pixels that are 1" away from one another to have values of ~1 here).  I think if we want to do this "right" we should do the following:
- get central pixel coordinates
- get one pixel up coordinates
- get one pixel to the right coordinates
- convert to SkyCoord
- get separations from central pixel to neighbors via c_center.separation(c_next)
- get angle between coordinates: c_center.position_angle(c_1) - c_center.position_angle(c_2).  Should be very close to 90!
- compute pixel area as separation_1 * separation_2 * np.abs(np.sin(angle))
- compute scale as sqrt(area)

That at least gives something that is rotation invariant and shouldn't have weird behavior near the poles, though I'm open to other suggestions.

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
+    footprint = utils.create_footprint(wcs, model.shape)
+    l3_wcsinfo.ra_corn1 = footprint[0][0]
+    l3_wcsinfo.ra_corn2 = footprint[1][0]
+    l3_wcsinfo.ra_corn3 = footprint[2][0]
+    l3_wcsinfo.ra_corn4 = footprint[3][0]
+    l3_wcsinfo.dec_corn1 = footprint[0][1]
+    l3_wcsinfo.dec_corn2 = footprint[1][1]
+    l3_wcsinfo.dec_corn3 = footprint[2][1]
+    l3_wcsinfo.dec_corn4 = footprint[3][1]
+    l3_wcsinfo.s_region = utils.create_s_region(footprint)
```
schlafly, line 765: This looks good.  FWIW, I think this is giving the "outer corners of the corner pixels", rather than the centers of the corner pixels.  I think that's a good thing.

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
+
+    rm = l3_wcsinfo.rotation_matrix
+    xrot = abs(math.atan2(rm[1][0], rm[0][0]))
```
schlafly, line 768: Let's compute orientat_local as c_center.position_angle(c_up) using the SkyCoord created above.  We'll define it as the position angle relative to north of the Y axis of the image.

For orientat what you have could be okay, but I'm leery about assuming transform.pc_rotation_matrix exists?

@nden , conceptually we want to fill out these fields as appropriate for a tangent plane projection.  But the gWCS object can support ~arbitrary transformations and one could imagine stuffing in an extra no-op layer into the gWCS object so that reaching into the guts of the gWCS and expecting the parameters to be the right ones for the tangent plane projection seems dangerous.  It's probably true that we could get all of the "local" values only using the forward transformation and local differences.  We could also get all of the projection center values if we're allowed to evaluate the forward transformation at x_ref, y_ref, which will be outside the image but everything will be well defined there; i.e., the only reason we couldn't evaluate it there would be if we set the bounding box to explicitly prevent that.  The only thing I don't know how to get would be one of the ra/dec reference or the x/y reference point, which conceptually only exists in a tangent plane projection and so I don't know how to ask the generic gWCS object for it without supposing a form.  Any ideas there?

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
-    # l3_wcsinfo.ra_corn4 = from bounding box
-    # l3_wcsinfo.dec_corn1 = from bounding box
-    # l3_wcsinfo.dec_corn2 = from bounding box
-    # l3_wcsinfo.dec_corn3 = from bounding box
-    # l3_wcsinfo.dec_corn4 = from bounding box
-    # l3_wcsinfo.orientat = ???
-    # l3_wcsinfo.orientat_local = ???
+    l3_wcsinfo.x_ref = -transform.offset_0.value  # cdelt1
+    l3_wcsinfo.y_ref = -transform.offset_1.value  # cdelt2
+    l3_wcsinfo.pixel_scale = (transform.factor_3.value + transform.factor_4.value) / 2.0
```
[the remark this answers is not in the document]
stscieisenhamer: Would the preference be to make the code more roman-specific and drop the averaging? Or leave as is in the wild chance to never have to look at this code again, even if pixels become non-square?

romancal/resample/resample.py
stscieisenhamer: duh, completely left out the conversion to sky...

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
     l3_wcsinfo = model.meta.wcsinfo
-    transform = wcsinfo.forward_transform
+    transform = wcs.forward_transform
 
     # Basic WCS info
     l3_wcsinfo.projection = "TAN"
+    l3_wcsinfo.pixel_shape = model.shape
     l3_wcsinfo.rotation_matrix = transform["pc_rotation_matrix"].matrix.value.tolist()
     l3_wcsinfo.dec_ref = transform.lat_6.value
     l3_wcsinfo.ra_ref = transform.lon_6.value
```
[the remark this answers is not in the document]
nden: The `pc_rotation_matrix` name comes from gwcs, so it should be fine though we need a check in the gwcs for uniqueness of names. We need to add also a name to the celestial rotation transform so it's easy to get to it as `lat_6` and `lon_6` are not guaranteed.

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
-    # l3_wcsinfo.ra_corn3 = from bounding box
-    # l3_wcsinfo.ra_corn4 = from bounding box
-    # l3_wcsinfo.dec_corn1 = from bounding box
-    # l3_wcsinfo.dec_corn2 = from bounding box
-    # l3_wcsinfo.dec_corn3 = from bounding box
-    # l3_wcsinfo.dec_corn4 = from bounding box
-    # l3_wcsinfo.orientat = ???
-    # l3_wcsinfo.orientat_local = ???
+    l3_wcsinfo.x_ref = -transform.offset_0.value  # cdelt1
+    l3_wcsinfo.y_ref = -transform.offset_1.value  # cdelt2
```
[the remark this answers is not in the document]
nden: The comment is wrong - it's crpix in FITS .
The current name of the models is `crpix1` and so `wcs.forward_transform['crpix1'].offset.value` is what is a better call. I don't like the current names and since I doubt they are used I may propose changing them in gwcs to `x_ref` and `y_ref`.

romancal/resample/resample.py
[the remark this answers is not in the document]
nden: The above comment applies here too - the current transform names are `cdelt1` and `cdelt2`

romancal/resample/resample.py
```diff
@@ -732,30 +734,40 @@ def gwcs_into_l3(model, wcsinfo):
...
+    l3_wcsinfo.pixel_scale = (transform.factor_3.value + transform.factor_4.value) / 2.0
+
+    world_center = wcs(*[v / 2.0 for v in model.shape])
```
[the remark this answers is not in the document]
nden: The [docstring for the L3 model maker utility](https://github.com/spacetelescope/roman_datamodels/blob/main/src/roman_datamodels/maker_utils/_datamodels.py#L162) says the shape is in Python order, (y, x). The WCS object expects inputs as (x, y) - reverse the shape above.

romancal/resample/resample.py
nden: We already have a method implementing the above. 
https://github.com/spacetelescope/stcal/blob/main/src/stcal/alignment/util.py#L402

romancal/resample/resample.py
nden: I think I addressed this above.

romancal/assign_wcs/utils.py
```diff
@@ -358,29 +358,30 @@ def compute_fiducial(wcslist, bounding_box=None):
...
+    bbox = wcs.bounding_box
 
     if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
+        bbox = wcs_bbox_from_shape(shape)
 
     # footprint is an array of shape (2, 4) - i.e. 4 values for RA and 4 values for
     # Dec - as we are interested only in the footprint on the sky
-    footprint = model.meta.wcs.footprint(bbox, center=True, axis_type="spatial").T
+    footprint = wcs.footprint(bbox, center=True, axis_type="spatial").T
```
mairanteodoro, line 384: Just out of curiosity, should we always use the center of the pixel (`center=True`) when creating the footprint? Wouldn't it be better to allow that parameter to be passed in when calling `create_footprint()`, and set its default value to `True`?

mairanteodoro (commented):
Thanks, @stscieisenhamer! It looks good to me. I left a couple of comments.

romancal/resample/resample.py
```diff
@@ -732,30 +735,69 @@ def gwcs_into_l3(model, wcsinfo):
...
+        l3_wcsinfo.s_region = utils.create_s_region(footprint)
+
+    try:
+        l3_wcsinfo.rotation_matrix = transform[
+            "pc_rotation_matrix"
+        ].matrix.value.tolist()
+        l3_wcsinfo.dec_ref = transform.lat_6.value
+        l3_wcsinfo.ra_ref = transform.lon_6.value
+        l3_wcsinfo.x_ref = -transform["crpix1"].offset.value
+        l3_wcsinfo.y_ref = -transform["crpix2"].offset.value
```
schlafly, line 771: Along the same lines as Mairan, is it the case that we only need to depend on one of either
crpix1, crpix2, or
lat_6, lon_6
existing?  I guess I'd prefer the former and get the latter two via wcs.pixel_to_world(x_ref, y_ref) if that doesn't throw an out-of-bounds error?

nden: No, it's not guaranteed to always have lat_6 or lon_6. The number attached to the end comes from the order of the models. It's better to index the compound transform with the model name to retrieve the parameters.
 I commented above on the correct way to get to the values but those comments were resolved without addressing them.

stscieisenhamer: Yes, needed to ask about that. That transform does not have a name and will need to be added. Any suggestions on a name?

FYI: the model currently appears as follows:
```
Components:
    [0]: <Shift(offset=-118.80708394, name='crpix1')>

    [1]: <Shift(offset=-90.02345166, name='crpix2')>

    [2]: <AffineTransformation2D(matrix=[[-0.93358043,  0.35836795], [ 0.35836795,  0.93358043]], translation=[0., 0.], name='pc_rotation_matrix')>

    [3]: <Scale(factor=0.00002778, name='cdelt1')>

    [4]: <Scale(factor=0.00002778, name='cdelt2')>

    [5]: <Pix2Sky_Gnomonic()>

    [6]: <RotateNative2Celestial(lon=10.0029245, lat=0.0015345, lon_pole=180.)>
```
and
```
>>> print(wcs.forward_transform.submodel_names)
('crpix1', 'crpix2', 'pc_rotation_matrix', 'cdelt1', 'cdelt2', 'None_0', 'None_1')
```

schlafly: I think if crpix1 & crpix2 work, we can get lat_6 and lon_6 by wcs.pixel_to_world(x_ref, y_ref) in a generic way, so that solves the issue?

stscieisenhamer:
[Regressions Tests attempt 1](https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/717/)

romancal/resample/resample.py
```diff
@@ -732,30 +735,79 @@ def gwcs_into_l3(model, wcsinfo):
...
+
+    try:
+        l3_wcsinfo.rotation_matrix = transform[
+            "pc_rotation_matrix"
+        ].matrix.value.tolist()
+    except Exception:
+        log.warning(
+            "WCS has no clear rotation matrix defined by pc_rotation_matrix. Calculating one."
+        )
+        l3_wcsinfo.rotation_matrix = utils.calc_rotation_matrix(l3_wcsinfo.orentat, 0.0)
```
schlafly, line 788:
```suggestion
        l3_wcsinfo.rotation_matrix = utils.calc_rotation_matrix(l3_wcsinfo.orientat, 0.0)
```

schlafly (approved):

romancal/resample/resample.py
stscieisenhamer: disturbing that got through all the checks. adding more checks...

PR #1188, opened 2024-04-18
RCAL-816: add function to populate L3 mosaic meta.basic.

mairanteodoro:
Resolves [RCAL-816](https://jira.stsci.edu/browse/RCAL-816)

This PR adds a function to populate the L3 mosaic output from resample with basic info about times, observation setup, and others. A unit test was also implemented to test the new function.

**Regression tests**
~https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/711/~
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/712/

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -1,4 +1,5 @@
 import logging
+from typing import List
 
 import numpy as np
 from astropy import units as u
@@ -7,7 +8,6 @@ from roman_datamodels import datamodels, maker_utils
 
 from ..datamodels import ModelContainer
 from . import gwcs_drizzle, resample_utils
-from typing import List
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -7,6 +7,7 @@ from roman_datamodels import datamodels, maker_utils
 
 from ..datamodels import ModelContainer
 from . import gwcs_drizzle, resample_utils
+from typing import List
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
@@ -152,6 +153,9 @@ class ResampleData:
         self.blank_output = maker_utils.mk_datamodel(
             datamodels.MosaicModel, shape=tuple(self.output_wcs.array_shape)
         )
+
+        # update meta.basic
+        populate_mosaic_basic(self.blank_output, input_models)
 
         # update meta.cal_step
         self.blank_output.meta.cal_step = maker_utils.mk_l3_cal_step(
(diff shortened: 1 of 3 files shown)


romancal/resample/resample.py
```diff
@@ -759,3 +763,79 @@ def gwcs_into_l3(model, wcsinfo):
...
+    )
+    output_model.meta.basic.time_mean_mjd = np.mean(
+        [x.exposure.mid_time.mjd for x in input_meta]
+    )
+    output_model.meta.basic.max_exposure_time = np.max(
+        [x.exposure.exposure_time for x in input_meta]
+    )
+    output_model.meta.basic.mean_exposure_time = np.mean(
+        [x.exposure.exposure_time for x in input_meta]
+    )
```
schlafly, line 804: I think these two are set in update_exposure_times using the exposure time image; they are supposed to be the pixel-wise mean and maxes not the exposure-wise means and maxes.  Probably these lines can just be deleted, but check if update_exposure_times is actually setting these?

romancal/resample/tests/test_resample.py
```diff
@@ -620,3 +620,72 @@ def test_resampledata_do_drizzle_default_single_exposure_weight_array(
 
     assert np.any(output_models_many_to_one[0].weight > 0)
     assert np.any(output_models_many_to_many[0].weight > 0)
+
+
+def test_populate_mosaic_basic(exposure_1):
```
schlafly, line 625: This looks good to me, but I think I would either simplify the tests or give it some more challenging tests?  If I understand correctly, you've got the code for handling cases like multiple surveys here, but you're only testing a single exposure where most of the test code isn't needed?

schlafly (commented):
This looks good to me.  I left some minor comments in line.

romancal/resample/resample.py
mairanteodoro: Nice catch, @schlafly! Indeed, both `basic.max_exposure_time` and `basic.mean_exposure_time` are being set in `update_exposure_times`, corresponding to the pixel-wise max and mean, respectively:

https://github.com/spacetelescope/romancal/blob/76176b06ea57b1bfe0de6d47e8f2f8187a748db4/romancal/resample/resample.py#L499-L504

I removed them from the current PR.

romancal/resample/tests/test_resample.py
mairanteodoro: Done.

schlafly (commented):
This looks good to me.  @ddavis-stsci , we need to figure out how we want to populate the skycell location and association product type.  This can be a separate PR.  I think that maybe the associate file should include the skycell name and whether it is a visit or pass or full product, and then this would pull that information from the association file?

romancal/resample/resample.py
```diff
@@ -759,3 +763,73 @@ def gwcs_into_l3(model, wcsinfo):
...
+        else "MULTIPLE"
+    )
+
+    # instrument data
+    output_model.meta.basic.optical_element = input_meta[0].instrument.optical_element
+    output_model.meta.basic.instrument = input_meta[0].instrument.name
+
+    # skycell location
+    output_model.meta.basic.location_name = "TBD"
+
```
ddavis-stsci, line 833: Why is this location_name? Why not skycell_name? 
will we need the name of the projection call as well?

ddavis-stsci (commented):
This seems to be designed for the prompt products only. 
How will this be expanded for PASS and Full products?

romancal/resample/resample.py
mairanteodoro: That's the attribute's name in RAD; I concur that `skycell_name` sounds better than `location_name`.

schlafly: RTB hadn't (and still hasn't, as far as I know) decided if we want to call these skycells or patches or ....  We ended up with "location_name" as some kind of generic word.  I'd propose we fill this with the record name from the tessellation file, which includes both the projection cell part and the sky cell part.

We'll specify product_type = 'PASS' to handle passes, etc., and the pass / visit metadata will get sentinel values indicating that they span multiple visits, as in Mairan's current PR.

schlafly (approved):
Looks good, approved.  Note that this will change metadata and require more regtest updates.  Let's defer to future work, but I imagine we'll want the association files to include "product_type" and "location" information that eventually goes into the corresponding columns in the metadata.

PR #1207, opened 2024-05-01
Make sure logs get included in L3 products.

schlafly:
This PR leads HLP products to have their "cal_logs" attribute filled out.

Previously these were not being populated because stpipe.core was explicitly checking that the output datamodel was an ImageModel.  This PR changes that to check if it's either an ImageModel or a MosaicModel.  It also clears the CalLogs when making a new output model to avoid some initial filler values.

Finally, a big portion of the HLP is processing the input L2s; flux calibration, sky matching, outlier detection.  Those logs go into the L2 products which we do not save.  This adds a new individual_image_cal_logs extension parallel to cal_logs that houses the individual L2 log messages.  We probably want that?

**Checklist**
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below. https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/725/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,8 @@ stpipe
 - Update ``stpipe.core.finalize_results`` to record the CRDS information
   only if a step uses reference files. [#1201]
 
+- Populate logs for L3 files in addition to L2 files [#1207]
+
 resample
 --------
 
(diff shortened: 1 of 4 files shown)


schlafly:
Regtests are "passing"; two new failures because of the added individual_image_cal_logs bit in the asdf file.  If we like this approach, we should exclude those from compare_asdf.

schlafly:
After adding individual_image_cal_logs to the list of extensions to ignore, regtests actually pass.
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/726/

ddavis-stsci (approved):
It looks a bit messy but all the info appears to be copied.

PR #1208, opened 2024-05-01
RCAL-801: update skymatch to return updated models if the input is an asn

mairanteodoro:
Resolves [RCAL-801](https://jira.stsci.edu/browse/RCAL-801)

This PR ensures that `SkyMatchStep` always returns a `ModelContainer` with the updated datamodel.

I believe `SkyMatchStep` is due some refactoring (specifically regarding the intended use of memory management, as seen in JWST's `SkyMatchStep`), but I think that it should be done separately from the current PR.

**Regression Tests**
- https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/721/

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 0.14.1 (unreleased)
 ==================
 
+skymatch
+--------
+- Update step to always return a ``ModelContainer``. [#1208]
+
 patch_match
 -----------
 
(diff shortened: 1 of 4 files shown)


ddavis-stsci (approved):
LGTM

Issue #1154, opened 2024-03-21
HLP input filenames


stscijgbot-rstdms:
_Issue [RCAL-797](https://jira.stsci.edu/browse/RCAL-797) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

The names of the input L2 images used to create the mosaic were not correctly recorded in the metadata. Instead, the value of the L2 product metadata value meta.filename was recorded, which in our test did not match the file names on disk, which were altered using shell commands. In this case, the values of meta.filename were the original archive-style file names, which use the WFI observation identifier.

See RSOCDR-124

stscijgbot-rstdms:
_Comment by [Eddie Schlafly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eschlafly) on [JIRA](https://jira.stsci.edu/browse/RCAL-797?focusedCommentId=695105#comment-695105):_

Following the tag up on 4/19, we decided that this meant storing the actual file names rather than the meta.filenames in the resample members schema:

<https://github.com/spacetelescope/romancal/blob/main/romancal/resample/resample.py#L315>

stscijgbot-rstdms:
_Comment by [Eddie Schlafly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eschlafly) on [JIRA](https://jira.stsci.edu/browse/RCAL-797?focusedCommentId=699977#comment-699977):_

Addressed in <https://github.com/spacetelescope/romancal/pull/1209],> closing.

PR #1209, opened 2024-05-01
Put real file names in meta.resample.meta.

schlafly:
This PR puts the real file names in meta.resample.members rather than the individual file name meta.filename attributes.  This is intended to address https://github.com/spacetelescope/romancal/issues/1154 , but is rather awkward.

The challenge is that the ModelContainer passed to resample doesn't know about the original file names.  This PR digs around the private contents of the datamodel to grab the input file name, and if that fails, falls back to img.meta.filename.  But relying on img._asdf._fname is clearly asking for trouble.

The alternative would be to teach ModelContainer about the input file names if it's read from an association.  That would add a new filenames member to ModelContainer that defaults to None, and sets the new filenames to be equal to _models if everything is a string.  Then resample would check that attribute and use it if it's not None.

Thoughts on what the preferred approach is here?  @ddavis-stsci , @mairanteodoro , you may have thoughts here?

**Checklist**
- [ ] updated relevant tests
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below. https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/727/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -81,6 +81,8 @@ resample
 
 - Include logs of individual L2 products [#1207]
 
+- Resample members should use actual file names from association file [#1209]
+
 outlier_detection
 -----------------
 
--- a/romancal/datamodels/container.py
+++ b/romancal/datamodels/container.py
@@ -153,6 +153,7 @@ class ModelContainer(Sequence):
         self.asn_table = {}
         self.asn_table_name = None
         self.asn_pool_name = None
+        self.filepaths = None
 
         try:
             init = Path(init)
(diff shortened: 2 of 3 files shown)


ddavis-stsci:
I think that the container should store the asn member names. As is, the container loses the given filename and assumes that the meta.filename is the name of disk. That is clearly incorrect here. 
If I understand your code you want to reset the meta.filename for the member files and rewrite them to disk?
I'm not sure that is a good idea.
If I understood Tyler correctly they would be happy with storing the asn member filenames in the output products? 
I assume you think that this is not sufficient?

schlafly:
Yes, the alternative I proposed is adding a new attribute to the model container that contains the file names.  I can code that option up later.

mairanteodoro:
I too agree with using `ModelContainer` to handle the ASN member names.

ddavis-stsci (approved):
lgtm

PR #1214, opened 2024-05-02
Rcal 803 Initial resampling to skycell in the hlp

### LINKED ISSUE #1162: HLP add ability to resample to projection cells
_Issue [RCAL-803](https://jira.stsci.edu/browse/RCAL-803) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

If the given association contains a skycell in the product name search for the cell info and pass an external wcs to resample.

ddavis-stsci:
Resolves [RCAL-803](https://jira.stsci.edu/browse/RCAL-803)

Closes #1162

This PR adds the ability to use an association to resample the members to the output product based on a skycell definition.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)
The regression tests at https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/732/
are passing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ tweakreg
 general
 -------
 
+- Initial resample to a skycell in the hlp [#1214]
+
 - Add preview files to HLP tests [#1199]
 
 - Allow ``ModelContainer`` to work properly with context manager. [#1147]
--- a/romancal/pipeline/highlevel_pipeline.py
+++ b/romancal/pipeline/highlevel_pipeline.py
@@ -1,15 +1,30 @@
 #!/usr/bin/env python
 import logging
-from os.path import basename
+from os.path import basename, isfile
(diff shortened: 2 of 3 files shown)


schlafly (approved):
This looks good to me.

Some notes:
* I hadn't thought about this before, but it's a little ~interesting that we are doing outlier detection & sky matching using the full SCAs, but the projection using only the skycell area.  For the outlier detection that's wasteful, but probably not horribly so.  For the sky matching it produces different results, but probably different results in a mildly beneficial way.  I think we should just document that behavior for now.
* I agree with the code comments that we'll probably want to refactor this to get resample to take a WCS directly rather than via a file.
* I think I had the impression that you wanted to add some associated regtest demonstrating the L3 pipeline on skycells?

PR #1216, opened 2024-05-03
RCAL-798: Populate the l3 product individual_image_meta block

stscieisenhamer:
Resolves [RCAL-789](https://jira.stsci.edu/browse/RCAL-789)
Requires roman_datamodels [PR#348](https://github.com/spacetelescope/roman_datamodels/pull/348)

This PR addresses implements the filling of the MosaicImage `meta.individual_image_meta` block.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


stscieisenhamer:
I have left this as draft because:
- No units tests yet
- There are now a number of "fill the level 3 meta from level 2" functions. Methinks this could do with a refactor to consolidate, and possibly move into roman_datamodels, much of the functionality for readability/just better code. Leaving this to the opinion of the reviewers whether this refactor should be done and, if so, should it be done in the PR or another.

schlafly:
Thanks Jonathan.  Do you have an L3 file you can point me to that I can take a look at?

stscieisenhamer:
An example is found here

/grp/roman/eisenhamer/rcal-798-l3meta/roman_l3_metacheck.asdf

Note: This is saved from the test_l3_wcsinfo test because that test actually runs resample.

schlafly:
Okay, I see, the dictionary issue is from meta.ref_file.crds, which seems okay.  If we wanted to do that better we would do some kind of dictionary flattening first.  Presumably tests are failing because the RDM change isn't in yet?

stscieisenhamer:
Yes, the unknown dtype error is the error. Without the string-ing-fying, the dict is just an object which nothing likes. The ref_files.crds situation is the only one the currently exists, in which case flattening would be OK. But if more cases come up and flattening is attempted, one will need to ensure there is no key-name clashes.

schlafly:
Okay, let's hold this until @PaulHuwe can take a look at it and the corresponding RDM PR, and then merge after the RDM update goes in.

It would also be good to either run regtests or manually grab a couple cal files from artifactory and run the HLP on them with this PR, just in case the real exposure metadata is different enough from the filler metadata to cause issues.  I don't think I expect that, though.

stscieisenhamer:
[regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/373/)

stscieisenhamer:
[new regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/374/)

stscieisenhamer:
[Latest relevant regtest](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/375/). Differences found are expected.

Believe all is now clear and ready for final review.

schlafly (approved):
Looks good to me, merging.

PR #1222, opened 2024-05-06
rcal-827 Add regression test for skycell mosaic generation

ddavis-stsci:
Resolves [RCAL-827](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR adds a regression test for the L3 resampling of input image onto a skycell. 

**Checklist**
- [N/A] added entry in `CHANGES.rst` under the corresponding subsection
- [x] updated relevant tests
- [x] added relevant label(s)


schlafly (commented):
Can you also add the bits needed to construct L3_mosaic_asn.json and r0099101001001001001_F158_visit_r274dp63x31y81_i2d.asdf to the regtest script?  The latter just looks like strun --disable-crds-steppars roman_hlp L3_mosaic_asn.json.  Not sure if there's anything more to L3_mosaic_asn.json than asn_from_list, the three file names, and the output name?

ddavis-stsci:
I'm not sure why the test needs to add asn_from_list? 
In this case the L3_mosaic_asn.json is constructed to match a skycell. If the inputs are changed test will likely fail. 
So I'm not sure what adding the asn construction to the test gains us.

schlafly:
I just want to capture whatever you did in order to produce the L3_mosaic_asn.json.  If it's a fixed string, that's fine too, but let's put that string in the script.

schlafly:
I think the new version of this has an issue where you wanted to do WFI01 exposures 0001 through 0003, but you did WFI01 through WFI03 instead?

In any case, I went ahead and added the creation of these regtest files to the regtest creation script here
https://github.com/spacetelescope/romancal/pull/1228/files#diff-ab1578cd89f53a8e1bcabc27ab2bd67d6671c93b587d0236e33b1ffb81f7b0b4R189-R193
so I'm not worried about that part anymore.

ddavis-stsci:
Oops, Yes I did mean exposures.
Corrected, thanks

romancal/regtest/test_mos_pipeline.py
```diff
@@ -128,3 +129,61 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+    # r0000101001001001001_01101_0002_WFI01_cal.asdf \
+    # r0000101001001001001_01101_0003_WFI01_cal.asdf \
+    # -o L3_m1_asn.json"
+    # os.system(asn_cmd)  # nosec
+
+    for cal_file in cal_files:
+        rtdata.get_data(cal_file)
+
+    input_asn = "L3_mosaic_asn.json"
+    rtdata.get_data(f"WFI/image/{input_asn}")
```
braingram, line 166: This is overwriting the association you generated above with the skycell name and explains the regression test failure.

braingram: This was not the cause of the failure but is an issues and explains why the output file contains:
`r0099101001001001001_F158_prompt_visit_r274dp63x31y81`
instead of:
`r0099101001001001001_r274dp63x31y81_F158_prompt_visit`

romancal/regtest/test_mos_pipeline.py
```diff
@@ -128,3 +129,59 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+    _, serialized = asn.dump()
+    with open(asn_name, "w") as f:
+        f.write(serialized)
+
+    # asn_cmd = f"asn_from_list --product-name='r0099101001001001001_r274dp63x31y81_F158_prompt_visit'\
+    # r0000101001001001001_01101_0001_WFI01_cal.asdf \
+    # r0000101001001001001_01101_0002_WFI01_cal.asdf \
+    # r0000101001001001001_01101_0003_WFI01_cal.asdf \
+    # -o L3_m1_asn.json"
+    # os.system(asn_cmd)  # nosec
```
braingram, line 160:
```suggestion
```
This association is overwritten on line 166.

romancal/regtest/test_mos_pipeline.py
```diff
@@ -128,3 +129,59 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+    pipeline = MosaicPipeline()
+    diff = compare_asdf(rtdata.output, rtdata.truth, **ignore_asdf_paths)
+    assert diff.identical, diff.report()
+
+    model = rdm.open(rtdata.output, lazy_load=False)
+
+    pipeline.log.info(
+        "DMS356 MSG: Testing the creation of a Level 3 mosaic image"
+        + passfail(model.meta.cal_step.resample == "COMPLETE")
+    )
```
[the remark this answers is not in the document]
braingram: Also does this need to be `DMS373`?

romancal/regtest/test_mos_pipeline.py
```diff
@@ -128,3 +129,52 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+    ]
+
+    asn = asn_from_list(
+        cal_files, product_name="r0099101001001001001_r274dp63x31y81_F158_prompt_visit"
+    )
+    asn["target"] = "r274dp63x31y81"
+    asn_name = "L3_mosaic_asn.json"
+    _, serialized = asn.dump()
+    with open(asn_name, "w") as f:
+        f.write(serialized)
```
braingram, line 153:
```suggestion
```
The association is being saved to `L3_mosaic_asn.json` which is being overwritten on line 158.

braingram (changes requested):

PR #1226, opened 2024-05-06
pass exclude/ignore options to table meta comparison

braingram:
This PR allows passing `ignore` paths that include items in `astropy.Table` metadata by parsing the `exclude_paths` handled by `DeepDiff` and passing them through to the diff called on the table metadata. A test is added for the new feature.

@schlafly would you try this with the table date metadata issue you mentioned? Thanks!

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/regtest/regtestdata.py
+++ b/romancal/regtest/regtestdata.py
@@ -609,11 +609,20 @@ class TableOperator(NDArrayTypeOperator):
         difference = self._compare_arrays(a, b)
         # also compare meta  for tables
         if level.t1.meta != level.t2.meta:
+            kwargs = {}
+            if diff_instance.exclude_paths:
+                meta_path = level.path() + "['meta']"
+                kwargs["exclude_paths"] = [
+                    f"root{path.removeprefix(meta_path)}"
+                    for path in diff_instance.exclude_paths
+                    if path.startswith(meta_path)
+                ]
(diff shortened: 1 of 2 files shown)


schlafly:
Yes.  I am regenerating the regtest files right now, though, so it will be a bit.  Here was the approach I took:
https://github.com/schlafly/romancal/blob/ba1e79d2893602a9f52ca398fa4d8d5ed7efcf32/romancal/regtest/regtestdata.py#L617
Just making sure I follow, my version does the one dumb necessary thing, and your version takes the whole exclude_paths and copies them so they apply to the Table metadata as well.  In particular, roman.meta.date is in there, and we hope that handles meta.date here as well?

braingram:
Thanks! No rush from my end. I gave this a look (as a reprieve from hunting down memory usage) and saw what might be a relatively general approach.

For this PR let's say there is a `astropy.Table` for the `source_catalog` attribute of a datamodel. If that table contains a `meta.date` that you want `compare_asdf` to ignore you could modify `ignore_asdf_paths` to include:
```
source_catalog.meta.date
```
and it would ignore specifically the `date` metadata entry for the `source_catalog` table.

braingram:
This can likely be "revived" if the changes are wanted. Since it currently has conflicts I'm closing it until that point.

PR #1228, opened 2024-05-07
Regtest / integration fixes for B14

schlafly:
This PR makes a handful of minor changes in order to get things integrated for B14.  In combination with https://github.com/spacetelescope/roman_datamodels/pull/349 and the new regtest files in /grp/roman/eschlafly/romanisim/20240503 we pass tests, etc..  Some bits of this PR:

* Update context to 58.
* Ignore individual_image_metadata in compare_asdf because compare_asdf doesn't currently handle non-numeric array types (i.e., it barfs when told to compare two numpy arrays of strings)
* Ignore the 'date' meta keyword in astropy tables, since this keyword gets updated with the table creation date in the PSF catalog module.
* Update the file names of the L3 regtest files without skycells.
* Update the regtest creation script to have the right file names.


ddavis-stsci (approved):
LGTM

PR #1233, opened 2024-05-10
Fix bug when setting meta.background in SkyMatchStep.

mairanteodoro:
Resolves [RCAL-837](https://jira.stsci.edu/browse/RCAL-837)

This PR addresses a bug in `SkyMatchStep` that was causing meta.background.subtracted to always be set to `None` instead of `True`/`False`, which was preventing `ResampleStep` from properly using the sky level as determined by `SkyMatchStep` and set in `meta.background.level`.

**Regression tests**

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


mairanteodoro:
Attaching comparison of the relevant arrays between the output from the regression test against the truth file from Artifactory. Black and white circles indicate the position of the worst absolute and the worst fractional differences, respectively, as reported by the regtest (ran locally, which, by the way, differs from the coordinates reported when ran on Jenkins).

- `data` array:
  -
output_truth_comparison

- `err` array:
  - 
output_truth_err_comparison

- `var_flat` array:
  - 
output_truth_var_flat_comparison

- `var_poisson` array:
  - 
output_truth_var_poisson_comparison

- `var_rnoise` array:
  - 
output_truth_var_rnoise_comparison

- `weight` array:
  - 
output_truth_weight_comparison

schlafly (approved):
Looks good.  My rule is that I use "is" when I am comparing objects that may be _the same object_, and == when I am comparing values.  In this context I was surprised by using is to compare bools on 336 but otherwise looks good.

mairanteodoro:
> Looks good. My rule is that I use "is" when I am comparing objects that may be _the same object_, and == when I am comparing values. In this context I was surprised by using is to compare bools on 336 but otherwise looks good.

Thanks for catching that overlook of mine. ;)

nden:
@mairanteodoro This is not attached to a milestone. Was this work included in the last release?

mairanteodoro:
> @mairanteodoro This is not attached to a milestone. Was this work included in the last release?

No... it's not even in the `main` branch. I don't know what happened here...

nden:
@mairanteodoro OK, I added a milestone.

PR #1241, opened 2024-05-17
Replace ModelContainer with ModelLibrary

braingram:
This PR replaces `ModelContainer` with a `ModelLibrary` subclass.

Please see stpipe for the core documentation on `ModelLibrary`.

The main benefit is that `ModelLibrary` (in contrast to `ModelContainer`) will allow passing data between steps without holding it in memory (and instead keeping datamodels "on disk").

This will allow for future improvements to steps (especially those part of a pipeline) to allow them to run in a generically memory efficient way (the interface to an "in memory" and an "on disk" `ModelLibrary` are identical). Importantly this PR does not aim to make improvements to the steps (to limit the already wide scope of the changes). Follow-up PRs will be needed to address issues with some steps where they are coded in a way that keeps all models in memory (defeating the memory saving possible with `ModelLibrary`). To provide one example, the tweakreg step contains a call to [create_astrometric_catalog](https://github.com/spacetelescope/romancal/blob/459e8e345bbb1ba6bfd76d0d8a90f2058089791e/romancal/tweakreg/astrometric_utils.py#L29) which currently expects a list of models. In this PR the models are read from the library, which loads all models in memory. It is possible to change the code avoid reading in all of the models. However this would require changes to several utility functions (some in stcal) which are technically not required to add the `ModelLibrary` but are required to see performance benefits from using the library.

Regression tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365
shows 5 minor metadata differences.
- [test_resample_single_file](https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365/job/27526845392#step:29:3070) where `exptype` and `group_id` are now in the output file
- [test_level3_mos_pipeline](https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365/job/27526845392#step:29:2048) where `asn.pool_name` is now `none` instead of `dummy value` and `asn.table_name` is now `L3_regtest_asn.json` instead of `dummy value`
- [test_tweakreg](https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365/job/27526845392#step:29:1802) where `asn` and `exptype` are now in the output file
- [test_level2_grism_processing_pipeline](https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365/job/27526845392#step:29:899) where `asn` `exptype` and `group_id` are now in the output file
- [test_level2_image_processing_pipeline](https://github.com/spacetelescope/RegressionTests/actions/runs/9962595365/job/27526845392#step:29:214) where `asn` and `exptype` are now in the output file

Fixes https://github.com/spacetelescope/romancal/issues/1303

Todo:
- [x] address unit test differences
- [ ] address regtest difference
- [x] update documentation
- [x] port over jwst model library tests
- [x] add support for `__init__` from in memory association
- [x] add "index" API for cross-model attribute access (perhaps a `map_function` method)
- [x] finalize naming for `__setitem__` `__getitem__` `discard`
- [x] update minimum requirements
- [x] finalize stpipe changes (more may be required based on `__setitem__` changes)
- [x] add `ModelLibrary` docs

Closes #

This PR addresses ...

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


perrygreenfield:
I find the use of `__getitem__` and `__setitem__` somewhat nonintuitive when using the library theme. Wouldn't either methods such as `borrow` or `checkout`  be more appropriate for the first, and `return` for the second?

perrygreenfield:
Perhaps `return_model()`.  But `unborrow` or `shelve` would also work. We can leave the `__getitem__ `interface in, but not use it explicitly in most of the code.

braingram:
I think `borrow_model` `return_model` and `discard_model` might be a cohesive set. I will look at what would be required in `stpipe` to drop the `Sequence` check.

braingram:
@mairanteodoro when you have a chance would you take a look at this PR. It replaces all uses of `ModelContainer` with a new `ModelLibrary`. I tried to keep other changes to a minimum but there were some step changes. There are remaining things to do before opening this for review including:
- updating documentation
- porting the [jwst tests](https://github.com/braingram/jwst/blob/model_library/jwst/datamodels/tests/test_library.py) to roman
- finalizing the required updates to stpipe
- updating the minimum requirements (the library uses some newer asdf api)

At the moment the unit tests are passing except for a few I skipped:
- several tests run steps with invalid inputs (`None`, `''`, etc) some of the steps produce errors, other produce warning log messages, I didn't see a consistent theme so figuring out how to update these will require some discussion
- [test_tweakreg_multiple_groups_valueerror](https://github.com/spacetelescope/romancal/blob/6c3e5632ec0d46cfa8f9e584288a2bc9b55609c7/romancal/tweakreg/tests/test_tweakreg.py#L1084) checks that the step is skipped for what looks like valid inputs. I think the changes in this PR may have fixed a bug with the `group_id` where on `main` it's [not being passed](https://github.com/spacetelescope/romancal/blob/6c3e5632ec0d46cfa8f9e584288a2bc9b55609c7/romancal/tweakreg/tweakreg_step.py#L621) to `tweakwcs`
- [test_outlier_do_detection_write_files_to_custom_location](https://github.com/spacetelescope/romancal/blob/6c3e5632ec0d46cfa8f9e584288a2bc9b55609c7/romancal/outlier_detection/tests/test_outlier_detection.py#L143) checks that the median file exists after a run of outlier detection (which used `save_intermediate_results=False`) it's my understanding that the step should delete the median file for this configuration. If so, this test and the step have bugs.

I also noticed an issue with outlier detection noted [here](https://github.com/spacetelescope/romancal/pull/1146#issuecomment-2118131737) where it's using `single=False` which produces a single combined resampled image for all input images that is then used as the input to the median calculation (which takes the median over an N=1). Changing this would change the regression tests results so I didn't update that in this PR.

The regression tests are running here: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-Developers-Pull-Requests/detail/Roman-Developers-Pull-Requests/789/pipeline/247 (I'll update this comment when they finish).

I expect the hlp regtest will fail (as it's failing locally) with a minor file difference where the current truth file does not contain the `asn_pool_name` or `asn_table_name` from the input association. With this PR the values are carried through the pipeline so I believe they are the intended result. I do have questions on this and a recent change to `ModelContainer` in https://github.com/spacetelescope/romancal/pull/1209. I don't see any tests for the feature added in that PR so I'm not sure if the changes in this PR have any regressions with respect to that feature.

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+            model = self._load_member(index)
+            if not self._on_disk:
+                # it's ok to keep this in memory since _on_disk is False
+                self._model_store[index] = model
+
+        # track the model is "in use"
+        self._ledger[index] = model
+        return model
+
+    def __setitem__(self, index, model):
```
emolter, line 303: Does the setter method remove the model from memory after saving to the temporary file?  I'm relatively new to looking at this type of code so I'm probably missing it, but I don't see where that happens here or in `self._model_store`

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+
+        # un-track this model
+        del self._ledger[index]
+
+        # and store it
+        self._model_store[index] = model
+
+        # TODO should we allow this to change group_id for the member?
+
+    def discard(self, index, model):
```
emolter, line 315: It seems to me it would be nice to be able to do both of the following:
```
with library:
    library.discard(0)
    library[0].discard()
```
Are both of these supported?

Also, I would expect discarding a model to remove it from the library entirely, including the temporary file where it is stored.  Does that happen?

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+        if copy:
+            copy_func = lambda value: value.copy()  # noqa: E731
+        else:
+            copy_func = lambda value: value  # noqa: E731
+        with self:
+            for i, model in range(len(self)):
+                attr = model[attribute]
+                self.discard(i, model)
+                yield copy_func(attr)
+
```
emolter, line 471: Random note, but IMO it would be nice to have a convenience function for splitting a ModelLibrary by exposure type here, i.e., a more general version of [this function](https://github.com/spacetelescope/jwst/blob/master/jwst/master_background/master_background_step.py#L239) in JWST's MasterBackground step.  I think slightly different versions of this already exist in multiple places in the JWST code base, and this seems like a good opportunity to make it generic.

romancal/datamodels/tests/test_datamodels.py
emolter, line 1: Where are the unit tests for the ModelLibrary?  Is there a good reason to remove this entirely instead of replacing it?

emolter (commented):
Most of this "review" is just me asking clarification questions about how this works; thanks for humoring that.  Not sure I have the expertise yet to add much more than that.

romancal/datamodels/library.py
braingram: Good point. I had similar thoughts about the `discard` naming. In this case `discard` means to return a model to the library but not write it out to a (possibly new) temporary file. The model will still be in the library and can be later "borrowed" again). The main use is when you only need to read something from a model (like the wcs) but don't want to incur the cost of writing out the model when it's returned to the library. Do you think it would be clearer if this modified vs un-modified "return" was an argument to `return_model` (or `shelve`)? Perhaps something like:
```python
with library:
    model = library.borrow(0)  # borrow the 0th model
    # do some read-only stuff with the model
    library.shelve(0, model, modified=False)   # with modified=True being the default
```
To address the other part of your comment supporting `library.discard(0)` sounds like a good idea and perhaps `library.discard(model)`. I can't currently think of way of supporting `library[0].discard` without patching a `discard` method onto the model. I briefly considered this but I worry the added complexity would be hard to get ride and prone to bugs.

romancal/datamodels/library.py
braingram: For a "in memory" library (when `_on_disk=False`) references to the models are held by the `_model_store`. They won't be candidates for garbage collection until the library is collected. For an "on disk" library the store doesn't hold any references to the models. In the "on disk" mode the models will be removed from memory only after the step removes all references. One of the goals of the "borrow"/"shelve" API is to encourage step developers to be aware of when the step is using a model. It however does not prevent the step from keeping references to the models and preventing them from being freed.

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+        self._asn_n_members = asn_n_members
+        self._on_disk = on_disk
+
+        self._open = False
+        self._ledger = {}
+
+        # FIXME is there a cleaner way to pass these along to datamodels.open?
+        self._memmap = memmap
+
+        if self._on_disk:
```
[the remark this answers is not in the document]
braingram: Yes!

It largely covers how `ModelContainer` is often used and for small associations is almost certainly more efficient. I believe @nden mentioned the idea of dynamically setting (or unsetting) this based on the size of the association which I think we should consider. I think for most uses we'll want to have `on_disk=False` and only enable it if we expect (or know) the step to run into memory issues. In that mode we could define a threshold where if `on_disk=True` and the library contains less than N MB of files then `on_disk` is turned off and a warning or log message is reported. An alternative would be to have it be default `on_disk=True` and use a similar threshold.

The current prototype defaults `on_disk=False` and exposes it via a pipeline parameter for [hlp/mosaic](https://github.com/spacetelescope/romancal/pull/1241/files#diff-15cbca0fd660d7d5eac167ece3633d2a1a665d776bf180b07617f06b5fb40a7bR42). I think this is the most predictable mode but does mean a user (or processing) would need to determine the best value for this setting at run time.

romancal/datamodels/library.py
emolter: Is there a straightforward way to check if the model has been modified when it is shelved, and only incur that cost if so?  My worry would be that a random developer or user would be unaware of the modified keyword (e.g., after having copied a random example from elsewhere in the code base) and just leave it as default, even though they do not intend to modify the model.

emolter: Also, I see your point about `library[0].discard` and I think your proposed `library.discard(model)` would be a suitable replacement

braingram: I'm not aware of a straightforward way to check if a model has been modified. The risk you mention about copying an example is a good point. For catching that type of error I think we can rely on the code reviews and tests. Depending on how we expose "on_disk" (what it will default to and if it will be dynamic) we may want to unit test every step that uses the library for both `on_disk=True` and `on_disk=False` (or specifically disallow one mode). This might be an argument for making `on_disk=False` the default and only setting it by a pipeline/step parameter (and making sure that the PR that enables it tests the code in both modes).

ddavis-stsci (commented):
This also seems to cause many issues with the unit tests,
test_outlier_detection.py,
test_skymatch and,
test_tweakreg
Do we have an idea of how much work is required to correct these? 

The items in the library are also read only? 
```
ml.asn['products'][0]['members'][1]['exptype'] = 'background'
  File "<stdin>", line 1, in <module>
TypeError: 'mappingproxy' object does not support item assignment
```
Can this be relaxed? For instance it might be convenient to add catalogs to the association for SDF processing.

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+    # pretty easily with the actual models.
+    # @property
+    # def members(self):
+    #     return self.asn['products'][0]['members']
+
+    @property
+    def group_names(self):
+        names = set()
+        for member in self._members:
+            names.add(member["group_id"])
```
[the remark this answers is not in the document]
braingram: Thanks for pointing this out. I missed that this is different than jwst (which allows setting group id in the association). I believe this was added in part to allow users to define grouping for tweakreg, skymatch, resample (any step that uses group id). It's more efficient to have this group id in the association instead of loading each model to compute the group id (which is why the group id is now being added to the [level 3 association generation for jwst](https://github.com/spacetelescope/jwst/issues/8523)).

The format here should match the format defined by the [current model container](https://github.com/spacetelescope/romancal/blob/ed6187ffa790fc6a37dd96f7365ba1ef8261fa74/romancal/datamodels/container.py#L499) and used by tweakreg, skymatch and resample in romancal. I can remove the part of this code that allows the group id to be defined from the association if roman isn't planning on supporting that feature. However the library does need to define the group id for the mentioned steps to work. Is there something incorrect in the format? If so would you provide an example?

romancal/datamodels/tests/test_datamodels.py
braingram: I ported over the unit tests in: 3b8716c1368c1370bb7920e06de2c9f1e5e2c616

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+    """
+
+    pass
+
+
+class _OnDiskModelStore(MutableMapping):
+    def __init__(self, memmap=False, directory=None):
+        self._memmap = memmap
+        if directory is None:
+            # when tem
```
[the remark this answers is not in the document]
braingram: Removed in 65955b2d3d69ecd3e53df14933f5e7fc24eb7bc3

romancal/datamodels/library.py
```diff
@@ -0,0 +1,504 @@
...
+            self._tempdir = tempfile.TemporaryDirectory(dir="")
+            # TODO should I make this a path?
```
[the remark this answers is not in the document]
braingram: Switched to a path in 65955b2d3d69ecd3e53df14933f5e7fc24eb7bc3

ddavis-stsci:
I'm not aware of any plans to add group_id to the Roman associations so 
I don't think that this is the
correct place for it in the model library. If some steps need it we can 
discuss adding it but I don't
see any use cases for Roman.

ddavis-stsci:
On 5/31/24 9:37 AM, Brett Graham wrote:
>
>     This also seems to cause many issues with the unit tests,
>     test_outlier_detection.py, test_skymatch and, test_tweakreg Do we
>     have an idea of how much work is required to correct these?
>
> Thanks for giving this a look. Did you pip install the package before 
> testing? I ask because the changes in this PR require a minor 
> modification to stpipe (see the modified pyproject.toml which points 
> to the stpipe fork). Without those changes many tests will fail as 
> stpipe will attempt to treat the |ModelLibrary| as a |ModelContainer|.
>
No, I just checked out your branch. I'll give that a try for the next 
tests.

>     The items in the library are also read only?
>
>     |ml.asn['products'][0]['members'][1]['exptype'] = 'background'
>     Traceback (most recent call last): File "<stdin>", line 1, in
>     <module> TypeError: 'mappingproxy' object does not support item
>     assignment |
>
>     Can this be relaxed? For instance it might be convenient to add
>     catalogs to the association for SDF processing.
>
> The models are modifiable but the association metadata is read only. I 
> made this read only because modifying the association after models 
> have been loaded (or after the library is created when |asn_exptypes| 
> is used) leads to mismatches between the models and the association. 
> Using the example, let's assume member 1 has a exptype of science. If 
> the model is loaded from the library (or from the existing container), 
> it will contain the 'science' exptype in it's metadata (at 
> |model.meta.asn.exptype|). If now the exptype is changed in the 
> association there is no way to update the corresponding model metadata 
> and the asn will report this as the updated value but the model 
> metadata will still report 'science'. Making the asn read only helps 
> to prevent this (although it's still possible to modify the
>
Here I am thinking ahead that we'll need difference images (color maps) 
and may need to designate a science exposure as a background to be 
subtracted from the science image.
We don't have any algorithms yet but just trying to be flexible.
>
> model metadata and have it not match the asn).
>
> Would you expand on the use case for adding catalogs to the 
> association? Would it work to add these to the association before the 
> library is created? I don't think I've yet added creating a library 
> from an in memory association but that should be straightforward and 
> might help here.
>
I think for the production products we can create the catalog names that 
are associated with an image file, thus spending hours discussing file 
names.
For custom catalogs,
strun romancal.step.TweakRegStep --catfile my_favorite_cat.asdf
it would be good to be able to add / replace the catalog file in the asn.
We can probably ignore this for now and focus on the production products.

braingram:
> I'm not aware of any plans to add group_id to the Roman associations so I don't think that this is the correct place for it in the model library. If some steps need it we can discuss adding it but I don't see any use cases for Roman.

Thanks for the reply. I think this is something that should be considered as it can provide dramatic performance improvements for grouping (one of the primary features of the container). What's the right venue for discussing this?

schlafly:
I think conceptually "group" = "exposure" for Roman.  i.e., for tweakreg, we want to group all of the exposures together and compute a common astrometric shift for all SCAs.  I'm not actually aware of what "group" is controlling for sky matching or outlier detection; since multiple detectors in the same exposure cannot overlap, I can't figure out really anything that grouping can do in outlier detection.  For sky matching, I can imagine it being used as a fast proxy for determining that it isn't necessary to check to see if WFI01 and WFI02 in the same exposure overlap.  I could also imagine some mode that fits a common sky offset for groups rather than SCAs; I believe that we fit one offset per SCA.  We don't have more complicated groups than that.  Likewise for resampling I'm missing something about what changing the grouping can do.  For the ELP (i.e., tweakreg), we always have exactly one group; all the SCAs get tweakreg'd together.

ddavis-stsci:
Can you point me to some of the JWST docs for this?
We should probably bring it up a Thursday romancal tag-up meeting.

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -54,66 +54,94 @@ def process(self, input_models):
...
+                    )
+
+        # if skipping for any reason above...
+        if self.skip:
+            # set meta.cal_step.outlier_detection to SKIPPED
+            with library:
+                for i, model in enumerate(library):
+                    model.meta.cal_step["outlier_detection"] = "SKIPPED"
+                    library[i] = model
+            return library
```
schlafly, line 99: Good to duplicate the existing behavior, but FWIW, I think we'd be better off crashing or throwing an exception or ... in the case that someone sends something that's not an image.  Skipping feels more likely to cause problems than crashing does.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -187,10 +185,8 @@ def process(self, input):
         #          observations. This should not occur on-prem
         if result.meta.exposure.type == "WFI_IMAGE":
             if file_type == "asdf":
+                # FIXME run tweakreg?
                 result.meta.cal_step.tweakreg = "SKIPPED"
```
schlafly, line 189: FWIW here, I'm a bit confused by setting the step as "skipped" after self.tweakreg(tweakreg_input) below, but I suspect what's going on here is that we usually want to run tweakreg with the full 18 detectors rather than an individual file.  I actually haven't seen the SDP code that controls that.

@ddavis-stsci , can you clarify the logic here?  Thanks!

romancal/regtest/test_tweakreg.py
```diff
@@ -67,12 +67,12 @@ def test_tweakreg(rtdata, ignore_asdf_paths, tmp_path):
...
-    pts = np.linspace(0, 4000, 30)
-    xx, yy = np.meshgrid(pts, pts)
-    coordtweak = wcstweak.pixel_to_world(xx, yy)
-    coordtrue = wcstrue.pixel_to_world(xx, yy)
+    with asdf.open(orig_uncal) as orig_model_asdf:
+        wcstrue = orig_model_asdf["romanisim"]["wcs"]  # simulated, true WCS
+        pts = np.linspace(0, 4000, 30)
+        xx, yy = np.meshgrid(pts, pts)
+        coordtweak = wcstweak.pixel_to_world(xx, yy)
+        coordtrue = wcstrue.pixel_to_world(xx, yy)
```
schlafly, line 75: Just making sure, this is unrelated to the PR but will lead files to clean up better?

romancal/resample/resample.py
```diff
@@ -299,41 +307,44 @@ def resample_many_to_one(self):
...
+                del data, inwht
+                members.append(str(img.meta.filename))
+                self.input_models.discard(i, img)
+
+        # FIXME: what are filepaths here?
+        # members = (
+        #     members
+        #     if self.input_models.filepaths is None
+        #     else self.input_models.filepaths
+        # )
```
schlafly, line 347: This is my fault.  There was a request to store the actual file names of the L2 input files (as opposed to their meta.filenames, which may become disconnected).  filepaths got populated with the file names from the association file in the model container here: https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/datamodels/container.py#L184

romancal/resample/resample.py
```diff
@@ -385,63 +398,65 @@ def resample_variance_array(self, name, output_model):
...
+                        f"No data for '{name}' for model "
+                        f"{repr(model.meta.filename)}. Skipping ..."
+                    )
+                    continue
+                elif variance.shape != model.data.shape:
+                    log.warning(
+                        f"Data shape mismatch for '{name}' for model "
+                        f"{repr(model.meta.filename)}. Skipping..."
+                    )
+                    continue
```
schlafly, line 415: For what it's worth, these could equally well be failures & exceptions.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -103,52 +100,39 @@ def process(self, input):
...
+            with images:
+                for i, member in enumerate(images.asn["products"][0]["members"]):
+                    filename = member["expname"]
+                    if filename in catdict:
+                        # FIXME: I'm not sure if this captures all the possible combinations
+                        # for example, meta.tweakreg_catalog is set by the container (when
+                        # it's present in the association). However the code in this step
+                        # checks meta.source_catalog.tweakreg_catalog. I think this means
+                        # that setting a catalog via an association does not work. Is this
+                        # intended? If so, the container can be updated to not support that.
```
schlafly, line 129: This will change a meaningful amount when we stop storing the source detection catalog in the image metadata.  When that happens, tweakreg will need to get a list of images and associated catalogs.  I suspect that means that a lot of this logic could be simplified (e.g., all catalogs will be specified by a file name, and tweakreg doesn't need special logic to separate user catalogs from normal source catalogs).

romancal/tweakreg/tweakreg_step.py
```diff
@@ -167,166 +151,174 @@ def process(self, input):
...
-                    # read catalog from structured array
-                    catalog = Table(
-                        np.asarray(image_model.meta.source_detection.tweakreg_catalog)
+        with images:
+            for i, image_model in enumerate(images):
+                if image_model.meta.exposure.type != "WFI_IMAGE":
+                    # Check to see if attempt to run tweakreg on non-Image data
+                    self.log.info("Skipping TweakReg for spectral exposure.")
+                    # Uncomment below once rad & input data have the cal_step tweakreg
+                    # image_model.meta.cal_step.tweakreg = "SKIPPED"
```
schlafly, line 160: @mairanteodoro , this can probably be uncommented now?

romancal/tweakreg/tweakreg_step.py
```diff
@@ -167,166 +151,174 @@ def process(self, input):
...
-        if len(grp_img) == 1 and not ALIGN_TO_ABS_REFCAT:
-            self.log.info("* Images in GROUP 1:")
-            for im in grp_img[0]:
-                self.log.info(f"     {im.meta.filename}")
-            self.log.info("")
+        if len(group_indices) == 1 and not ALIGN_TO_ABS_REFCAT:
+            # self.log.info("* Images in GROUP 1:")
+            # for im in grp_img[0]:
+            #     self.log.info(f"     {im.meta.filename}")
+            # self.log.info("")
```
schlafly, line 267: I presume that this and related logging issues are a temporary expedient rather than a proposed change?

romancal/datamodels/library.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def discard(self, index, model):
+        # TODO it might be worth allowing `discard(model)` by adding
+        # an index of {id(model): index} to the ledger to look up the index
+        if index not in self._ledger:
+            raise BorrowError("Attempt to discard non-borrowed model")
+
+        # un-track this model
+        del self._ledger[index]
+        # but do not store it
```
schlafly, line 337: Maybe just repeating Ned's comment, discarding is a pretty common case in your code and currently you always do discard(i, model).  But I think at present discard(i, model) doesn't look at model?  In that case I would definitely make discard(...) take either i or model but not both.  The current syntax kind of makes you use:
```
model = library[i]
...
library.discard(model, i)
```
instead of
```
function(library[i])
library.discard(i)
```

It also made me ask myself "what is it doing with that model I just gave back to it that I thought I wanted to throw away?!"

schlafly (commented):
Lot of work here!  I had a few comments.  I still need to look at the main library code.

romancal/datamodels/library.py
braingram: "throw away" might be another name to consider :)

I am starting to lean towards `shelve` (with a `modify` argument that defaults to `True`) this would replace `discard` with something like `library.shelve(model, modify=False)`. One potential downside to this is that internally the library operates on the index of the model within the container (so that references to the model don't need to be kept). Supporting either `discard(model)` or `shelve(model)` would require keeping some mapping of model to index. This could be done with weak references but that would require a more complex library. We could instead keep references the model and accept that between a `borrow` and a `shelve` even if the step code frees all references the model would stay in memory because of the library.

romancal/regtest/test_tweakreg.py
braingram: Exactly! I can certainly drop this or put it in a separate PR. i likely just made the change as I was trying to understand the code.

romancal/resample/resample.py
braingram: Thanks for commenting on this. I don't yet fully understand how these are used or carried through the pipelines. Some discussion on expectations for this would be helpful. For example, when running the `mosaic` pipeline is the expectation that these will be the original input filenames? If so, I think this  code in skymatch will lose the filepaths:
https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/skymatch/skymatch_step.py#L100

romancal/tweakreg/tweakreg_step.py
braingram: Thanks! That's very helpful to know. I am curious to hear more as the ability to set the catalog from the association, the model metadata and the "catfile" has led to [some confusion](https://github.com/spacetelescope/jwst/pull/8424#discussion_r1571302034) (and bugs) in jwst.

romancal/tweakreg/tweakreg_step.py
braingram: This is a partially proposed change at the moment. It does highlight one pattern that has some negative consequences when the `ModelLibrary` is `on_disk`. The logging commented out here (if the library is `on_disk`) will result in every model being read from disk to print out the `meta.filename` of each model. This is shortly followed by every model then being re-opened to set the `cal_step` status. I think a more complete change would be to combine these loops or remove the logging if printing out the filenames isn't that important (I don't think most steps do this).

romancal/resample/resample_step.py
```diff
@@ -68,43 +68,49 @@ class ResampleStep(RomanStep):
...
+            output = f"{os.path.commonprefix([x['expname'] for x in input_models.asn['products'][0]['members']])}.asdf"
+            if len(output) == 0:  # FIXME won't this always at least be ".asdf"?
                 # set default filename if no common prefix can be determined
                 output = "resample_output.asdf"
         else:
             raise TypeError(
-                "Input must be an ASN filename, a ModelContainer, "
+                "Input must be an ASN filename, a ModelLibrary, "
                 "a single ASDF filename, or a single Roman DataModel."
             )
```
mairanteodoro, line 104: This is another example that would benefit from letting `ModelLibrary` handle the input type.

romancal/skymatch/skymatch_step.py
```diff
@@ -52,11 +51,12 @@ class SkyMatchStep(RomanStep):
 
     def process(self, input):
         self.log.setLevel(logging.DEBUG)
-        self._is_asn = False
+        self._is_asn = False  # FIXME: where is this used?
```
mairanteodoro, line 54: I don't think we need that for now. Maybe it was a placeholder variable that was intended to be used as an input parameter? Regardless of the initial intent, I think it is safe to remove it.

mairanteodoro (commented):
Looks good to me. I left a few comments/suggestions.

romancal/flux/flux_step.py
```diff
@@ -48,33 +48,38 @@ class FluxStep(RomanStep):
...
-        elif isinstance(input, ModelContainer):
+        elif isinstance(input, ModelLibrary):
             input_models = input
             single_model = False
         else:
             raise TypeError(
-                "Input must be an ASN filename, a ModelContainer, "
+                "Input must be an ASN filename, a ModelLibrary, "
                 "a single ASDF filename, or a single Roman DataModel."
             )
```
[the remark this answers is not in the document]
braingram: Thanks! I think that's a good idea. I wonder if adding some of this logic to [RomanStep._datamodels_open](https://github.com/spacetelescope/romancal/blob/f593a51435bab3719849a248bf50e25db1ec5dc6/romancal/stpipe/core.py#L36) might help here (and other places). That classmethod is called by stpipe on the step input when attempting to resolve the config, reference files, etc. If it accepted:
- string -> if '.asdf' open as DataModel if '.json' (or '.yaml') open as ModelLibrary
- DataModel -> return DataModel
- ModelLibrary -> return ModelLibrary
then I think roman would be able to use `pars-` files, prefetch references, and any other stpipe infrastructure that uses `_datamodels_open`. I think we would still need some logic in this step (or in ModelLibrary) as we want a ModelLibrary even if we have a DataModel. It might boil down to something like:
```python
def process(input):
    opened_input = self._datamodels_open(input)
    if isinstance(opened_input, DataModel):
        input_models = ModelLibrary([opened_input])
    else:
        input_models = opened_input
```
One other issue to consider is how modification of the input is treated. With jwst there is a general "rule-of-thumb" that the input is not modified. With the above psuedocode if input is a model a copy would be need to be made to avoid modifying the input. Is there a similar convention for roman?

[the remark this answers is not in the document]
mairanteodoro: > One other issue to consider is how modification of the input is treated. With jwst there is a general "rule-of-thumb" that the input is not modified. With the above psuedocode if input is a model a copy would be need to be made to avoid modifying the input. Is there a similar convention for roman?

I'm not 100% sure, but I don't think we have that for Roman.

braingram:
To recap some of the discussion from tag-up today.

"borrow" API
-----
The winner is...
```python
with lib:
    model = lib.borrow(0)
    ...
    lib.shelve(model)
```
The replacement for `discard` is:
```python
lib.shelve(model, modify=False)  # modify defaults to True
```
and for uses where a model is replaced (with a new model)
```python
lib.shelve(model, index)
```
`index` will default to `None` (or some other "unset" value) and the library will maintain a model to index mapping. This mapping will mean that between a `borrow` and a `shelve` the library will retain a reference to the "borrowed" model preventing it from being garbage collected (even if the step removes all references). So for the following example:
```python
with lib:
    model = lib.borrow(0)
    new_model = model.copy()
    del model  # because of the internal model to index mapping this isn't sufficient to free the memory for model
    lib.shelve(new_model, 0)
```
This seems acceptable given the reduced API complexity (`shelve` does not need to always receive a model and an index) and if it becomes a performance bottleneck (which seems unlikely) the internal mapping could be updated to use `weakref` (at the expense of increased complexity).

expanding the `ModelLibrary` (adding a wing?)
----
There is currently no way to "append" a model to the library. Appending is possible with `ModelContainer` however when examining the usage in jwst and romancal in (almost) all cases using a list of models is equivalent and simpler. The one exception is outlier detections use of [resample](https://github.com/spacetelescope/romancal/blob/f593a51435bab3719849a248bf50e25db1ec5dc6/romancal/resample/resample.py#L188) where the `drizzled_models` are accumulated in a `ModelContainer`. However in this instance a list of filenames can work (see https://github.com/spacetelescope/romancal/pull/1260 for some more context on this code in romancal).

It is possible to add a "append" method to the `ModelLibrary`. However this would complicate bookkeeping for:
- model index to temporary file relationship
- association member list data and model metadata

As the current uses can be replaced with lists, I suggest we don't implement "append" and instead think of the `ModelLibrary` as a way to read models from an association and pass them through the pipeline. Some items we discussed today was a possibility of initializing a `ModelLibrary` with "blank" models (to be filled by the pipeline) and using the `ModelLibrary` as a way to organize members in an association (assigning them different `exptypes` and/or adding new models). These operations are currently possible only by turning the library into a list of models then back into a library (although neither the library nor `ModelContainer` allow saving the association). Some thought would have to go into how to implement these features in the library (or if they're more suited to other structures either existing or to-be-made).

tracking files used by the pipeline
----
The `filepaths` in `ModelContainer` are used to track which input files are used to generate an output file. This is an issue that has also been raised for jwst and some thought could be given to how `ModelLibrary` might play a role in this tracking. It may make sense (or be required) to have steps update this information but I have not given the issue much thought at this point.

`ModelLibrary` in stpipe
----
I am currently considering `stpipe` as the "home" for `ModelLibrary`. This is a dependency for both romancal and jwst which could implement sub-classes that:
- define the `group_id` determination
- add `filepaths` support (for roman)
- inform the library of any specifics of the datamodel differences

Given the ease in porting the core library from jwst to romancal (few internal library changes were needed, it was mostly updating cal code) I expect the main difficulty with moving the library to `stpipe` will be:
- coordinating releases, versions etc (ie fixing a bug in stpipe may require setting the dev version as a requirement) this is already an issue but may be more prominant as the library develops
- to test the library in stpipe we may need to make a non-abstract DataModel for that package. This may have the positive side-effect of increasing test coverage for that feature in stpipe.

romancal/skymatch/skymatch_step.py
braingram: Thanks! Removed in: c1089888393ebf83604fa0bac42b8f3e87fbae82

romancal/resample/resample_step.py
braingram: Looking at this again I have questions about the refactoring here and https://github.com/spacetelescope/romancal/pull/1241#discussion_r1629493747

The tldr; is that I don't think further refactoring is needed for `ModelLibrary` and the cleanup of the input handling could be in a separate PR.

The flux step input is relatively straightforward. However resample is complicated by a few things:
- `blendeheaders` is only set if the input is `DataModel` (and not if the input a filename for a single datamodel)
- the `output` for a `ModelContainer` is set based on the common filename whereas the `output` for a `ModelContainer` read from an association file is based on the product name

Any refactoring of the input handling will either need to replicate these differences (and won't result in significant or possibly any clean up) or resolve the differences above (which would require more discussion and possibly test changes). Given these issues I think these changes warrent a separate PR (and aren't required by the `ModelLibrary`).

schlafly: FYI, I think blendheaders is just a Webb holdover and we have done basically everything we need to for the L3 metadata.  i.e., we could just get rid of blendheaders.

The single datamodel resample case I don't think is one we need to worry a lot about; if we want to change the behavior there to support a simpler code flow, we could do that.  But I agree we can take that up in a separate PR.

braingram:
I opened the stpipe PR adding `ModelLibrary` https://github.com/spacetelescope/stpipe/pull/156

My time is otherwise occupied but I will (hopefully soon) update this PR and open it for review but wanted to post a message now in case anyone wants to look at and weigh-in on the stpipe PR. Any feedback is appreciated!

romancal/flux/flux_step.py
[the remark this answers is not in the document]
braingram: I'm going to mark this as "resolved". I agree that refactoring this and moving it to some common location is a great idea. However I'd like to restrict the scope of this PR (which is already quite large). Let me know if it's helpful to open an issue or ticket for this.

romancal/outlier_detection/outlier_detection_step.py
braingram: I opened an issue to track this change: https://github.com/spacetelescope/romancal/issues/1302 If you're rather I make the change in this PR I'm happy to do so.

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -140,6 +151,12 @@ def test_outlier_init_default_parameters(pars, base_image):
     assert step.resample_suffix == f"_outlier_{pars['resample_suffix']}.asdf"
 
 
+# FIXME: This test checks if the median image exists on disk after outlier detection.
+# Howver "save_intermediate_results=False" so this file should not be saved even if
+# in_memory=False (which only means the file will temporarily be produced if needed).
+@pytest.mark.skip(
```
braingram, line 157: I opened an issue to track this skip:

https://github.com/spacetelescope/romancal/issues/1303

It's possible that setting `save_intermediate_files` to True would be enough to fix this but I did not do so in this PR to try to limit the scope. If it's preferred to make the fix vs add this skip I'm happy to do so.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1071,17 +1100,24 @@ def test_tweakreg_handles_multiple_groups(tmp_path, base_image):
...
+            res.shelve(r, modify=False)
+
+
+# FIXME: this test previously passed only because Tweakreg overwrote
+# the model.meta.group_id using the output from _common_name which incorrectly
+# ignored the setting of "program" in this test. This meant that tweakreg
+# found 2 groups for this test data (because ModelContainer.models_grouped)
+# worked as intended but tweakreg then overwrote the grouping and then
+# incorrectly skipped itself.
+@pytest.mark.skip(reason="This test previously incorrectly passed")
```
braingram, line 1119: I opened in issue to track this skip:
https://github.com/spacetelescope/romancal/issues/1304

romancal/pipeline/exposure_pipeline.py
braingram: Resolving due to lack of response.

romancal/resample/resample.py
braingram: @schlafly I believe this is now handled in resample (instead of the container):
https://github.com/spacetelescope/romancal/pull/1241/files#diff-f87241f9a890ac31d6813379a09b6eb7e1052e3ab53039449a8f0df91aa030e1R351
but it would be great if this could be confirmed (I don't see any test failures but I'm not sure if there is a test for this feature).

romancal/resample/resample.py
braingram: I opened an issue to track this change: https://github.com/spacetelescope/romancal/issues/1308

romancal/tweakreg/tweakreg_step.py
braingram: Resolving due to lack of response.

romancal/flux/flux_step.py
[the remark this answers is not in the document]
braingram: I opened an issue to track this change: https://github.com/spacetelescope/romancal/issues/1309

romancal/resample/resample_step.py
braingram: I opened an issue to track the input handling logic: https://github.com/spacetelescope/romancal/issues/1309

romancal/tweakreg/tests/test_tweakreg.py
braingram: I'm marking this as resolved as based on https://github.com/spacetelescope/romancal/pull/1301 this test will be deleted. The skip is necessary here as this PR fixes the bug that is allowing it to pass.

romancal/tweakreg/tweakreg_step.py
mairanteodoro: > @mairanteodoro , this can probably be uncommented now?

Done in [PR#1276](https://github.com/spacetelescope/romancal/pull/1276).

braingram:
Using the [mosaic regression test](https://github.com/spacetelescope/romancal/blob/459e8e345bbb1ba6bfd76d0d8a90f2058089791e/romancal/regtest/test_mos_pipeline.py#L36) data and running the mosaic pipeline on my local machine (with cached CRDS reference files).

For romancal main: takes 4 minutes and 8 seconds, peaks at 8.7 GB memory usage
For this PR: takes takes 2 minutes and 55 seconds, peaks at 7.1 GB memory usage

With this PR the mosaic pipeline takes 70% of the previous time and uses 81% of the previous memory.

romancal/resample/resample.py
schlafly: Thanks, sorry for losing this thread!  The issue was that meta.filename can easily become disconnected from the actual file name on disk, and there was a request for storing the actual file names on disk used to create a L3 file, rather than the meta.filenames that contributed.  My initial impression here
https://github.com/spacetelescope/romancal/pull/1241/files#diff-f87241f9a890ac31d6813379a09b6eb7e1052e3ab53039449a8f0df91aa030e1R351
is that this returns to the previous behavior of using meta.filename.  If there's a convenient way to ask the container for the original on-disk filenames, that would be best.

braingram: Thanks. Is there a test for this (so I can poke it to see what's going on with and without this PR)?

To make sure I'm understanding the goal, let's say an association is provided to the mosaic pipeline that lists members "foo.asdf" and "bar.asdf" which have meta.filename attributes that are something else "chutes.asdf" and "ladders.asdf". The goal is to have the output resampled product contain "foo.asdf" and "bar.asdf". I think that should be easy to do by using the members info from `Library.asn` but I'll have to run some tests.

mairanteodoro (commented):
Looks good to me! Thanks, @braingram!

romancal/datamodels/library.py
```diff
@@ -0,0 +1,82 @@
...
+            raise OSError("Cannot read ASN file.") from e
+        return asn_data
+
+    def _filename_to_group_id(self, filename):
+        """
+        Compute a "group_id" without loading the file as a DataModel
+
+        This function will return the meta.group_id stored in the ASDF
+        extension (if it exists) or a group_id calculated from the
+        FITS headers.
```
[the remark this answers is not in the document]
braingram: Oh my... let's pretend that never made it in. :) I'll make that switch. Thanks!

braingram:
Added the `update webbpsf data` label as many tests started complaining about lack of webbpsf data.

romancal/datamodels/library.py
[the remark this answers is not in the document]
braingram: Fix in [d73405e](https://github.com/spacetelescope/romancal/pull/1241/commits/d73405ef0c046cb70e12a1ecf370f64c12f32514)

romancal/resample/resample.py
braingram: @schlafly I switched to populating this using the asn members (which are the input filenames) in:
[449a58d](https://github.com/spacetelescope/romancal/pull/1241/commits/449a58d5f9d3029aaf151c46f581c77b38fc46db)
I didn't see any test differences so I'm assuming there is no test for this behavior (and the mosaic pipeline input files all have meta.filenames that match the real filenames).

I modified the test data for the mos pipeline test (locally) renaming one file in the association and on disk (but leaving it's meta.filename the same) and then run the pipeline. The result from resample recorded the modified name (the real filename) and not meta.filename.

I'd be happy to open an issue to track adding a test for this or adding one to this PR (I think it should be possible to add as a resample only unit test but that would require some investigation). Which would you prefer, an issue to track adding the test or adding it to this PR?

schlafly: Perfect, thanks, that looks great.  Just making sure I hadn't missed it before, that's a new feature in ModelLibrary that wasn't available to me when I wrote that bit of code before?  Great.  It looks like when the ModelLibrary is constructed from a list of models, these expnames default to meta.filename, but that seems like it isn't a regression since for some set of in-memory models I don't know what to do otherwise?

I tried to trace the code briefly and was a bit confused about what the resample step is now allowed to support as input.  The docstring is presumably out of date:
https://github.com/spacetelescope/romancal/blob/449a58d5f9d3029aaf151c46f581c77b38fc46db/romancal/resample/resample.py#L61-L62
This bit:
https://github.com/spacetelescope/romancal/blob/449a58d5f9d3029aaf151c46f581c77b38fc46db/romancal/resample/resample.py#L80-L84
says either a list of models or a ModelLibrary.
This bit
https://github.com/spacetelescope/romancal/blob/449a58d5f9d3029aaf151c46f581c77b38fc46db/romancal/resample/resample.py#L134-L135
suggests to me that it has to be a ModelLibrary (which presumably then always has the asn member).

I don't have strong preferences for an issue vs. an addition to this PR for a new test.  It looks like this PR is pretty close to being ready to merge, so I'm happy to delay a new test to an issue.  Thanks!

braingram: Thanks!

> Just making sure I hadn't missed it before, that's a new feature in ModelLibrary that wasn't available to me when I wrote that bit of code before?

Exactly. I think the members entries previously got lost in skymatch due to:
https://github.com/spacetelescope/romancal/blob/747667895099a80ef283838a8499bac391cb1c46/romancal/skymatch/skymatch_step.py#L100

> It looks like when the ModelLibrary is constructed from a list of models, these expnames default to meta.filename, but that seems like it isn't a regression since for some set of in-memory models I don't know what to do otherwise?

Yup! Although we certainly could change that if something else seems more useful. The general idea is to make the `ModelLibrary` always have a "valid" association (even if it's made up based off the models) so the step code has a consistent interface (with `ModelContainer` I think you only get something in `asn_table` if the input is an association).

Re: resample step

Thanks for the catch. Yes the `ResampleData` class now only accepts a `ModelLibrary`. The `ResampleStep` should accept the same input as before (except for a `ModelLibrary` instead of a `ModelContainer`).
I pushed a change to the docstring in: https://github.com/spacetelescope/romancal/pull/1241/commits/6184692d1af66db786b59769598fc1ce6a346b97
and the error message in: https://github.com/spacetelescope/romancal/pull/1241/commits/899c9aa38bc275bc654d161571bca106cec56e67

I still have to rebase this to deal with the horde of conflicts following https://github.com/spacetelescope/romancal/pull/1314 so the CI is unhappy temporarily.

braingram: I believe the linked commits addressed the documentation issues (thanks for finding those!) and I was able to "rebase". Rerunning the regtests shows the same results as linked in the above description (the run link is up-to-date).

schlafly:
Great.  As far as you're concerned, this is good to merge?  @ddavis-stsci , anything else that you would like to see?

schlafly (approved):
Looks good to me, thanks for all of this!

PR #1260, opened 2024-06-03
use resample many_to_many in outlier detection

braingram:
This PR changes outlier detection to use `many_to_many` when resampling during median calculation. This PR produces 1 expected difference in regression test results.
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/805/

With the current code on main the `test_level3_mos_pipeline` test runs an association with 3 models (each from a different exposure) through the mosaic pipeline. Setting a breakpoint just prior to the call to `create_median`:
https://github.com/spacetelescope/romancal/blob/ed6187ffa790fc6a37dd96f7365ba1ef8261fa74/romancal/outlier_detection/outlier_detection.py#L108-L111
we can see that `len(drizzled_models) == 1`. This conflict with the [algorithm description](https://roman-pipeline.readthedocs.io/en/latest/roman/outlier_detection/outlier_detection.html) in the docs that states:
```
Each dither position will result in a separate grouped mosaic, so only a single exposure ever contributes to each pixel in these mosaics.
```

Looking at the [outlier detection unit tests](https://github.com/spacetelescope/romancal/blob/main/romancal/outlier_detection/tests/test_outlier_detection.py) I don't see one that both:
- runs the step with `resample_data=True`
- introduces outliers
The only unit test that reaches:
https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/outlier_detection/outlier_detection.py#L86-L88
is [test_skymatch_always_returns_modelcontainer_with_updated_datamodels](https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/outlier_detection/tests/test_outlier_detection.py#L341C5-L341C72) which doesn't introduce any outliers and runs the step with all 0 data.

Changing [test_outlier_do_detection_find_outliers](https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/outlier_detection/tests/test_outlier_detection.py#L194C5-L194C44) (which introduces and detects outliers) to use resampling is insufficient to show the impact of this PR. Without this PR, the introduced CRs (value=1E5) and 'drizzled' together with empty pixels from the other image (value=0). This produces a single `drizzled_model` with a value of 5E4 for each PR. The median (N=1) produces the same value, when blotted back to the input image wcs the 5E4 values are far enough below the 1E5 values of the CRs to allow them to be detected. This is true for any value used for the CR (since the input images are noiseless with all 0 error).

Furthermore the test appears to be checking that all introduces CRs (even those introduced in `img2`) are flagged in `img1`:
https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/outlier_detection/tests/test_outlier_detection.py#L253
returns 10 flagged CRs:
```python
(array([ 5, 15, 25, 35, 45, 65, 75, 85, 95, 99]), array([45, 25, 25,  5, 85, 65, 65,  5, 45,  5]))
```
whereas only 5 were introduced into `img1`:
https://github.com/spacetelescope/romancal/blob/e559890ee4679835b11e98e17f33ac819733176d/romancal/outlier_detection/tests/test_outlier_detection.py#L204-L206

Switching single (so that now `many_to_many` is used) revealed a few other issues included:
- drizzled models overwriting each other due to filenames sharing the same base
- drizzled models output including only the last used group
- resample background correction check not matching the check in resample

The updated unit test in this PR uses 3 images with the first 2 having CRs, each having 1 "source" and checks that:
- CRs added to image 0 are flagged in image 0
- CRs added to image 1 are flagged in image 1
- no CRs are flagged in image 2

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "tweakwcs >=0.8.6",
     "spherical-geometry >= 1.2.22",
     "stsci.imagestats >= 1.6.3",
-    "drizzle >= 1.13.7",
+    "drizzle >= 1.14.0",
     "webbpsf == 1.2.1",
 ]
 dynamic = [
--- a/romancal/outlier_detection/tests/test_outlier_detection.py
+++ b/romancal/outlier_detection/tests/test_outlier_detection.py
@@ -191,30 +191,35 @@ def test_outlier_do_detection_write_files_to_custom_location(tmp_path, base_imag
     assert all(x.exists() for x in outlier_files_path)
 
 
-def test_outlier_do_detection_find_outliers(tmp_path, base_image):
+def test_find_outliers(tmp_path, base_image):
     """
     Test that OutlierDetection can find outliers.
     """
-    img_1 = base_image()
-    img_1.meta.filename = "img_1.asdf"
-    img_2 = base_image()
-    img_2.meta.filename = "img_2.asdf"
+    cr_value = Quantity(100, "DN / s")
+    source_value = Quantity(10, "DN / s")
+    err_value = Quantity(10, "DN / s")  # snr=1
+
+    imgs = []
+    for i in range(3):
+        img = base_image()
+        img.data[42, 72] = source_value
+        img.err[:] = err_value
+        img.meta.filename = f"img{i}_suffix.asdf"
+        img.meta.observation.exposure = i
+        imgs.append(img)
 
     # add outliers
+    img_0_input_coords = np.array(
+        [[5, 25, 45, 65, 85], [45, 25, 85, 65, 5]], dtype=np.int64
+    )
     img_1_input_coords = np.array(
-        [(5, 45), (25, 25), (45, 85), (65, 65), (85, 5)], dtype=[("x", int), ("y", int)]
-    )
-    img_2_input_coords = np.array(
-        [(15, 25), (35, 5), (75, 65), (95, 45), (99, 5)], dtype=[("x", int), ("y", int)]
-    )
-    img_1.data[img_1_input_coords["x"], img_1_input_coords["y"]] = Quantity(
-        100000, "DN / s"
-    )
-    img_2.data[img_2_input_coords["x"], img_2_input_coords["y"]] = Quantity(
-        100000, "DN / s"
+        [[15, 35, 75, 95, 99], [25, 5, 65, 45, 5]], dtype=np.int64
     )
 
-    input_models = ModelContainer([img_1, img_2])
+    imgs[0].data[img_0_input_coords[0], img_0_input_coords[1]] = cr_value
+    imgs[1].data[img_1_input_coords[0], img_1_input_coords[1]] = cr_value
+
+    input_models = ModelContainer(imgs)
 
     outlier_step = OutlierDetectionStep()
     # set output dir for all files created by the step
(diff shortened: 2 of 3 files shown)


schlafly:
This looks fine to me, but @mairanteodoro and you should discuss, and I suspect that there are related unit tests that will need updating.

braingram:
Thanks @schlafly!

@mairanteodoro and I have a meeting tomorrow to discuss the sky subtraction (which is partially handled in this PR and partially in https://github.com/spacetelescope/romancal/pull/1233). After updating the `flag_outlier` unit test for this PR I don't think the sky subtraction was the only issue so there are a number of changes in this PR.

I'll open this PR for review once the unit and regtests finish (I expect at least the `mosaic` regtest to fail and any others that use outlier detection).

pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "tweakwcs >=0.8.6",
     "spherical-geometry >= 1.2.22",
     "stsci.imagestats >= 1.6.3",
-    "drizzle >= 1.13.7",
+    "drizzle >= 1.14.0",
```
braingram, line 37: This was needed as an outlier at the very edge of the image wasn't being picked up in drizzle <1.14.0.

schlafly:
Presumably the regtests change meaningfully with this PR; would you mind including something like the new image, old image, and the difference?  Thank you!

braingram:
> Presumably the regtests change meaningfully with this PR; would you mind including something like the new image, old image, and the difference? Thank you!

I think the only regtest that is impacted is the mosaic pipeline test. This uses only 3 images and produces very minor differences in the number of detected outliers. With main:
```
2024-06-04 16:06:53,489 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 6373 (0.04%)
2024-06-04 16:06:54,670 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 11289 (0.07%)
2024-06-04 16:06:55,898 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 6441 (0.04%)
```
with this PR:
```
2024-06-04 16:01:20,422 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 6144 (0.04%)
2024-06-04 16:01:21,795 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 15462 (0.09%)
2024-06-04 16:01:22,888 - stpipe.MosaicPipeline.outlier_detection - INFO - New pixels flagged as outliers: 6151 (0.04%)
```

Overall I don't think the output change is very meaningful given the current regression tests.
Here is the `stpreview by <fn> 16 16` output for the truth file:
truth
and the new output with this PR:
new

schlafly:
I agree that I can't see anything in the preview images; the stars are too small and just look like hot pixels.  I don't actually think the L2 files are that problematic; some of the structure in the background is from the non-linearity reference files having issues.  The spatial gradient is a little unexpected but is probably very low amplitude.  By eye it doesn't look to me like we're masking all of the stars, for example, but have you actually looked at the generated outlier masks?

mairanteodoro (approved):
Looks good to me! Thanks for the fix, @braingram!!

braingram:
> I agree that I can't see anything in the preview images; the stars are too small and just look like hot pixels. I don't actually think the L2 files are that problematic; some of the structure in the background is from the non-linearity reference files having issues. The spatial gradient is a little unexpected but is probably very low amplitude. By eye it doesn't look to me like we're masking all of the stars, for example, but have you actually looked at the generated outlier masks?

I looked at the outlier masks and there aren't large differences. Since main has seen some changes since I made those comparisons they'll need to be updated. The mosaic test only uses 3 images so the median across groups (many_to_many) vs the drizzled combinations of all groups (many_to_one) produces similar "median" data.

schlafly (approved):
Great, thanks, approving!

braingram:
@schlafly Is there anything else you'd like to see for this PR?

schlafly:
No, thanks, I thought I approved, thank you!

PR #1262, opened 2024-06-03
rcal-846 Add mosaic info to the science products RTD page

### LINKED ISSUE #1252: Add mosaic data description to RTD
_Issue [RCAL-846](https://jira.stsci.edu/browse/RCAL-846) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

The mosaic data should be described in the RTD section for science products ([https://roman-pipeline.readthedocs.io/en/latest/roman/data_products/science_products.html#)]

ddavis-stsci:
Resolves [RCAL-846](https://jira.stsci.edu/browse/RCAL-846)

Closes #1252

This PR adds a description of the i2d (mosaic) data to read-the-docs. 

**Checklist**
- [x] added relevant label(s)
- [N/A] ran regression tests, post a link to the Jenkins job below. [How to run regression tests on a PR](https://github.com/spacetelescope/romancal/wiki/Running-Regression-Tests-Against-PR-Branches)


docs/roman/data_products/science_products.rst
```diff
@@ -139,3 +139,39 @@ The calibrated products are the result of an average over all integrations (``ca
...
+|  err                 | Required | float32    | e\ :sup:`-`/ s            |  nrows x ncols                |
++----------------------+----------+------------+---------------------------+-------------------------------+
+|  weight              | Required | float32    | N/A                       |  nrows x ncols                |
++----------------------+----------+------------+---------------------------+-------------------------------+
+|  var_poisson         | Required | float32    | (e\ :sup:`-`/ s)\ :sup:`2`|  nrows x ncols                |
++----------------------+----------+------------+---------------------------+-------------------------------+
+|  var_rnoise          | Required | float32    | (e\ :sup:`-`/ s)\ :sup:`2`|  nrows x ncols                |
++----------------------+----------+------------+---------------------------+-------------------------------+
+|  var_flat            | Required | float32    | (e\ :sup:`-`/ s)\ :sup:`2`|  nrows x ncols                |
++----------------------+----------+------------+---------------------------+-------------------------------+
```
schlafly, line 167: Did you check the units here?  The flux calibration step is supposed to make sure the units are MJy/sr (or the square of that for the variances), but we may have errors, of course.  I had understood the context image to be N x nrows x ncols, where N depends on the number of input images, and is 1 for sky cells with fewer than 32 input images.  Not true?

schlafly (commented):
Looks good to me, though I left a comment inline with some questions about some of the fields in the i2d files...

docs/roman/data_products/science_products.rst
ddavis-stsci: I must have looked at an old file. I've updated the units. 

for the number of planes in the context image I thought as you said, it should be the number of input images modulo 32. I in the output I see
context (NDArrayType): shape=(2, 9191, 8642), dtype=uint32 # Context Data
The number of input images was 8 (not all overlapping). 
Maybe a bug?

schlafly:
Was this the number of input SCAs or exposures?  If the latter, maybe the number of SCAs is >32, and this is the fact that we're not grouping?  But yeah, I suspect a bug of some kind.

ddavis-stsci:
8 detectors, the asn code looks for individual detector files that touch 
a skycell. In this case
WFI01, WFI02, WFI10 & WFI11 for the given visit and skycell.

schlafly (approved):
Weird, thanks.  Let's file a ticket to investigate the number of planes to expect, but looks good to me.

PR #1269, opened 2024-06-11
Fix comparison operator.

mairanteodoro:
This should fix the issue mentioned by George Cleary on Slack: https://stsci.slack.com/archives/C01FP900P38/p1718113704757169

**Checklist**
- [X] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/source_catalog/tests/test_source_catalog.py
+++ b/romancal/source_catalog/tests/test_source_catalog.py
@@ -381,7 +381,7 @@ def test_do_psf_photometry(tmp_path, image_model):
     # check the number of sources that have been detected
     assert len(cat) == 7
     # check that all sources have both position and flux determined (ignore errors/flags)
-    assert all(len(cat[x]) and cat[x] is not [None, np.nan] for x in psf_colnames)
+    assert all(len(cat[x]) and cat[x] != [None, np.nan] for x in psf_colnames)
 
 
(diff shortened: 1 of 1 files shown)


braingram:
@mairanteodoro would you also add the webbpsf markers to these tests? They're causing failures in stpipe:
https://github.com/spacetelescope/stpipe/actions/runs/9445531594/job/26013349354#step:10:3493

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -381,7 +381,7 @@ def test_do_psf_photometry(tmp_path, image_model):
     # check the number of sources that have been detected
     assert len(cat) == 7
     # check that all sources have both position and flux determined (ignore errors/flags)
-    assert all(len(cat[x]) and cat[x] is not [None, np.nan] for x in psf_colnames)
+    assert all(len(cat[x]) and cat[x] != [None, np.nan] for x in psf_colnames)
```
schlafly, line 384: I suspect we want "not in" rather than !=, right, or do I misunderstand?  Actually, cat is a Table, so cat[x] should be a numpy array, so we probably want something like np.all(np.isfinite(cat[x]))?

schlafly (approved):
Looks good, thanks.  (but please wait to merge until tests pass!)

PR #1270, opened 2024-06-11
allow pars files

braingram:

Closes #

This PR addresses ...

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


stscieisenhamer:
Some questions:

First, how does this "enable" parameter reference files? From the above, the implication is that parameter reference files are json files. However, at least for jwst, parameter reference files are asdf files. 

This is also implying that stpipe parameter reference mechanism is not in the core `stpipe`. Is this true and if so, why is the implementation different between roman and jwst?

There may be need for some more discussion, because of the potentially related [CCD-1465](https://jira.stsci.edu/browse/CCD-1465).

braingram:
> Some questions:

Thanks for taking a look and for the link to the CCD ticket. This isn't quite ready for review and I wasn't sure if the infrastructure is ready for these changes. I started the PR because I was looking at a different issue and fixed `_datamodels_open` as a side-effect.

> First, how does this "enable" parameter reference files? From the above, the implication is that parameter reference files are json files. However, at least for jwst, parameter reference files are asdf files.

It allows `Step._datamodels_open` to open an association file (stored as json or yaml). The `foo.json` mentioned above is the association file not the parameter file. Attempting to open an association fails (as noted above) with romancal (it does not fail with jwst, that's the bug fixed here).

> This is also implying that stpipe parameter reference mechanism is not in the core `stpipe`. Is this true and if so, why is the implementation different between roman and jwst?

The parameter reference fetching is all handled by `stpipe`, the same as with jwst.

> There may be need for some more discussion, because of the potentially related [CCD-1465](https://jira.stsci.edu/browse/CCD-1465).

Thanks for the link. This PR is fixing a bug that will allow the parameter file to be fetched whereas that looks to be related to the format of the parameter file (this PR doesn't change anything about the format).

schlafly:
Looks good to me; certainly will be glad not to need --disable-crds-steppars anymore.  @ddavis-stsci , I remember you had thoughts here; care to review?

braingram:
Thanks! I requested @stscieisenhamer as another reviewer as he had some questions about these changes.

braingram:
@stscieisenhamer any more questions on this PR?

stscieisenhamer (approved):

PR #1273, opened 2024-06-13
RCAL-856: WCS should roundtrip with sufficient accuracy

### LINKED ISSUE #1272: WCS does not roundtrip
_Issue [RCAL-856](https://jira.stsci.edu/browse/RCAL-856) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

It was discovered that the Roman WCS transforms do not roundtrip with the necessary accuracy.  This affects the simulator, general testing and imviz. it is a combination of several different issues and this ticket tracks the work and solutions to those issues.

nden:
Resolves [RCAL-856](https://jira.stsci.edu/browse/RCAL-856)

Closes #1272 

This PR  is the last one in a chain of PRs aiming to fix roundtripping of WCS transforms.
The only change here is to test data, ensuring the WCS is consistent and the inverse is wihtin the bounding box.

The order the PRs should be merged is
- [ ] spacetelescope/gwcs#498
- [ ] spacetelescope/stcal#263
- [ ] this one

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


ddavis-stsci (approved):
LGTM

schlafly:
I think I missed the part of this test that checks that things round trip?

This WCS contains only the v2v3tosky bit, and not the CRDS distortion bit.  That's good since it makes for an easier unit test but bad since it won't catch the only case we have had problems here so far, where the failure to round trip was in the CRDS distortion file.  Do we intend to check the CRDS reference files as well?

I note that the bounding box is inconsistent with the size of the image, and that we're pretending that the sizes of the pixels are ~1".  I think that's likely fine.  romanisim has some code for making up a "fake" distortion function here
https://github.com/spacetelescope/romanisim/blob/main/romanisim/tests/test_wcs.py#L19-L38
and using that in unit tests instead of the actual CRDS distortion functions.  It's not obvious to me that that is helpful here but I link it in case you would have rather used some kind of distortion function.

nden:
This requires a change in gwcs which I haven't had time to complete. I'll take it out of Draft when it's ready.

nden:
This is not needed after all. Closing.

PR #1275, opened 2024-06-17
pin numpy <2

braingram:
Numpy 2.0 was released. This is causing many tests to fail due to incompatible depedencies:
https://github.com/spacetelescope/romancal/actions/runs/9539663192/job/26290288431

These include:
- drizzle (soon to be addressed in https://github.com/spacetelescope/drizzle/pull/149)
- opencv (see https://github.com/opencv/opencv-python/issues/943)

This PR pins numpy to <2.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)


CHANGES.rst
```diff
@@ -10,6 +10,8 @@ Documentation
 general
 -------
 
+- pin numpy to <2 [#1275]
+
```
ddavis-stsci, line 14: We don't generally add dependency versions to the changes file.

ddavis-stsci (approved):

PR #1276, opened 2024-06-17
RCAL-830: update tweakreg to use source catalog

mairanteodoro:
Closes https://github.com/spacetelescope/romancal/issues/1317
Resolves [RCAL-830](https://jira.stsci.edu/browse/RCAL-830)

This PR addresses the integration between `TweakRegStep` and `SourceCatalogStep`.

**Regression tests**
Three consecutive runs with consistent and expected (due to changes in this PR) results:
- https://github.com/spacetelescope/RegressionTests/actions/runs/10009795321
- https://github.com/spacetelescope/RegressionTests/actions/runs/10010386214
- https://github.com/spacetelescope/RegressionTests/actions/runs/10012275757

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


schlafly (commented):
Looks good to me.  I left some comments inline.

One important question is what we want the step to return.  The current design is that you run the step on an image and get a catalog.  That gets saved ~automatically if it's the only step you run.  Now you run the step on an image and get an image out, and the catalog is a side effect that gets saved.  That feels more awkward at some level but it matches the design of other steps where it makes more sense, and I don't know if it makes sense for the step to return both the model and the associated catalog.  The biggest downside of the new design would be that we can't get the resulting catalog out in memory; we have to save it to use it.  That probably sounds fine to me, though as part of the DR catalogs we may want to assemble a bunch of catalog objects in memory before merging them and writing them out.  Another possibility would be continuing to return the SourceCatalog and to add additional code to the pipeline to mark things complete and update the file name, but that's pretty unappealing too.

Do you see any other alternatives here?  It is possible for a step to return two models or something?

romancal/source_catalog/source_catalog_step.py
```diff
@@ -43,6 +43,7 @@ class SourceCatalogStep(RomanStep):
         ci2_star_threshold = float(default=1.8)  # CI 2 star threshold
         suffix = string(default='cat')        # Default suffix for output files
         fit_psf = boolean(default=True)      # fit source PSFs for accurate astrometry
+        return_updated_model = boolean(default=False)   # should the updated model be returned?
```
schlafly, line 46: Can we expand the comment here?  Maybe: should image model be updated and returned, instead of returning the catalog?  I realize one doesn't have much space, though.

romancal/source_catalog/source_catalog_step.py
```diff
@@ -94,35 +95,68 @@ def process(self, input_model):
...
-        return source_catalog_model
+            # always save segmentation image + catalog
+            self.save_base_results(segment_img, source_catalog_model)
+
+            # return the updated model or the source catalog object
+            if self.return_updated_model:
+                # setting the suffix to something else to prevent
+                # step from  overwriting source catalog file with
+                # a datamodel
+                self.suffix = "source_catalog"
```
schlafly, line 136: Can we make this something without an underscore?  sourcecatalog?

romancal/source_catalog/source_catalog_step.py
```diff
@@ -94,35 +95,68 @@ def process(self, input_model):
...
+            segmentation_model.data = segment_img.data.astype(np.uint32)
+            self.save_model(segmentation_model, suffix="segm", force=True)
+        # save the source catalog
+        self.save_model(source_catalog_model, suffix="cat", force=True)
+
+
+def update_metadata(model, output_catalog_name):
+    # update datamodel to point to the source catalog file destination
+    model.meta["source_detection"] = maker_utils.mk_source_detection(
+        **{"tweakreg_catalog_name": output_catalog_name}
```
schlafly, line 160: Is this the same as mk_source_detection(tweakreg_catalog_name=output_catalog_name)?

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -317,8 +329,8 @@ def test_l3_input_model_unchanged(mosaic_model, tmp_path):
         save_results=False,
     )
 
-    assert_allclose(original_data, mosaic_model.data, atol=5.0e-7)
-    assert_allclose(original_err, mosaic_model.err, atol=5.0e-7)
+    assert_allclose(original_data, mosaic_model.data, atol=5.0e-5)
+    assert_allclose(original_err, mosaic_model.err, atol=5.0e-5)
```
schlafly, line 333: Should we worry about the lower tolerance here?  I think the only action we intend to take is to add the background and then subtract it again.  It is possible for that to introduce issues but it seems like they should be very small; I didn't expect this PR to change the accuracy of that?

romancal/tweakreg/tweakreg_step.py
```diff
@@ -175,37 +175,36 @@ def process(self, input):
...
                 else:
-                    raise AttributeError(
-                        "Attribute 'meta.source_detection.tweakreg_catalog' is missing."
-                        "Please either run SourceDetectionStep or provide a"
-                        "custom source catalog."
-                    )
-                # remove 4D numpy array from meta.source_detection
-                if is_tweakreg_catalog_present:
-                    del image_model.meta.source_detection["tweakreg_catalog"]
+                    catalog = Table.read(catalog_name, format=self.catalog_format)
```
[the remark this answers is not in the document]
schlafly: This still makes sense to me.

schlafly (commented):
A few minor changes inline.  The most important thing to me is understanding why the tolerance needed to change from atol=5e-7 to atol=5e-5; I worry that that's a sign that there's a real issue?

romancal/source_catalog/source_catalog_step.py
ddavis-stsci: Yes this should match what is in the suffix library, sourcedetection

romancal/source_catalog/tests/test_source_catalog.py
mairanteodoro: We needed to change the datatype for the `data` array in the `make_test_image()` method to `float32`, which caused larger differences, and, consequently, we had to increase the tolerance by a factor of 100. If we don't set the data array to `float32` then we run into validation issues with saving the mock models.

romancal/regtest/test_wfi_pipeline.py
```diff
@@ -36,6 +36,7 @@ def test_level2_image_processing_pipeline(rtdata, ignore_asdf_paths):
     output = "r0000101001001001001_01101_0001_WFI01_cal.asdf"
     rtdata.output = output
     args = [
+        "--steps.source_catalog.return_updated_model=True",
```
schlafly, line 39: I think we should remove this from here and instead put it in exposure_pipeline.py; i.e., the pipeline should know it needs to call source_catalog with this argument.

schlafly (commented):
This looks good to me.  It looks to me like we still also support the old source_detection catalog in tweakreg; since we can no longer make those catalogs, we should remove that?  I also suggest in line that we should make the ELP know about the return_updated_model argument rather than making users know to specify that argument if they want the pipeline to run successfully.

romancal/source_catalog/source_catalog_step.py
```diff
@@ -43,6 +43,8 @@ class SourceCatalogStep(RomanStep):
         ci2_star_threshold = float(default=1.8)  # CI 2 star threshold
         suffix = string(default='cat')        # Default suffix for output files
         fit_psf = boolean(default=True)      # fit source PSFs for accurate astrometry
+        return_updated_model = boolean(default=False)   # should the image model be updated and returned
```
braingram, line 46: Why not return both?

braingram: To fill in some details from a discussion with @schlafly (thanks!). It sounds like the goal here is to:
- allow a user to call this step and have it just return the catalog
- have the pipeline call this step and update the model passed as input

Having the step return both would mean the step would have to manage the file saving to assign different suffixes (and remove `idx`) for each result (the [ami_analyze step](https://github.com/spacetelescope/jwst/blob/ce9a135f3bc6bba21e95dd47a91fc4d9ba5a1991/jwst/ami/ami_analyze_step.py#L29) in jwst does this).

I'm just leaving this comment for future reference. I initially commented to try and understand the usage (this is the first time I've seen the spec used in this way) and I think now I understand. Thanks and feel free to mark this as "resolved".

mairanteodoro: Not sure as to how `stpipe` will handle multiple returned results.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -151,7 +151,7 @@ def process(self, input):
             if result.meta.exposure.type == "WFI_IMAGE":
                 result = self.flatfield(result)
                 result = self.photom(result)
-                result = self.source_detection(result)
```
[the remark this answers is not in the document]
mairanteodoro: I tried
```python
result = self.source_catalog(result, return_updated_model=True)
```
but it fails (keyword is not known). I'll check with @ddavis-stsci to see if we can do something here.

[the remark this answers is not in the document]
braingram: Since it's part of the `spec` it's expected in the `Step.__init__` (and a few other places). The `__call__` is an [alias](https://github.com/spacetelescope/stpipe/blob/01f1956f3b3c3750a26790d5217cf909cad4b95a/src/stpipe/step.py#L571) to `run` which [only supports](https://github.com/spacetelescope/stpipe/blob/01f1956f3b3c3750a26790d5217cf909cad4b95a/src/stpipe/step.py#L411) `*args`.

Perhaps `return_updated_model` could be a `Step` attribute, set in the pipeline with:
```
self.source_detection.return_updated_model = True
```
and checked in the `Step.process` with:
```
if getattr(self, "return_updated_model, False):
```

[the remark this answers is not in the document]
mairanteodoro: Thanks for the suggestion, @braingram! Let me give it a shot.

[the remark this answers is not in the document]
mairanteodoro: Looks like this approach worked, @braingram!

schlafly:
This looks reasonable to me.  Can you confirm that you can run the ELP successfully and that regtests pass, except for compare_asdf failures having to do with missing source_detection stuff and slight changes to the WCS?

I think @braingram's suggestion was to remove return_updated_model from the spec and then replace this line with a getattr() call that uses False as a default.
https://github.com/spacetelescope/romancal/pull/1276/files#diff-20b0d7c2e6ef2cb1dfd2df9d42ecdabdd86fb400b73755cb4a58598ddd200c2eR133
I think that looks straightforward, but you know better.

mairanteodoro:
> This looks reasonable to me. Can you confirm that you can run the ELP successfully and that regtests pass, except for compare_asdf failures having to do with missing source_detection stuff and slight changes to the WCS?
> 
Yes, I can confirm that we can run the ELP successfully.

I'm re-running the regression tests right now (locally and via GA).

romancal/source_catalog/source_catalog_step.py
braingram, line 46:
```suggestion
```

romancal/source_catalog/source_catalog_step.py
```diff
@@ -94,35 +96,76 @@ def process(self, input_model):
...
-                segmentation_model.data = segment_img.data.astype(np.uint32)
-                self.save_model(segmentation_model, suffix="segm")
+            output_model = model
 
-        return source_catalog_model
+            # always save segmentation image + catalog
+            self.save_base_results(segment_img, source_catalog_model)
+
+            # return the updated model or the source catalog object
+            if self.return_updated_model:
```
braingram, line 133:
```suggestion
            if getattr(self, "return_updated_model", False):
```

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -432,3 +446,205 @@ def test_do_psf_photometry_column_names(tmp_path, image_model, fit_psf):
...
+    os.chdir(tmp_path)
+    step = SourceCatalogStep()
+    result = step.call(
+        mosaic_model,
+        bkg_boxsize=50,
+        kernel_fwhm=2.0,
+        snr_threshold=snr_threshold,
+        npixels=npixels,
+        save_results=save_results,
+        return_updated_model=return_updated_model,
```
braingram, line 631:
```suggestion
```

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -432,3 +446,205 @@ def test_do_psf_photometry_column_names(tmp_path, image_model, fit_psf):
...
+    expected_result,
+    expected_outputs,
+    tmp_path,
+):
+    """
+    Test that the proper object is returned in the call to SourceCatalogStep
+    and that the desired output files are saved to the disk with the correct type.
+    """
+    os.chdir(tmp_path)
+    step = SourceCatalogStep()
```
braingram, line 623:
```suggestion
    step = SourceCatalogStep()
    # to mimic what happens in the elp set the "hidden" parameter
    # to cause this step to return a model instead of a catalog
    if return_updated_model:
        step.return_updated_model = True
```

braingram (commented):
I think these changes would allow removal of the option from the spec. I think of the spec as part of the user-facing API and as @schlafly noted the "return_updated_model" is only something the pipeline is concerned with.

mairanteodoro:
> I think these changes would allow removal of the option from the spec. I think of the spec as part of the user-facing API and as @schlafly noted the "return_updated_model" is only something the pipeline is concerned with.

Thanks, @braingram!

schlafly:
Thanks Mairan.  A few requests:
- Can you move the compare_asdf bit of test_tweakreg down below this test so that we can see that this test passes?  https://github.com/spacetelescope/romancal/blob/main/romancal/regtest/test_tweakreg.py#L69-L88  My fault for the poor ordering and merging these two tests.
- I don't think I expected any pixel values to have changed.  This bit of the report suggests otherwise.  Do you know why pixel values changed?
```
arrays_differ': {"root['roman']['data']": {'abs_diff': <Quantity 269146.1 DN / s>,
'n_diffs': 259,
'worst_abs_diff': {'index': (1257,
2275),
'value': <Quantity 131072. DN / s>},
'worst_fractional_diff': {'index': (3587,
1296),
'value': <Quantity 1.0063206>}},
```

I think I infer from https://github.com/spacetelescope/RegressionTests/actions/runs/9808347547 that we can pass tests at present and so this really is a change due to this PR?  But most normal main scheduled tests fail so it's a bit hard to say.

mairanteodoro:
> * Can you move the compare_asdf bit of test_tweakreg down below this test so that we can see that this test passes?  https://github.com/spacetelescope/romancal/blob/main/romancal/regtest/test_tweakreg.py#L69-L88  My fault for the poor ordering and merging these two tests.

Done.


> * I don't think I expected any pixel values to have changed.  This bit of the report suggests otherwise.  Do you know why pixel values changed?

I believe that the difference in the data array is due to `RomanSourceCatalog.convert_sb_to_flux_density()` that is not recovering the input value after being executed.

For example, when we call `RomanSourceCatalog.convert_l2_to_sb()`, and then run the reverse method (`RomanSourceCatalog.convert_sb_to_l2()` on the output, we should be able to recovery the original data. That's exactly what we see here:

```
--> Before running RomanSourceCatalog.convert_l2_to_sb(): self.model.data[1257,2275] = -1397121220608.0 DN / s  <--+
                                                                                                                   |
self.convert_l2_to_sb()                                                                                            |
                                                                                                                   |
--> After running RomanSourceCatalog.convert_l2_to_sb(): self.model.data[1257,2275] = -464422600704.0 MJy / sr.    | same result, as expected
                                                                                                                   |
self.convert_sb_to_l2()                                                                                            |
                                                                                                                   |
--> After running reverse operation: self.model.data[1257,2275] = -1397121220608.0 DN / s  <-----------------------+
```

That doesn't happen for `RomanSourceCatalog.convert_sb_to_flux_density()`, though:

```
--> Before running RomanSourceCatalog.convert_sb_to_flux_density(): self.model.data[1257,2275] = -464422600704.0 MJy / sr  <--+
                                                                                                                              |
self.convert_sb_to_flux_density()                                                                                             |
                                                                                                                              |        
--> After running RomanSourceCatalog.convert_sb_to_flux_density(): self.model.data[1257,2275] = -130425602048.0 uJy           | should be the same value
                                                                                                                              |
self.convert_flux_density_to_sb()                                                                                             |
                                                                                                                              |
--> After running reverse operation: self.model.data[1257,2275] = -464422567936.0 MJy / sr  <---------------------------------+
```

When we convert `-464422567936.0 MJy/sr` back to `DN/s` (using `RomanSourceCatalog.convert_flux_density_to_sb()`), we get `-1397121089536.0 DN/s`, which, when compared to the input (`-1397121220608.0 DN/s`), explains the observed difference:

`1397121220608.0 - 1397121089536.0 = 131072.0 DN/s`.

schlafly (commented):
This looks good to me.  One comment inline.

@ddavis-stsci , I think this also resolves https://github.com/spacetelescope/romancal/issues/1317 ; want to take a look before merging?

romancal/tweakreg/tweakreg_step.py
```diff
@@ -152,51 +152,95 @@ def process(self, input):
...
-                    long_axis = axis + "centroid"
-                    if long_axis in catalog.colnames:
-                        catalog.rename_column(long_axis, axis)
+        with images:
+            for i, image_model in enumerate(images):
+                if image_model.meta.exposure.type != "WFI_IMAGE":
+                    # Check to see if attempt to run tweakreg on non-Image data
+                    self.log.info("Skipping TweakReg for spectral exposure.")
+                    # Uncomment below once rad & input data have the cal_step tweakreg
+                    # image_model.meta.cal_step.tweakreg = "SKIPPED"
```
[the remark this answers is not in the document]
mairanteodoro: Thanks for catching that; it was not intentional.

ddavis-stsci:
> This looks good to me. One comment inline.
> 
> @ddavis-stsci , I think this also resolves #1317 ; want to take a look before merging?

Yes this appears to be fixed,
From my test run - 
2024-07-24 11:51:13,601 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0001_WFI01_cal.asdf
[stpipe.ExposurePipeline] Saved model in r0000101001001001001_01101_0001_WFI01_cal.asdf
2024-07-24 11:51:14,539 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0003_WFI01_cal.asdf
[stpipe.ExposurePipeline] Saved model in r0000101001001001001_01101_0003_WFI01_cal.asdf
2024-07-24 11:51:15,286 - stpipe.ExposurePipeline - INFO - Saved model in r0000101001001001001_01101_0004_WFI01_cal.asdf
[stpipe.ExposurePipeline] Saved model in r0000101001001001001_01101_0004_WFI01_cal.asdf
2024-07

schlafly (approved):
Go ahead and merge & okify so that regtests pass?  Thanks!

PR #1291, opened 2024-06-28
add upper pin for photutils

braingram:
photutils 1.13.0 is causing test failures and appears to be incompatible with romancal:
https://github.com/spacetelescope/RegressionTests/actions/runs/9716745018/job/26821243423#step:27:2134

I started a CI run with main here which has all the `test_psf` tests failing: https://github.com/spacetelescope/romancal/actions/runs/9717591343/job/26823614739

This PR adds an upper pin for photutils.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


larrybradley:
I took a quick look and the `add_sources` function in `lib/tests/test_psf` used private attributes in `PSFPhotometry` that no longer exist.  There's actually no need to use `PSFPhotometry` in that function; `photutils` has standalone function(s) for creating simulated images from models.  I will fix this function next week.

larrybradley:
Also, it's probably a good idea to add `photutils` to the devdeps CI test job.

braingram:
Thanks.

@ddavis-stsci how about we merge this to fix the tests until @larrybradley updates them for the new version.

The devdeps are testing against photutils main: https://github.com/spacetelescope/romancal/actions/runs/9741383642/job/26880534266#step:10:101

ddavis-stsci (approved):

PR #1297, opened 2024-07-08
Fix issue in constructing skycell WCS.

schlafly:
This PR fixes some issues in the of the skycell WCS.

Formerly, the code used the center pixel of the WCS to mark the edge of the bounding box.  That meant that the bounding box only covered a quarter of the image, leading to issues.

The "shift" used to move from the projection center was also using the edge of the bounding box.  We now use the projection center in x/y coordinates.

This is intended to address an issue that @snbianco raised.  I've placed a new example L3 image at
/grp/roman/eschlafly/misc/r0099101001001001001_F158_visit_r274dp63x31y81_i2d.asdf
that should have this issue fixed, but please check!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,12 @@ stpipe
 - Add ``ModelContainer`` support to ``Step._datamodels_open`` to allow
   loading "pars-*" files from CRDS. [#1270]
 
+mosaic_pipeline
+---------------
+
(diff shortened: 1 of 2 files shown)


ddavis-stsci (approved):
LGTM

snbianco:
This resolves the problem I was running into, thank you!

PR #1301, opened 2024-07-12
use monkeypatch in tweakreg tests to prevent global setting change that causes other tests to fail

braingram:
2 tests in tweakreg change the global "ALIGN_TO_ABS_REFCAT" flag. This can cause later tests run in the same process to fail when those tests expect absolute alignment to occur.
Here's an example run (using romancal main) with 1 of the offending tests and 1 dependent test showing the failure (tweakreg skip) caused by the global change:
https://github.com/spacetelescope/RegressionTests/actions/runs/9911986940

This PR uses `monkeypatch` to avoid changing the global state.

Here is a similar run as above (the 2 tests) with this PR:
https://github.com/spacetelescope/RegressionTests/actions/runs/9912169607
and a full regtest (with this PR) run with no errors:
https://github.com/spacetelescope/RegressionTests/actions/runs/9912060014

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/tweakreg/tests/test_tweakreg.py
+++ b/romancal/tweakreg/tests/test_tweakreg.py
@@ -1038,7 +1038,7 @@ def test_fit_results_in_meta(tmp_path, base_image):
     ]
 
 
-def test_tweakreg_returns_skipped_for_one_file(tmp_path, base_image):
+def test_tweakreg_returns_skipped_for_one_file(tmp_path, base_image, monkeypatch):
     """
     Test that TweakRegStep assigns meta.cal_step.tweakreg to "SKIPPED"
     when one image is provided but no alignment to a reference catalog is desired.
@@ -1047,7 +1047,7 @@ def test_tweakreg_returns_skipped_for_one_file(tmp_path, base_image):
     add_tweakreg_catalog_attribute(tmp_path, img)
 
     # disable alignment to absolute reference catalog
-    trs.ALIGN_TO_ABS_REFCAT = False
+    monkeypatch.setattr(trs, "ALIGN_TO_ABS_REFCAT", False)
     res = trs.TweakRegStep.call([img])
 
     assert all(x.meta.cal_step.tweakreg == "SKIPPED" for x in res)
@@ -1081,7 +1081,7 @@ def test_tweakreg_handles_multiple_groups(tmp_path, base_image):
     )
 
 
-def test_tweakreg_multiple_groups_valueerror(tmp_path, base_image):
+def test_tweakreg_multiple_groups_valueerror(tmp_path, base_image, monkeypatch):
     """
     Test that TweakRegStep throws an error when too few input images or
     groups of images with non-empty catalogs is provided.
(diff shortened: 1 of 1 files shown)


ddavis-stsci:
Why do we need a monkeypatch here. 
I thought that the test was to see if tweakreg will be skipped if the proper catalog for absolute alignment is not provided. Doesn't setting `ALIGN_TO_ABS_REFCAT = False` just skip the alignment no matter what?

braingram:
`ALIGN_TO_ABS_REFCAT` is a global flag in tweakreg:
https://github.com/spacetelescope/romancal/blob/459e8e345bbb1ba6bfd76d0d8a90f2058089791e/romancal/tweakreg/tweakreg_step.py#L42

monkeypatching is required here as there are tests that change but then do not reset this flag. These tests turn off absolute alignment so any test run in the same process after these tests will also have absolute alignment disabled. By monkeypatching the absolute alignment disabling will be temporary only for the test that is using the monkeypatch.

braingram:
I should clarify that this is what's causing the random regression test failures where tweakreg is randomly skipped for `test_level2_image_processing_pipeline` for some runs. The linked runs in the description show a minimal reproducer and fix for the issue.

ddavis-stsci:
My point is that the test should not be setting this flag. 
As written it appears the test only verifies that tweakreg is skipped for the absolute alignment if told to do so. 
The test should verify that if `ALIGN_TO_ABS_REFCAT = True` and the proper catalogs are not available then tweakreg will skip the absolute alignment without resetting the variable.

Tweakreg should never be randomly skipped. The regression tests should be deterministic. If they are not there is another bug we should look for.

braingram:
This PR makes no change to the test logic. It only fixes the non-deterministic nature of the regtests (as seen in the runs linked in the description).

mairanteodoro:
I think `TweakRegStep` needs to be update.
I don't think we need to set `ALIGN_TO_ABS_REFCAT` anymore. For Roman, we always align to an absolute reference catalog (either provided by the user or using the default catalog).

For now, to have the unit tests that force `TweakRegStep` to be skipped working properly, we have two options:
- use the `Step.run()` method instead of `.call()`;
- use monkeypatch, as suggested by @braingram in this PR.

I think the easiest option is the monkeypatch.

mairanteodoro (commented):
Looks good to me. Thanks for the fix @braingram!

braingram:
Thanks! I opened an issue to track the removal of the global flag:
https://github.com/spacetelescope/romancal/issues/1306

ddavis-stsci:
I chatted with @schlafly and @mairanteodoro and we agree the best fix for this issue is to remove these two tests from the unit tests for now. Can you remove these?

braingram:
I'm closing this PR because a different fix is requested. I don't at the moment have time to redo this work. The issue #1306 can track the work for whoever takes it on.

mairanteodoro:
> I'm closing this PR because a different fix is requested. I don't at the moment have time to redo this work. The issue #1306 can track the work for whoever takes it on.

Thanks for pointing this out, @braingram!

PR #1319, opened 2024-07-19
use monkeypatch in patch_match tests to avoid global state change

braingram:
This fixes an issue encountered during testing for https://github.com/spacetelescope/romancal/pull/1222 which revealed that the patch match tests modify global state (impacting later tests). A relatively minimal reproducer (using the #1222 branch) is to run:
```
pytest --bigdata romancal/patch_match/test/test_patch_match.py romancal/regtest/test_mos_pipeline.py::test_hlp_mosaic_pipeline
```
which triggers the `IndexError: index 0 is out of bounds for axis 0 with size 0` error when the mosaic pipeline fails to find the patch table entry for the requested patch. This is due to patch table modification here:
https://github.com/spacetelescope/romancal/blob/747667895099a80ef283838a8499bac391cb1c46/romancal/patch_match/test/test_patch_match.py#L43
(which is why running just the test_hlp_mosaic_pipeline test does not fail as long as the patch match tests were not included in the test search).

This PR moves the patch table modifications into a fixture that is set to autouse (within the patch match tests). Running the above example command (with #1222 rebased on top of the source branch for this PR) works as expected (without the above error).

Regression tests show no errors: https://github.com/spacetelescope/RegressionTests/actions/runs/10013300421

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/patch_match/test/test_patch_match.py
+++ b/romancal/patch_match/test/test_patch_match.py
@@ -24,10 +24,9 @@ of its columns so that subsets of the table selected to reduce the filesize stil
 retain the same index obtained.)
 """
 
-import inspect
-import os
-import os.path
+from pathlib import Path
 
+import asdf
 import astropy.coordinates as coord
 import astropy.modeling.models as amm
 import astropy.units as u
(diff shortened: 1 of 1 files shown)


braingram:
I vote for this going in after https://github.com/spacetelescope/romancal/pull/1241 (unless there is a barrier to approval for #1241).

ddavis-stsci (changes requested):
The patches subset is only for my internal testing and should not be used for the regression tests.

braingram:
> The patches subset is only for my internal testing and should not be used for the regression tests.

I'm confused.

This fixes the error you're getting for https://github.com/spacetelescope/romancal/pull/1222 where the patches subset (which is [included](https://github.com/spacetelescope/romancal/blob/main/romancal/patch_match/test/patches_subset.asdf) in the public unit tests) is causing the new regression test to fail.

What is your suggestion for resolving this?

ddavis-stsci:
The patch file to be used is
/grp/roman/scsb/tesselation/patches.asdf
not the one I hacked for tests.

braingram:
This PR doesn't change that for the regression tests (which are the only tests that can see `/grp`).

I don't know what patch file you "hacked for tests" but this PR has nothing to do with that.

ddavis-stsci:
You don't set the patch table as
PATCH_SUBSET_PATH = Path(__file__).parent / "patches_subset.asdf"
in this PR? The patches_subset was only to test to see if the file 
reading was size dependent.

The patch table should be /grp/roman/scsb/tesselation/patches.asdf

braingram:
Yes this PR sets the python variable `PATCH_SUBSET_PATH ` used in the `patch_match` unit tests to use the patch table subset that was added in https://github.com/spacetelescope/romancal/pull/1161

This has no impact on the environment variable `PATCH_TABLE_PATH` environment variable.

ddavis-stsci (approved):

perrygreenfield (approved):
LGTM.

PR #1327, opened 2024-07-23
provide formatter to log_records

braingram:
Requires https://github.com/spacetelescope/stpipe/pull/171

Regression tests all pass:
https://github.com/spacetelescope/RegressionTests/actions/runs/10063338609/job/27821316132

See the above linked stpipe PR for more details.

During step runs, stpipe records log messages to `Step.log_records`. This can result in steady accumulation of memory as these are `LogRecord` instances that retain references to their arguments. That means that a log message like the following:
```python
log.info("Calibration array: %s", big_array)
```
Will prevent `big_array` from being garbage collected until the `Step` in which the log message occurred can be collected. One such problematic log message exists in stpipe (see the linked PR) which logs the input datamodel for a step.

Within the context of romancal, running the `elp` pipeline with romancal (and stpipe) main results on the `WFI/image/r0000101001001001001_01101_0001_WFI01_uncal.asdf` regression test file results in the following memory usage:
Screenshot 2024-07-23 at 4 01 20 PM
Note that the orange line has a steady upwards trend (with some steps resulting in transient memory peaks).
With this PR (and the stpipe PR) the memory usage peak is similar (this PR does nothing to reduce the memory for the high memory using steps) but the upward trend is removed:
Screenshot 2024-07-23 at 4 08 24 PM

Plan for merging
-----

Since romancal currently depends on stpipe main the merging of this PR and the stpipe PR https://github.com/spacetelescope/stpipe/pull/171 will require some coordination. One possible option is to update this PR to remove the dependency to my branch and then merge this PR immediately after the stpipe PR.

I've opened both PRs for review but please hold off merging until both PRs are ready to go.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "stcal>=1.7.3,<1.8.0",
     # "stcal @ git+https://github.com/spacetelescope/stcal.git@main",
     #"stpipe >=0.5.0",
-    "stpipe @ git+https://github.com/spacetelescope/stpipe.git@main",
+    "stpipe @ git+https://github.com/braingram/stpipe.git@log_records",
     "tweakwcs >=0.8.6",
     "spherical-geometry >= 1.2.22",
     "stsci.imagestats >= 1.6.3",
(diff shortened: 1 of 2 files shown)


ddavis-stsci (approved):
LGTM

braingram:
I rebased this (to pick up changes in romancal main) and switch the `stpipe` pin to `main` (rather than my fork) now that https://github.com/spacetelescope/stpipe/pull/171 is merged.

PR #1331, opened 2024-07-25
pin stpipe < 0.6.1 on 0.15.x release branch

braingram:
See https://github.com/spacetelescope/stpipe/pull/174 for details about the stpipe 0.6.1 release.

The tldr is that it contains breaking changes to fix a memory leak. The changes are compatible with romancal main following https://github.com/spacetelescope/romancal/pull/1327 but as #1327 will not be backported to the released 0.15.x branch this PR adds an upper pin to allow the released versions of romancal to not use stpipe 0.6.1.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "scipy >=1.11",
     "stcal>=1.7.3,<1.8.0",
     # "stcal @ git+https://github.com/spacetelescope/stcal.git@main",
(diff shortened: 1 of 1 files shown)


schlafly (approved):
Thanks!

zacharyburnett (approved):

PR #1339, opened 2024-07-30
Revert "pin stpipe < 0.6.1 on 0.15.x release branch"

braingram:
Reverts spacetelescope/romancal#1331

As there will not be a 0.15.3 release this PR reverts the changes that would have gone into that release.


braingram:
Marking this as draft as noted by @zacharyburnett rebasing will leave a cleaner commit history: https://github.com/spacetelescope/stpipe/pull/175

PR #1340, opened 2024-07-30
remove stcal upper pin

braingram:
Regtests show no failures:
https://github.com/spacetelescope/RegressionTests/actions/runs/10169343686

https://github.com/spacetelescope/stcal/pull/267 introduced breaking changes to 2 functions used from stcal:
- wcs_from_footprints
- update_s_region_imaging

This PR updates the romancal usage to allow removing the upper pin for stcal. At the moment (since the stcal changes are not released) this requires pinning stcal to the development version. When stcal is released a lower pin for the released version can be set.

I marked this PR as `no-changelog-entry-needed` and removed the changelog entry added in #1326 as there is no longer an upper pin (and no release was made with the changes in #1326).

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


schlafly (approved):
Looks good to me.

A few questions: this points to git stcal rather than just removing the upper pin.  That's okay, but do we need to point to the git version?  I think I might have expected instead that we need a new lower pin, probably >=1.8.0?

It also updates tweakwcs and drizzle minimum dependencies, which maybe are required by more recent stcal versions, but it wasn't obvious to me that those were needed here?

schlafly:
Great, makes sense.  Yes, I agree we'll need a new stcal release in the next ~two weeks in order to support this for the romancal delivery, and no, I don't think we need to try extra hard to support multiple versions of stcal.  Let's just make sure @zacharyburnett is on the same page re an stcal release before the Roman build is due.

zacharyburnett:
> Great, makes sense. Yes, I agree we'll need a new stcal release in the next ~two weeks in order to support this for the romancal delivery, and no, I don't think we need to try extra hard to support multiple versions of stcal. Let's just make sure @zacharyburnett is on the same page re an stcal release before the Roman build is due.

sounds good to me; I'll make a new draft release and prepare for `stcal 1.7.3`

PR #1342, opened 2024-07-31
Remove uses of `lazy_load=False`, update step input handling

### LINKED ISSUE #1341: Investigate need for `lazy_load`ing asdf files
Now that files can be loaded with a "lazy_tree" (https://github.com/spacetelescope/roman_datamodels/pull/358) some investigation of the "lazy_load" usage in romancal is worth undertaking:
https://github.com/search?q=repo%3Aspacetelescope%2Fromancal%20lazy_load&type=code

The use of "lazy_load=False" prevents much of the benefits of "lazy_tree" since with "lazy_load=False" all ASDF block data is loaded on the call to `asdf.open`.

braingram:
This PR removes (almost) all uses of `lazy_load=False` (the ones in `make_regtestdata.sh` were left unchanged).

In several places this required changes to how steps handle input. The closed ticket (#1341) more details but briefly:
```python
def process(self, init):
    with rdm.open(init) as model:
        # update model in place
    return model
```
The above typical step process function results in returning a closed model for any step when run with a filename as an input due to:
- rdm.open opening the file
- the existing of the `with` statement closing the file

Any portion of `model` not modified or accessed during the `process` run can result in an error in a later step. For example if the step does not access `model.err`, the `err` array will not be loaded before the input file is closed. Any later step that attempts to access `err` will encounter a closed file and produce an exception.

This PR switch the pattern to:
```python
def process(self, init):
    if isinstance(init, rdm.DataModel):
        model = init
    else:
        model = rdm.open(init)
```
This fixes the issue mentioned above preventing lazy loading and has the added benefit of allowing the steps to run without producing an unnecessary shallow-copy (clone) of the input (when `rdm.open` is called on a `DataModel`).

Regression tests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/10185037326

Closes #1341 

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ general
 
 - Replace ModelContainer with ModelLibrary [#1241]
 
+- Refactor general step input handling to avoid early closing of
+  input files to allow using more lazy loading [#1342]
+
 
 source_catalog
 --------------
--- a/romancal/assign_wcs/assign_wcs_step.py
+++ b/romancal/assign_wcs/assign_wcs_step.py
@@ -29,21 +29,25 @@ class AssignWcsStep(RomanStep):
 
     def process(self, input):
         reference_file_names = {}
-        with rdm.open(input) as input_model:
-            for reftype in self.reference_file_types:
-                log.info(f"reftype, {reftype}")
-                reffile = self.get_reference_file(input_model, reftype)
-                # Check for a valid reference file
-                if reffile == "N/A":
-                    self.log.warning("No DISTORTION reference file found")
-                    self.log.warning("Assign WCS step will be skipped")
-                    result = input_model.copy()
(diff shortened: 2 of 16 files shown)


schlafly:
Great, thanks Brett---looks like it ended up being a fairly straightforward change, albeit with a lot of indentation differences.

The rule is now: if you hand a step a datamodel, you get a datamodel back.  If you hand a step a filename, you get an open datamodel back, which you are responsible for closing or garbage collection eventually closes for you.  Let's let @ddavis-stsci weigh in, but looks good to me, thanks.

ddavis-stsci (approved):
My tests using the elp and mos pipelines pass so I think we're good to go.

PR #1348, opened 2024-08-02
rcal-827 Add reg test for DMS373

### LINKED ISSUE #1218: Add resample to skycell to hlp regression test
_Issue [RCAL-827](https://jira.stsci.edu/browse/RCAL-827) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Add resample to skycell to regression test for hlp

the patch file is located at ████████████████████████████ so that it is not in any individual area for the moment. 

A decision on where this file should be stored will be made and this location should be updated at that point.

ddavis-stsci:
Resolves [RCAL-827](https://jira.stsci.edu/browse/RCAL-827)

Closes #1218 

This PR adds a regression test for resampling data to a skycell (DMS 373). Currently it uses the SCSB skycell definitions but will be changed to the RTB definitions once available. 

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

The passing regression tests are at
https://github.com/spacetelescope/RegressionTests/actions/runs/10219749253
and
https://github.com/spacetelescope/RegressionTests/actions/runs/10220085169

The only issue is that we will need to have 
PATCH_TABLE_PATH=/grp/roman/scsb/tesselation/patches.asdf
added to the romancal.yml file for GitHub actions (https://github.com/spacetelescope/RegressionTests/pull/112) before merging.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Documentation
 general
 -------
 
-- Add regression test for DMS373, mosaic pipeline [#]
+- Add regression test for DMS373, mosaic pipeline [#1348]
 
 - Update the exposure pipeline to accept a roman datamodel as input [#1296]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Documentation
 general
 -------
 
+- Add regression test for DMS373, mosaic pipeline [#]
+
 - Update the exposure pipeline to accept a roman datamodel as input [#1296]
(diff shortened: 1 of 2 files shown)


schlafly (approved):
This looks good to me.  Please update the test ticket to have the file name used here.  I think I would recommend including some sky cell and "valid" language in the log message to make more contact with the words in the test ticket.

PR #1351, opened 2024-08-05
RCAL-885: Updated Sky Background for Schema

### LINKED ISSUE #1350: Update RCAL to use Sky Background Maker Utility
_Issue [RCAL-885](https://jira.stsci.edu/browse/RCAL-885) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Update RCAL to use the roman_datamodel maker utility for sky background. Also add test to ensure that background keywords are populated in individual_image_meta.

PaulHuwe:
Resolves [RCAL-885](https://jira.stsci.edu/browse/RCAL-885)

Closes #1350 

This PR updates sky background usage in code and tests to use maker utilities.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


PaulHuwe:
One regression test failed, in an expected way:
https://github.com/spacetelescope/RegressionTests/actions/runs/10241364423

mairanteodoro (commented):
Looks good to me, @PaulHuwe! Thanks!

romancal/resample/resample.py
```diff
@@ -328,7 +334,13 @@ def resample_many_to_one(self):
...
                     and img.meta.background.level is not None
                 ):
-                    data = img.data - img.meta.background.level
+                    if img.data.unit == img.meta.background.level.unit:
+                        data = img.data - img.meta.background.level
+                    else:
+                        data = img.data - (
+                            img.meta.background.level.value
+                            * img.meta.photometry.conversion_megajanskys
+                        ).to(img.data.unit)
```
schlafly, line 343: Are these bits needed to support unit tests or what?  I think in the real pipeline the first step in converting to MJy/sr so that sky matching should only ever see MJy/sr.  A case where sky matching sees DN / s and then we need to convert to MJy/sr seems like a likely bug.

PaulHuwe:
Regression tests have completely passed.

romancal/resample/resample.py
PaulHuwe: I believe unit tests are what spurred this code. I can remove this and update the tests, if preferred?

schlafly: Yes, ideally this code would just be
https://github.com/spacetelescope/romancal/pull/1351/files#diff-f87241f9a890ac31d6813379a09b6eb7e1052e3ab53039449a8f0df91aa030e1R251
and I worry that any unit changing is likely hiding bugs.  If it's easy to fix the tests so that they work with the right units, that would be great.

And it's not ideal to always work in DN / s---it feels like we should be working in MJy/sr.  But it's especially not ideal for this to be working in mixed units (e.g., sky subtraction in DN/s, but applied to MJy/sr), because it feels like in that case something is definitely screwed up and we're coding around that when in reality we should be raising an error.

PaulHuwe: Code reverted and tests fixed.

schlafly:
This looks good to me, but would you mind saying a few words about why the changes to resample & outlier detection were needed?  I think I would have guessed we could get away with no changes there---what am I missing?

PaulHuwe:
> This looks good to me, but would you mind saying a few words about why the changes to resample & outlier detection were needed? I think I would have guessed we could get away with no changes there---what am I missing?

They need a unit update because the `wfi_image` models they were using were still in DN/s (which is the default when using `maker_utils.mk_level2_image`.

schlafly (approved):
Looks good to me, thanks!

zacharyburnett:
does this also need to be rebased with changes from `main`? (do the recent `main` commits affect this at all?)

zacharyburnett:
regtest running with `PATCH_TABLE_PATH` at https://github.com/spacetelescope/RegressionTests/actions/runs/10289887423

schlafly:
I can do that now, but no, they seem unlikely to be relevant.

zacharyburnett:
merging to support `roman_datamodels 0.21.0`

PR #1352, opened 2024-08-05
Update absolute separation default.

### LINKED ISSUE #1329: Increase `abs_separation` default value in `tweakreg`
The default value for `abs_separation` in romancal (0.1) matches the previous default value in jwst.

However, https://github.com/spacetelescope/jwst/pull/8445 increased this parameter to 1.0 with mention of the need for this to be 1.5x the tolerance (0.7). However a later PR added a slightly different check https://github.com/spacetelescope/jwst/pull/8476 to verify `abs_)separation > sqrt(2) * (abs_)tolerance`.

The default values in romancal could likely use some updating (and a similar check added) as I just ran into the same issue fixed by the jwst PR where running `roman_elp` on an uncal file resulted in `RuntimeError: Number of output coordinates exceeded allocation` which went away when I set `abs_separation` to 1.0.

The failed filed were simulated images shared with me by @schlafly:
```
['r0000101001001001001_01101_0002_WFI10_uncal.asdf' and 'r0000101001001001001_01101_0003_WFI10_uncal.asdf']
```

schlafly:
This PR updates the default absolute separation used in tweakreg in romancal, as discussed in https://github.com/spacetelescope/romancal/issues/1329.

Closes #1329.

If multiple sources within 'tolerance' can be found, tweakreg may fail.  Accordingly, 'separation' should always be larger than 'tolerance.'  The default for tweakreg were not consistent with this requirement, which led to a failure to tweak the WCS in some cases.  This PR updates the default to avoid this case.

Regression tests fail due to very small differences in the stars that tweakreg selects for matching.  This strikes me as okay.  There's room in the future to choose better values for these parameters; they should by a few times larger than the expected uncertainty in the pointing.  But this resolves the crashes @braingram was seeing.

**Checklist**
- [X] ran regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/10251136980


schlafly:
@ddavis-stsci , mind reviewing?  I think we want this to go into B15 so that tweakreg doesn't crash on some SCAs.

braingram (approved):
LGTM. Adding a check similar to what was added in https://github.com/spacetelescope/jwst/pull/8476 might make sense as these are user (and pars file) controllable parameters. Having an informative error seems more useful than the step crashing. I don't have any objection to adding this check in a different PR but I suggest we keep the issue open if that's the plan (or make a new one).

braingram:
Ignore my comment about adding the check. Once this step uses stcal it will inherit the check:
https://github.com/spacetelescope/stcal/blob/809de2398dc3843821d71fc5cba88ae591a16eb9/src/stcal/tweakreg/tweakreg.py#L129

schlafly:
Perfect.  Yeah, I agree that this kind of check makes more sense downstream (really in tweakreg itself?) than in romancal / jwst.

ddavis-stsci (approved):

schlafly:
I okified the regtests and am merging.

PR #1354, opened 2024-08-08
RCAL-853: Apply velocity aberration correction to the WFI WCS

### LINKED ISSUE #1192: Implement velocity aberration in AssignWCS
The Webb pipeline includes a term in the WCS that accounts for differential aberration.  The scale factor comes from this stanza:
https://github.com/spacetelescope/jwst/blob/d57f6337ba1b5a6f14379f236b83f741afffa413/jwst/lib/set_velocity_aberration.py#L35
and this is ultimately incorporated into the WCS here:
https://github.com/spacetelescope/jwst/blob/d57f6337ba1b5a6f14379f236b83f741afffa413/jwst/assign_wcs/nircam.py#L89-L101

The reference is Cox 1997
https://www.stsci.edu/instruments/observatory/PDF/velocity.pdf
which comes from Lorentz transformations (e.g., eq. 15 & 16 here:
https://www.usna.edu/Users/physics/mungan/_files/documents/Publications/TPT44.pdf).

The code implements the velocity aberration as if it were an overall scale factor, equal in the direction of the Earth's motion and perpendicular to it.  It looks to me like that's correct to first order in beta=v/c but has second order correction terms.  I think beta is 10^-4.  My impression from Cox 1997 is that the higher order corrections would be ~0.01 mas, which is sufficiently tiny that we don't care, but it's worth checking my work there.

We are not accounting for velocity aberration in Roman, either in romancal or in romanisim.  Since we're using romanisim to check the romancal astrometry, we would need to introduce the correction into both packages at the same time.  The Webb code looks a little complicated, but I think ultimately it just ends up being (to the first order implementation there) as an additional scale equal to (1 + dot(beta, WFI_CEN_unit))^(-1)*(1/gamma), though the gamma term is 2nd order and could be left out.

Another missing piece is any information about the space craft ephemerides.  We should populate that in the simulations before we start computing this quantity.  I think that ends up being easy if we pretend that the spacecraft were exactly at L2.

stscieisenhamer:
Resolves [RCAL-853](https://jira.stsci.edu/browse/RCAL-853)

Closes #1192

This PR applies the velocity aberration correction to the WFI WCS.

**Checklist**
- [x] updated relevant tests
- [ ] added relevant label(s)


stscieisenhamer:
@schlafly Current PR includes basically a duplicate of the JWST correction, that of applying the single scale as given by `meta.velocity_aberration.scale_factor`.

However, the questions in the issues relate more to whether `scale_factor` is sufficiently close to what `ra_offset`/`dec_offset` are, such that they are indistinguishable, or, more importantly, no systematic effect is introduced using the single value.

Nominally, all these values are being calculated in RSDP. Are we tasked to provide those calculations, and hence also deciding whether the single scale_factor or the individual offsets are to be used in the correction?  Towards this, I am getting some sample data and seeing about making some empirical estimates.

schlafly:
Hi Jonathan,

We will need to compute scale_factor.  I think that's a simple function of the velocity of the spacecraft---some beta or gamma factor?  I found the computation of that factor in the Webb code, and put the link in the issue.

It depends on the speed of the spacecraft, which is more annoying to estimate.  I think the easiest thing to do would be to pretend that the spacecraft were exactly at L2, so that the spacecraft's velocity is exactly the same as the Earth's velocity times (l2_distance_from_sun / earth_distance_from_sun), where I guess I'm ignoring effects from the moon but that seems very small.

I had not appreciated until looking at your code that that the ra_ref and dec_ref correspond to the aberrated v2/v3 rather than the nominal v2/v3.  Those ra/dec_ref come from romanisim at present and so I think it is straightforward to get that right.  But I'm rather attached to the notion than ra_ref, dec_ref corresponds to v2_ref, v3_ref, which I claim this will break.  Is that how things work for Webb?

stscieisenhamer:
The previous comment generated much offline discussion. The summary and path forward is below; let me know if I got it wrong.

Concerning where the computation of `va_scale` occurs: For JWST, this is done in SDP using jwst cal code while the Level 1 file is built. For Roman, this can be done as JWST, or in the `set_telescope_pointing` equivalent.

For JWST, as described, the RA_REF/DEC_REF are the "semi-aberrated" versions. These values are determined in `set_telescope_pointing` and are calculated using the SIAF transformation from V1. V1 has been corrected. However, the SIAF transformation does not take into account the differential aberration, which is why the jwst wcs models add the shift component. These values are never updated in the headers. For Roman, these values can be updated in the `assign_wcs` step, or the `set_telescope_pointing` equivalent can take into account the DVA in the first place. 

Just recently learned (27Aug24) about the plans on back-porting the Level3/GAIA corrected solution to the Level 2 files. Details of that does not matter. Note that this process will invalidate all the wcsinfo keywords.

For the moment, I believe the plan is to leave the `assign_wcs` code as is. I can add the updating of the wcsinfo to ensure that RA_REF/DEC_REF represents what V2_REF/V3_REF is pointing at. Note that this updating will use the WCS object which will also include distortion.

schlafly:
It's fine to compute va_scale in romanisim (when the L1s are produced).

Let's follow Webb for ra/dec ref.  i.e., radec_ref only match v1v2_ref exactly for the boresight; the others don't match up exactly due to differential aberration but they're within a pixel or so.  This is tracked in the WCS but not in the wcsinfo.  I'll add an issue to rad bringing up whether we want to add v2a/v3a_ref which would allow these to match up better, though still not perfectly after tweakreg has been run since we don't update the wcsinfo with tweakreg information.

Re backporting the L2 tweakreg WCSes to the L1 files, we'll end up doing something like providing a pile of L2 WCS files separately from the rest of the L2s (which would include tweakreg, aberration tweaks) or just the tweakreg info plus code that turns the wcsinfo + tweakreg info into a new WCS (basically, repeating assign_wcs, aberration, ...).  I don't think that invalidates the wcsinfo keywords any worse than they are already invalid in the L2s.

Let's not update ra_ref/dec_ref yet and instead just continue following Webb.

romancal/assign_wcs/pointing.py
```diff
@@ -45,3 +45,62 @@ def v23tosky(input_model, wrap_v2_at=180, wrap_lon_at=360):
...
+    va_corr : astropy.modeling.CompoundModel, astropy.modeling.models.Identity
+        A 2D compound model that corrects DVA. If ``va_scale`` is `None` or 1
+        then `astropy.modeling.models.Identity` will be returned.
+
+    """
+    if va_scale is None or va_scale == 1:
+        return Identity(2)
+
+    if va_scale <= 0:
+        raise ValueError("'Velocity aberration scale must be a positive number.")
```
schlafly, line 81: Should we set it to 1 in this case for the moment?  Otherwise we need to rebuild all of the regression files, I think?  We can change this to an error in the future.

stscieisenhamer:
[Regression test](https://github.com/spacetelescope/RegressionTests/actions/runs/11917378739) (I believe)

romancal/assign_wcs/pointing.py
```diff
@@ -1,7 +1,12 @@
+import logging
+
 import numpy as np
-from astropy.modeling.models import RotationSequence3D, Scale
+from astropy.modeling.models import Identity, RotationSequence3D, Scale, Shift
 from gwcs.geometry import CartesianToSpherical, SphericalToCartesian
 
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
schlafly, line 8: We don't do this elsewhere AFAIK; was this intentional?

romancal/assign_wcs/pointing.py
```diff
@@ -45,3 +50,66 @@ def v23tosky(input_model, wrap_v2_at=180, wrap_lon_at=360):
...
+        v3_ref = 0
+
+    if v2_ref == 0 and v3_ref == 0:
+        return va_corr
+
+    # NOTE: it is assumed that v2, v3 angles and va scale are small enough
+    # so that for expected scale factors the issue of angle wrapping
+    # (180 degrees) can be neglected.
+    v2_shift = (1 - va_scale) * v2_ref
+    v3_shift = (1 - va_scale) * v3_ref
```
schlafly, line 109: This looks good to me.  FWIW, I went down this rabbit hole again
https://github.com/spacetelescope/rad/issues/437
which permanently confuses me.  I don't think we want any changes here related to that, but you could confirm whether I'm confused.

schlafly (approved):
Regtests are passing; this can go ahead, thanks.  It won't start actually doing something until we start making files with velocities in romanisim; that will start happening when we have merged https://github.com/spacetelescope/romanisim/pull/162 and make new files.

romancal/assign_wcs/pointing.py
schlafly: I was just confused.  The key point is that dva_corr_model(v2_ref, v3_ref) == (v2_ref, v3_ref), while the shift was telling me that that wasn't true.  But the shift exactly cancels out the motion of the center due to the scale and all is well.

romancal/assign_wcs/pointing.py
stscieisenhamer: Missed this cut and paste error. Removing.

romancal/assign_wcs/pointing.py
stscieisenhamer: As per slack note, this has resolved.

Issue #1264, opened 2024-06-04
Context images have incorrect dimensions


stscijgbot-rstdms:
_Issue [RCAL-855](https://jira.stsci.edu/browse/RCAL-855) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

When running data though the roman_mos pipeline the context images seem to have the wrong dimensionality.

 
```java
 ├─data (Quantity): shape=(9191, 8642), dtype=float32 # Science Data (MJy / steradian)
  ├─context (NDArrayType): shape=(2, 9191, 8642), dtype=uint32 # Context Data
  ├─err (Quantity): shape=(9191, 8642), dtype=float32 # Error Data (MJy / steradian)```
The input asn has eight exposures (WFI01 & WFI02) with three exposures each and detectors WFI10 and WFI11 to make the full 8.

 The number of “planes” is equal to {{(number of input images - 1) // 32 + 1 or,
```java
(8 - 1) // 32 + 1.
1.0```
`For this case but I'm seeing shape=(2,... planes`

The uncal input data is at ██████████████████████████████

The input asn file is attached below.

**Comment by stscijgbot-rstdms** [comment]
_Comment by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis) on [JIRA](https://jira.stsci.edu/browse/RCAL-855?focusedCommentId=710684#comment-710684):_

Using the above asn file and the current version of romancal (0.15.2.dev135+g73e8b77) the context images now have the correct dimentionality,
```java
segm = rdm.open('r0000101001001001001_01101_0003_WFI01_segm.asdf')
segm.data.shape
(4088, 4088)
```
So I'm closing this issue for now.

schlafly:
I'm confused about the file name here; this is a segmentation image from an L2 catalog, rather than the context component of an L3 image?

schlafly:
I went ahead and downloaded https://bytesalad.stsci.edu/ui/repos/tree/General/roman-pipeline/dev/truth/WFI/image/r0099101001001001001_r274dp63x31y81_prompt_F158_i2d.asdf and indeed, the context image is 2 x 4k x 4k, in line with your original report.  Only bits 0, 1, and 2 in the first context plane are ever set.  I think this is still an open issue?

PR #1355, opened 2024-08-08
fix number of context planes

braingram:
roman_datamodels defaults the context plane of the mosaic model to start with 2 planes:
https://github.com/spacetelescope/roman_datamodels/blob/5e996c37648bac7fa0ac3c4d5e0ce5b8ddaeb69a/src/roman_datamodels/maker_utils/_datamodels.py#L155

This is causing an extra context plane. This PR fixes the issue by setting n_images to 1.

Closes https://github.com/spacetelescope/romancal/issues/1264

Regression tests show 2 expected failures (which shows the issue is fixed): https://github.com/spacetelescope/RegressionTests/actions/runs/10303608077

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,6 +60,11 @@ tweakreg
 --------
 - Integration with ``SourceCatalogStep``: allow usage of results from ``SourceCatalogStep``. [#1276]
 
+resample
+--------
+
+- Fix incorrect number of starting planes for context image. [#1355]
+
 mosaic_pipeline
 ---------------
 
--- a/romancal/resample/resample.py
+++ b/romancal/resample/resample.py
@@ -153,8 +153,10 @@ class ResampleData:
         # NOTE: wait for William to fix bug in datamodels' init and then
         # use datamodels.ImageModel(shape=(nx, ny)) instead of mk_datamodel()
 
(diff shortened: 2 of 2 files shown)


schlafly (approved):
Great, thanks Brett, approving.

braingram:
Oooo, the CI picked up a new pars file:
https://roman-crds.stsci.edu/browse/roman_wfi_pars-jumpstep_0001.asdf
I guess that means pars files work :) and those tests are broken...

schlafly:
Yeah, that's "good" I guess?  I suspect this means some mixture of:
- we need to specify something like --disable-pars-files to avoid getting this file
- we need to specify some earlier context (?) that doesn't identify the new files
- we need to specify the rejection threshold in the tests (it looks to me like the primary change is that the detection threshold increases to 500 in the new files from 180 in the previous default, though I think my brief guess looking at the test would have been that the jump amplitude is ~70 sigma (500 / 7), so I wouldn't have originally expected it to be identified in either the new or old default)
- we need to disable the jump detection tests that we don't use.
We can discuss a little in 15 min at the tag up.

Issue #1302, opened 2024-07-14
OutlierDetection: raise exception instead of skipping on invalid input


braingram:
See https://github.com/spacetelescope/romancal/pull/1241/files#r1625005527

When input is invalid (for example, models are not images), raise a useful exception.

I suspect at least the [test_outlier_skips_step_on_exposure_type_different_from_wfi_image](https://github.com/spacetelescope/romancal/blob/459e8e345bbb1ba6bfd76d0d8a90f2058089791e/romancal/outlier_detection/tests/test_outlier_detection.py#L37) will need to be changed to accommodate this change.

PR #1357, opened 2024-08-08
Update Outlier Detection to use stcal

braingram:
Closes https://github.com/spacetelescope/romancal/issues/1302
Closes https://github.com/spacetelescope/romancal/issues/1423
Fixes [RCAL-926](https://jira.stsci.edu/browse/RCAL-926)

This PR updates outlier detection to use common code in stcal. It largely mirrors the code in https://github.com/spacetelescope/jwst/pull/8840

This PR also removes a few unused items from the outlier detection step spec:
- nlow
- nhigh
- grow
- allowed_memory
- kernel_size

And updates the input handling to raise an exception (instead of issuing a warning) for invalid input (as described in https://github.com/spacetelescope/romancal/issues/1302).

The docs updates in this PR reference the step spec rather than specifying defaults in the docs (which are currently out of sync with the code).
Link to page which has most of the docs updates: https://roman-pipeline--1357.org.readthedocs.build/en/1357/roman/outlier_detection/arguments.html

Regression tests all pass https://github.com/spacetelescope/RegressionTests/actions/runs/11114749475

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


CHANGES.rst
```diff
@@ -49,6 +49,11 @@ outlier_detection
 
 - Set ``single=True`` to use ``many_to_many`` when creating median image. [#1260]
 
+- Remove unused arguments to outlier detection. [#1357]
+
+- Update input handling to raise an exception on an invalid input instead
+  of issuing a warning and skipping the step. [#1357]
+
```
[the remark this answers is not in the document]
braingram: Changelog updated to new format. Let me know if this looks good to you.

pyproject.toml
```diff
@@ -28,8 +28,9 @@ dependencies = [
     # "roman_datamodels>=0.21.0,<0.22.0",
     "roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git",
     "scipy >=1.11",
-#    "stcal>=1.8.0,<1.9.0",
-    "stcal @ git+https://github.com/spacetelescope/stcal.git@main",
+    # "stcal>=1.8.0,<1.9.0",
+    # "stcal @ git+https://github.com/spacetelescope/stcal.git@main",
+    "stcal @ git+https://github.com/emolter/stcal.git@JP-3768",
```
braingram, line 33: Temporarily set to source branch for https://github.com/spacetelescope/stcal/pull/292 for testing that (and this) PR.

braingram:
For the small (3 input) association described in more detail https://innerspace.stsci.edu/display/SCSB/Romancal+ModelLibrary+Performance the memory profile of a run of the mosaic pipeline with this PR is:
Screenshot 2024-09-30 at 6 09 56 PM
Outlier detection (roughly the period between 18:03:30 until the dip before 18:04:30
- no longer produces the peak usage for the pipeline run
- consumes 9.6 GB of memory (vs 13.1 in the "post library" run)
- takes ~1 minute (vs 3 minutes in the "post library" run)

braingram:
Pinging @emolter since I can't request you as a reviewer. If you have a chance to look over this PR that would be greatly appreciated!

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+        The weight threshold for masking out low weight pixels.
+
+    save_intermediate_results : bool
+        if True, save the drizzled models and median model to fits.
+
+    make_output_path : function
+        The functools.partial instance to pass to save_median. Must be
+        specified if save_intermediate_results is True. Default None.
+
+    buffer_size : int
```
emolter, line 56: Maybe add a comment about what the buffer size gets set to when None?

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+    example_model = None
+
+    with input_models:
+        for i in range(len(input_models)):
+
+            model = input_models.borrow(i)
+            wht = build_driz_weight(
+                model,
+                # FIXME this was hard-coded to "ivm"
+                weight_type="ivm",
```
emolter, line 158: should this get fixed now, since `weight_type` is getting passed into this function?  Or are you saying that this may change the behavior of the function in a way that is beyond this PR scope?  If the second one, has a ticket been made to ensure this gets fixed soon?

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+def _flag_model_crs(image, median_data, snr):
+    cr_mask = flag_crs(image.data.value, image.err.value, median_data, snr)
+
+    # Update dq array in-place
+    image.dq |= cr_mask * np.uint32(pixel.DO_NOT_USE | pixel.OUTLIER)
+
+    log.info(f"{np.count_nonzero(cr_mask)} pixels marked as outliers")
+
+
+def detect_outliers(
```
emolter, line 236: probably a nitpick, but is the `detect_outliers` function really a utility that belongs in `utils`?

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+    if resample_data:
+        resamp = ResampleData(
+            library,
+            single=True,
+            blendheaders=False,
+            # FIXME prior code provided weight_type when only wht_type is understood
+            # both default to 'ivm' but tests that set this to something else did
+            # not change the resampling weight type. For now, disabling it to match
+            # the previous behavior.
+            # wht_type=weight_type
```
emolter, line 265: same question here

romancal/resample/resample.py
```diff
@@ -191,101 +195,113 @@ def do_drizzle(self):
...
             return self.resample_many_to_one()
 
-    def resample_many_to_many(self):
-        """Resample many inputs to many outputs where outputs have a common frame.
+    def resample_group(self, input_models, indices):
+        """Apply resample_many_to_many for one group
 
-        Coadd only different detectors of the same exposure (e.g. map SCA 1 and
-        10 onto the same output image), as they image different areas of the
-        sky.
```
emolter, line 199: It might be nice to have some indication of what a "group" is in the docstring of `resample_group` also, i.e., a modified version of these lines

romancal/resample/resample.py
```diff
@@ -191,101 +195,113 @@ def do_drizzle(self):
...
+        Parameters
+        ----------
+        input_models : ModelLibrary
 
-        Used for outlier detection
+        indices : list
```
emolter, line 205: Consider adding a one-line description of what the indices correspond to (I realize I didn't do any of these docstring updates for the jwst code either...)

emolter (commented):
Looks great aside from some minor questions and comments Brett!

Are there any regression tests that save intermediate files from the OutlierDetectionStep?  I don't see any regtest updates here, and I know in JWST we kept running into problems with intermediate filenames partially because those lacked test coverage

docs/roman/outlier_detection/arguments.rst
```diff
@@ -55,71 +58,47 @@ behavior of the processing:
...
+``--resample_data``
   Specifies whether or not to resample the input images when performing outlier
   detection.
 
-``--good_bits`` (string, default=0)
+``--good_bits``
   The DQ bit values from the input image DQ arrays that should be considered 'good'
   when creating masks of bad pixels during outlier detection when resampling the data.
   See `Roman's Data Quality Flags
   <https://github.com/spacetelescope/romancal/blob/main/romancal/lib/dqflags.py>`_
```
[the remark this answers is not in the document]
braingram: Thanks for finding that. Would you open a separate issue for this since this PR doesn't change the link.

romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -74,90 +59,50 @@ def process(self, input_models):
...
-        if self.skip:
-            # set meta.cal_step.outlier_detection to SKIPPED
-            with library:
-                for i, model in enumerate(library):
-                    model.meta.cal_step["outlier_detection"] = "SKIPPED"
-                    library.shelve(model, i)
+            def set_skip(model, index):
+                model.meta.cal_step["outlier_detection"] = "SKIPPED"
+
+            list(library.map_function(set_skip))
```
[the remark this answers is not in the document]
braingram: They won't ever both execute due to the `return` on line 68.

romancal/outlier_detection/utils.py
braingram: Good idea. I opened an issue: https://github.com/spacetelescope/romancal/issues/1428

romancal/outlier_detection/utils.py
braingram: I struggled to come up with a way to organize/name this so suggestions are welcome. Putting it in "imaging" (like in jwst) doesn't make sense here (as there's no non-imaging modes). Putting it in another "outlier_detection" submodule is overly redundant ("romancal.outlier_detection.outlier_detection.detect_outliers"). So I gave up and went with the standard "utils".

romancal/outlier_detection/utils.py
braingram: I expanded the issue to include this as well: https://github.com/spacetelescope/romancal/issues/1428

docs/roman/outlier_detection/arguments.rst
```diff
@@ -55,71 +58,47 @@ behavior of the processing:
...
-  not defined, no check is made. If the resampled image would be larger than specified,
-  an ``OutputTooLargeError`` exception will be generated. For example, if set to
-  ``0.5``, only resampled images that use less than half the available memory can be
-  created.
-
-``--in_memory`` (boolean, default=False)
+``--in_memory``
   Specifies whether or not to keep all intermediate products and datamodels in
   memory at the same time during the processing of this step.  If set to `False`,
   all input and output data will be written to disk at the start of the step
```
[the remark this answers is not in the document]
braingram: Good question. It looks like more of this description is out-dated. How does 0a5325e look?

romancal/outlier_detection/tests/test_outlier_detection.py
```diff
@@ -37,7 +35,7 @@ def test_outlier_skips_step_on_invalid_number_of_elements_in_input(base_image):
             res.shelve(m, i, modify=False)
 
 
-def test_outlier_skips_step_on_exposure_type_different_from_wfi_image(base_image):
+def test_outlier_raises_exception_on_exposure_type_different_from_wfi_image(base_image):
     """
     Test if the outlier detection step is skipped when the exposure type is different from WFI image.
```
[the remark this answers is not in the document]
braingram: Updated in bc7b840. Thanks again.

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+    good_bits : int
+        The bit values that are considered good when determining the
+        data quality of the input.
+
+    save_intermediate_results : bool
+        if True, save the models and median model to fits.
+
+    make_output_path : function
+        The functools.partial instance to pass to save_median. Must be
+        specified if save_intermediate_results is True. Default None.
```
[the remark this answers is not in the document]
braingram: Thanks! I removed the description of the non-existent default in e6ca79e

romancal/outlier_detection/utils.py
```diff
@@ -0,0 +1,314 @@
...
+
+    maskpt : float
+        The weight threshold for masking out low weight pixels.
+
+    save_intermediate_results : bool
+        if True, save the drizzled models and median model to fits.
+
+    make_output_path : function
+        The functools.partial instance to pass to save_median. Must be
+        specified if save_intermediate_results is True. Default None.
```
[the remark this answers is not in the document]
braingram: Thanks! I removed the mention of the default in e6ca79e

romancal/outlier_detection/utils.py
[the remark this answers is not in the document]
braingram: Updated in e3f7956

romancal/resample/resample.py
braingram: How does b8ba8af look?

docs/roman/outlier_detection/arguments.rst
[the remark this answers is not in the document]
emolter: https://github.com/spacetelescope/romancal/issues/1429

romancal/resample/resample.py
braingram: I held off defining "group" in b8ba8af and instead reworded this to say it resamples the models at the provided indices. "many_to_many" still has this "group" description.

docs/roman/outlier_detection/arguments.rst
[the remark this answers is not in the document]
emolter: looks good

romancal/outlier_detection/utils.py
emolter: looks good

romancal/resample/resample.py
emolter: looks good except (nitpick) the second instance of "and will not" could be changed to "nor"

romancal/resample/resample.py
emolter: looks good

romancal/outlier_detection/utils.py
emolter: It's not that many lines.  Is there a reason why it needs to be its own function, given that there is only one outlier detection mode for romancal?  That is, can you just embed the code into `OutlierDetectionStep`?

braingram: I considered that. I believe the convention for romancal (as described by @mairanteodoro ) is to have the steps handle the stpipe-specific stuff and s much of the "meat" of the processing occur in a separate function/class/utility. If it's ok with you let's see what he thinks.

romancal/resample/resample.py
braingram: I'm going to leave this as-is in part because I can't think of a good way to fit in a "nor". I know it's considered a universal gate so it should be able to do anything (bad joke)!

romancal/outlier_detection/utils.py
emolter: sounds good it's definitely nbd either way as far as I'm concerned

romancal/resample/resample.py
emolter: sounds good

emolter (approved):
All of my comments have been answered.  I'm still wondering though whether there is test coverage of the intermediate files somewhere in the regtests.

braingram:
> All of my comments have been answered. I'm still wondering though whether there is test coverage of the intermediate files somewhere in the regtests.

Thanks for taking a look and for the reminder on that point. There's a unit test that saves intermediate results:
https://github.com/spacetelescope/romancal/pull/1357/files#diff-efb2f2b9bafb3583972a4d54071f5cf19254a042c426746575536554269d1290R115

schlafly (approved):
This looks good to me.  A few questions, though:
- FWIW, either accepting weight_type as ivm with a default of IVM or passing it through correctly works fine for me.  We do want to be using that but we don't need to do that through hard coding.  This is especially true for the _median_without_resampling case, which we don't expect to use much in production.  
- It's great to move the common code to stcal and to improve performance.  It looks to me like it's the case that the new outlier_detection.utils module contains no Roman-specific code, except maybe through the ModelLibrary infrastructure.  Is that a target of future stcal common code consolidation?

braingram:
Thanks!

> This looks good to me. A few questions, though:
> 
>     * FWIW, either accepting weight_type as ivm with a default of IVM or passing it through correctly works fine for me.  We do want to be using that but we don't need to do that through hard coding.  This is especially true for the _median_without_resampling case, which we don't expect to use much in production.

Sorting out the weight handling sounds great. I opened an issue to track that https://github.com/spacetelescope/romancal/issues/1428 but I think it would be a little more involved. I don't expect we'll see regtest differences but I do think adding some unit tests (to verify that changing the weight type has the intended effect) is useful.

>     * It's great to move the common code to stcal and to improve performance.  It looks to me like it's the case that the new outlier_detection.utils module contains no Roman-specific code, except maybe through the ModelLibrary infrastructure.  Is that a target of future stcal common code consolidation?

Good point! I think there is more that could be done here once resample is moved to stcal. At the moment the link between outlier detection and resample puts some limitations on what can be pulled out of jwst/romancal.

zacharyburnett (approved):
more abstraction! the changelog looks good to me

PR #1360, opened 2024-08-09
Populate sky metadata with zero if matching is not possible.

schlafly:
Make sure that we send sensible metadata even when the L2 images don't overlap one another.  The existing code populates Nones here, which don't fit into the individual_image_metadata tables with the other numbers.

Regression tests are passing: https://github.com/spacetelescope/romancal/pull/1355

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,11 @@ tweakreg
 
 - Update default absolute separation for tweakreg.  [#1352]
 
+skymatch
+--------
+- Populate valid metadata even when then are no overlapping images to
+  match [#1360]
+
 
 0.15.1 (2024-05-15)
 ===================
--- a/romancal/skymatch/skymatch_step.py
+++ b/romancal/skymatch/skymatch_step.py
@@ -181,7 +181,7 @@ class SkyMatchStep(RomanStep):
         image.meta.background.subtracted = self.subtract
         image.meta.background.level = sky
 
-        if step_status == 'COMPLETE' and self.subtract:
+        if step_status == "COMPLETE" and self.subtract:
             image.data[...] = sky_image.image[...]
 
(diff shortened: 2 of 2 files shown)


ddavis-stsci:
This fixes the issue I was seeing,

rerun tests
strun --debug roman_mos r0000101001001001001_r274dp63x32y80_visit_F158_prompt_i2d_asn.json
2024-08-12 10:20:41,812 - stpipe.MosaicPipeline - INFO - MosaicPipeline instance created.
...
2024-08-12 10:32:58,174 - stpipe.MosaicPipeline - INFO - Saved model in r0000101001001001001_r274dp63x32y80_visit_F158_prompt_i2d.asdf
[stpipe.MosaicPipeline] Saved model in r0000101001001001001_r274dp63x32y80_visit_F158_prompt_i2d.asdf
2024-08-12 10:32:58,174 - stpipe.MosaicPipeline - INFO - Step MosaicPipeline done
[stpipe.MosaicPipeline] Step MosaicPipeline done

asdfattr.py r0000101001001001001_r274dp63x32y80_visit_F158_prompt_i2d.asdf cal_step
attributes:  cal_step
ex_list:  ['cal_step']
Meta Attribute  cal_step
                flux	COMPLETE
   outlier_detection	COMPLETE
            skymatch	COMPLETE
            resample	COMPLETE

Do we need this in for the release. My take is that it would be nice but not needed.

schlafly:
Regression tests are passing.  https://github.com/spacetelescope/RegressionTests/actions/runs/10365669711

It doesn't address a specific requirement and is in that sense probably optional, but I don't see why we wouldn't include it.  Likewise I think we should get in https://github.com/spacetelescope/romancal/pull/1355, as it fixes a bug, but isn't critical.  The latter requires some okify'ing, but only to get rid of the previous problematic extra context planes.

PR #1367, opened 2024-08-15
only load patch table when needed

### LINKED ISSUE #1366: Only load patch table when needed
Currently the patch table is loaded whenever romancal is loaded, including for things like the ELP which do not use the patch table.  The loading occurs whenever the module is imported.
https://github.com/spacetelescope/romancal/blob/main/romancal/patch_match/patch_match.py#L255

I believe the module is loaded whenever either pipeline is used, since the pipeline module pulls in the names of both steps.
https://github.com/spacetelescope/romancal/blob/b24429c68c5841f09877ad564a67fb6afb8f425a/romancal/pipeline/__init__.py#L6-L7
We should instead only load the patch table when something actually uses it.  I noticed this at the moment since I have a slow connection to /grp and couldn't figure out why it was taking a half hour to do something like strun roman_elp -h; unsetting PATCH_TABLE_PATH led this to instead only take a few seconds.

braingram:
This PR modifies the mosaic pipeline and patch match submodule to only load the patch table when required (instead of at import).

Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/10404395165

Closes #1366

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+0.16.3 (unreleased)
+===================
+
+mosaic_pipeline
+---------------
+
+- Only load patch table when needed. [#1367]
+
 0.16.2 (2024-08-23)
 ===================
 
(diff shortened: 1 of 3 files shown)


schlafly (approved):
Thanks Brett, that looks good to me!

PR #1373, opened 2024-08-20
RCAL-895: allow updating source catalog with tweaked WCS when running ELP

mairanteodoro:
Resolves [RCAL-895](https://jira.stsci.edu/browse/RCAL-895)

This PR implements the update of the source catalog coordinates using the improved WCS calculated by `TweakRegStep` when processing data through the ELP.

When running the ELP, at the end of the `TweakRegStep`, the `cat` file produced by `SourceCatalogStep` (containing all the sources coordinates) will be read in, updated with the new WCS function, and saved back to disk (overwriting the file content).

**Regression tests**
All tests are passing: https://github.com/spacetelescope/RegressionTests/actions/runs/11034994349

**Checklist**
- [x] updated relevant tests
- [X] added relevant label(s)


romancal/pipeline/exposure_pipeline.py
```diff
@@ -71,6 +71,8 @@ def process(self, input):
 
         # make sure source_catalog returns the updated datamodel
         self.source_catalog.return_updated_model = True
+        # make sure we update source catalog coordinates afer running TweakRegStep
+        self.tweakreg.update_source_catalog_coordinates = True
```
nden, line 75: It looks like this attribute is set here and only here. Patching the step object like this is not a good idea.
It should be added to the step spec with a default value.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1185,3 +1186,127 @@ def test_parse_catfile_raises_error_on_invalid_content(tmp_path, catfile_line_co
...
+    cat_ra_psf = cat.source_catalog["ra_psf"]
+    cat_dec_psf = cat.source_catalog["dec_psf"]
+
+    # calculate world coords using tweaked WCS
+    expected_centroid = img.meta.wcs.pixel_to_world(
+        cat.source_catalog["xcentroid"], cat.source_catalog["ycentroid"]
+    )
+    expected_psf = img.meta.wcs.pixel_to_world(
+        cat.source_catalog["x_psf"], cat.source_catalog["y_psf"]
+    )
```
nden, line 1240: Here again, I don't think there's reason to use the WCS API and do additional object conversions. The legacy interface should be fine.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1185,3 +1186,127 @@ def test_parse_catfile_raises_error_on_invalid_content(tmp_path, catfile_line_co
...
+    expected names, adds mock PSF coordinates, applies random shifts to the centroid
+    and PSF coordinates, and calculates the world coordinates for the centroids.
+    """
+    cat_model = rdm.SourceCatalogModel
+    source_catalog_model = maker_utils.mk_datamodel(cat_model)
+    # this will be the output filename
+    source_catalog_model.meta.filename = "img_1.asdf"
+
+    # read in the mock table
+    source_catalog = Table.read("img_1", format="ascii.ecsv")
```
nden, line 1265: Are we writing out ecsv files? If not should it read an asdf file?

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1185,3 +1186,127 @@ def test_parse_catfile_raises_error_on_invalid_content(tmp_path, catfile_line_co
...
+    )
+    expected_psf = img.meta.wcs.pixel_to_world(
+        cat.source_catalog["x_psf"], cat.source_catalog["y_psf"]
+    )
+
+    # compare coordinates (make sure tweaked WCS was applied to cat file coords)
+    np.testing.assert_array_equal(cat_ra_centroid, expected_centroid.ra.value)
+    np.testing.assert_array_equal(cat_dec_centroid, expected_centroid.dec.value)
+    np.testing.assert_array_equal(cat_ra_psf, expected_psf.ra.value)
+    np.testing.assert_array_equal(cat_dec_psf, expected_psf.dec.value)
```
schlafly, line 1246: Thanks Mairan.  The one thing I would like to see is some test that the new coordinates are different from the old coordinates.  The tests here test that the new ra/dec are consistent with the x/y coordinates and the new WCS, but we also sort of want to know that we "did something."  i.e., that because the new WCS is different from the old WCS, we have changed the positions.  I think the easiest thing is to just have some asserts that the new positions are different from the old positions at some tolerance.

schlafly (commented):
Thanks Mairan, looks, good, one comment inline.

CHANGES.rst
```diff
@@ -1,9 +1,12 @@
 0.16.2 (unreleased)
 ===================
 
+exposure_pipeline
+-----------------
+- Update source catalog file with tweaked coordinates. [#1373]
+
```
zacharyburnett, line 7:
```suggestion
```

now that #1375 is merged (switching change log handling to `towncrier`) this change log entry should be a file in `changes/` instead:
```shell
echo "Update source catalog file with tweaked coordinates." > changes/1373.general.rst
```

There currently isn't an `exposure_pipeline` change type but we can definitely add one to `pyproject.toml` if that's desired

(this is just because #1375 was merged while active PRs existed; new PRs will have these instructions in their checklist)

zacharyburnett (changes requested):

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Done. We added a test that checks for changes in the file content making sure that the coordinates are not too far off from each other.

CHANGES.rst
mairanteodoro: Done. Thanks, @zacharyburnett!

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1232,18 +1234,80 @@ def test_update_source_catalog_coordinates(tmp_path, base_image):
...
+        cat_updated.source_catalog["dec_centroid"],
+    )
+    np.testing.assert_allclose(
+        cat_original.source_catalog["ra_psf"],
+        cat_updated.source_catalog["ra_psf"],
+    )
+    np.testing.assert_allclose(
+        cat_original.source_catalog["dec_psf"],
+        cat_updated.source_catalog["dec_psf"],
+    )
```
schlafly, line 1310: Can you help me with this test?  The test is that the files are not bitwise identical, but the ras and decs didn't change meaningfully?  I was kind of expecting this test to show that the ras and decs _do_ change, and I read this test as showing the opposite; am I missing something?

I don't know what precision works, but I'd drop the filecmp.cmp test, and maybe make the tests look like:
```
for column in [ra, dec, centroid, psf]:
    assert np.allclose(orig.catalog[column], new.catalog[column], atol=1/60/60)  # assert nothing moved by more than an arcsecond
    assert not np.allclose(orig.catalog[column], new.catalog[column], atol=1/60/60/100)  # assert some things moved by at least a hundredth of an arcsecond
```
I'm kind of guessing what the right precisions are---it depends on how many stars are in the catalog, and how much you shifted their positions.  But we want to be showing that we did tweak the positions, and I feel like the current test is kind of showing the opposite?

Note 1e-7 rtol corresponds to roughly 0.1" or 1 pixel.  Do you know how big of a tweak you're applying?

romancal/pipeline/exposure_pipeline.py
schlafly: Mairan, Nadia and I chatted briefly about this earlier today, and I had gotten confused with the related self.source_catalog.return_updated_model above.  Let's make this default False for now, but it could go into the spec I think?

romancal/tweakreg/tests/test_tweakreg.py
schlafly: In a separate conversation Nadia mentioned that she was confused here since we otherwise use asdf files for catalogs.  I was also confused here (sorry Nadia).  But I think it's important that these CSV catalogs used in the tests pre-date our asdf source catalogs; this PR is re-using testing machinery there.  We could retire support for this kind of catalog but on the other hand I think we still want it for custom catalogs, etc..  So I'm okay with using these in these tests.

romancal/pipeline/exposure_pipeline.py
mairanteodoro: Just making sure we're on the same page, we can keep the line above in the ELP code, but we should add `update_source_catalog_coordinates` to `TweakRegStep` and set its default to `False`, correct?

schlafly: Yes, add update_source_catalog_coordinates to the tweakreg step spec with a default value of False, exactly.

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: The lines above check that the tweaked coordinates are not too far off (they should be within 1 pixel, since the set up of the mock catalog adds random shifts between -0.5 and 0.5 pixel). We have 22 sources in the catalog. I set `atol` and `rtol` to correspond to roughly half a pixel.

The test to prove that things did move was indeed missing and has been added in the updates to this PR. We are checking if things have moved by at least 1/100th of a pixel.

CHANGES.rst
```diff
@@ -21,7 +21,6 @@ resample
 
 pipeline
 --------
-
```
braingram, line 24: This is causing the CI failure. There's no need to change this file anymore so the CI checks that no commits touch it.

mairanteodoro: Thanks, Brett!

schlafly (approved):
Thanks, this looks good to me.  And it looks like regtests were passing last week; I presume that's still the case?

Maybe just mentioning this aloud for the moment, I originally expected regtests to fail, since we should be changing the values of the catalog entries.  But we don't do compare_asdf against the catalog files (reasoning that it's extra hard to guarantee they don't change), and so we don't expect this change to lead to regtest differences.

mairanteodoro:
> Thanks, this looks good to me. And it looks like regtests were passing last week; I presume that's still the case?

Yes, after fixing the "second okify issue" that you mentioned today in the standup meeting, the regtests for this PR are still passing: https://github.com/spacetelescope/RegressionTests/actions/runs/11034994349

zacharyburnett (approved):

PR #1377, opened 2024-08-22
remove MultilineLogger, split logs with `\n` into multiple messages

braingram:
I marked this PR as `no-changelog-entry-needed` as I don't believe it changes any public API.

The use of `MultilineLogger` is problematic as it causes all later created loggers to use this class (which is not 100% compatible with all logging usage). As the main thing it achieves is splitting log messages that contain newlines into multiple lines this PR also switches all log messages (that I was able to find) that contain a newline into multiple log messages. Note that since `MultilineLogger` is registered during an import it is unpredictable which loggers become instances of `MultilineLogger` (and which are the default `Logger`) so it's likely this PR introduces a small change in the log output for some modified log messages.

Although unknown at this point, `MultilneLogger` may be contributing to the issues currently seen with the `metrics_logger` (see https://jira.stsci.edu/browse/SCSB-174)

Regression tests running at https://github.com/spacetelescope/RegressionTests/actions/runs/10514305515

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/romancal/associations/association.py
+++ b/romancal/associations/association.py
@@ -197,7 +197,8 @@ class Association(MutableMapping):
         try:
             jsonschema.validate(asn_data, asn_schema)
         except (AttributeError, jsonschema.ValidationError) as err:
-            logger.debug("Validation failed:\n%s", err)
+            logger.debug("Validation failed:")
+            logger.debug("%s", err)
             raise AssociationNotValidError("Validation failed") from err
         return True
 
--- a/romancal/associations/generate.py
+++ b/romancal/associations/generate.py
@@ -94,7 +94,7 @@ def generate(pool, rules, version_id=None, finalize=True):
         logger.debug("New process lists: %d", total_reprocess)
         logger.debug("Updated process queue: %s", process_queue)
         logger.debug("# associations: %d", len(associations))
-        logger.debug("Seconds to process: %.2f\n", timer() - time_start)
+        logger.debug("Seconds to process: %.2f", timer() - time_start)
 
     # Finalize found associations
     logger.debug("# associations before finalization: %d", len(associations))
--- a/romancal/associations/lib/log_config.py
+++ b/romancal/associations/lib/log_config.py
@@ -3,7 +3,6 @@
 import logging
 import sys
 from collections import defaultdict
-from functools import partialmethod
 from logging.config import dictConfig
 
 __all__ = ["log_config"]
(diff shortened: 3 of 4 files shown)


ddavis-stsci:
We should ask @stscieisenhamer what problem the MultilineLogger was put in to solve. Since it is in the associations I'm assuming that he will know.

ddavis-stsci:
removing the MultilineLogger package from the associations results in all 29 DMS test numbers in the log file for the regression tests. 
All the unit tests passed and I see the regression tests pass so I think this change is good to go.

braingram:
Thanks for testing this. Which numbers are you referring to? Specifically, what's the log message for DMS413?

ddavis-stsci:
2186:2024-08-23 10:56:13,773 - stpipe.ExposurePipeline - INFO - DMS413 MSG: Testing that result is a Level 2 model.......Pass
2187:[stpipe.ExposurePipeline] DMS413 MSG: Testing that result is a Level 2 model.......Pass
2188:2024-08-23 10:56:13,773 - stpipe.ExposurePipeline - INFO - DMS413 MSG: Testing that there are 16 resultants in the input file.......Pass
2189:[stpipe.ExposurePipeline] DMS413 MSG: Testing that there are 16 resultants in the input file.......Pass
2190:2024-08-23 10:56:13,773 - stpipe.ExposurePipeline - INFO - DMS413 MSG: Testing that there are 16 resultants listed in the output file.......Pass
2191:[stpipe.ExposurePipeline] DMS413 MSG: Testing that there are 16 resultants listed in the output file.......Pass
2192:[metrics_logger] METRIC - DMS413 - test_16resultants_image_processing - PASS
(

braingram:
Thanks! I was recently looking at the code that parses this and I believe only the `2192:[metrics_logger] METRIC - DMS413 - test_16resultants_image_processing - PASS` line "works" for the parser. Is that your understanding as well?

ddavis-stsci:
> Thanks! I was recently looking at the code that parses this and I believe only the `2192:[metrics_logger] METRIC - DMS413 - test_16resultants_image_processing - PASS` line "works" for the parser. Is that your understanding as well?

I have not really looked at the metrics logger code. I can if needed...

braingram:
> I have not really looked at the metrics logger code. I can if needed...

Thanks for the offer. Hopefully that will not be needed. I don't think this PR will fully fix the log parsing issues but I think it's more a sign that using the logs for testing DMS requirements is problematic (given the state of logging in the pipeline). @zacharyburnett has an alternative solution in the works so hopefully things will change (for the better) soon.

stscieisenhamer (approved):
It was a solution to making messages all have a logging tag. If it is a problem, then it can be removed.

PR #1391, opened 2024-08-28
Populate segmentation model metadata for exposures and mosaics.

### LINKED ISSUE #1389: Segmentation map metadata is not getting propagated from L2 file
The following lines make a default segmentation map metadata and don't  bring in any of the metadata from the L2 file.  We need to copy over the relevant L2 metadata into the segmentation metadata.  The source catalog metadata I think does have the right bits, so this might look something like segmentation_model.meta = source_catalog_model.meta.
https://github.com/spacetelescope/romancal/blob/26d03937dfc7d77b6dc5ba0dbf64809403a78f47/romancal/source_catalog/source_catalog_step.py#L149-L159

schlafly:
Closes #1389 

We were previously populating segmentation image metadata with filler values that were not related to the input L2 or L3 image.  This PR adds code that copies the metadata from the catalogs into the segmentation images.

This now also fixes some file naming issues; the mosaic pipeline was saving the catalogs, as well as the step saving the catalogs, leading to two copies of the catalogs with different file names.  Now only the step saves the catalogs.  I have also changed some of the logic around the construction of the file names.

Regtests running here: https://github.com/spacetelescope/RegressionTests/actions/runs/10600468693/job/29383387213

**Checklist**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,11 @@ source_catalog
 
 - Populate segmentation image metadata. [#1391]
 
+resample
+--------
+
+- Use association product name for output meta.filename by default [#1391]
+
 0.16.2 (2024-08-23)
 ===================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ mosaic_pipeline
 
 - Only load patch table when needed. [#1367]
 
+source_catalog
+--------------
+
+- Populate segmentation image metadata. [#1391]
+
 0.16.2 (2024-08-23)
 ===================
 
--- a/romancal/outlier_detection/outlier_detection_step.py
+++ b/romancal/outlier_detection/outlier_detection_step.py
@@ -152,7 +152,7 @@ class OutlierDetectionStep(RomanStep):
                     intermediate_files_suffixes = (
                         "*blot.asdf",
                         "*median.asdf",
-                        f'*{pars.get("resample_suffix")}*.asdf',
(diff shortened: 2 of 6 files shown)


schlafly:
Regtests are now passing: https://github.com/spacetelescope/RegressionTests/actions/runs/10600468693/job/29383387213

ddavis-stsci (changes requested):
I see one issue with this so far. If I run the code with a json file I get the expected three files, 
r0099101001001001001_r274dp63x31y81_prompt_F158_cat.asdf
r0099101001001001001_r274dp63x31y81_prompt_F158_i2d.asdf
r0099101001001001001_r274dp63x31y81_prompt_F158_segm.asdf

If I change the product name and rerun the pipeline I also get the expected three files,
r0099101001001001001_user_choice_cat.asdf
r0099101001001001001_user_choice_i2d.asdf
r0099101001001001001_user_choice_segm.asdf

the issue is that this seems to remove the i2d file from the run before and I now see only
>ls -1 r0099101001001001001_r274dp63x31y81_prompt_F158*.asdf
r0099101001001001001_r274dp63x31y81_prompt_F158_cat.asdf
r0099101001001001001_r274dp63x31y81_prompt_F158_segm.asdf

I don't think we should be deleting files from a previous run.

schlafly:
Great.  regtests also passed, so I'm merging.
https://github.com/spacetelescope/RegressionTests/actions/runs/10600468693/job/29425009213

PR #1395, opened 2024-09-04
Rcal 896 update tweakregstep to use common code from stcal

mairanteodoro:
Resolves [RCAL-896](https://jira.stsci.edu/browse/RCAL-896)

The `TweakRegStep` class has been refactored to improve the handling of image alignment processes. This is the first step (long journey ahead) towards making the class more maintainable and more SOLID-compliant.

The refactoring includes the removal of unused imports, restructuring of methods for better readability, and the introduction of new helper functions to manage catalog parsing and image alignment. The test suite in `test_tweakreg.py` has been updated to reflect these changes, with several tests removed or modified to align with the new implementation.

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
+                    # remove source catalog
+                    del image_model.meta["tweakreg_catalog"]
+
+                    # retrieve fit status and update wcs if fit is successful:
+                    if "SUCCESS" in imcat.meta.get("fit_info")["status"]:
+                        # Update/create the WCS .name attribute with information
+                        # on this astrometric fit as the only record that it was
+                        # successful:
+
+                        # NOTE: This .name attrib agreed upon by the JWST Cal
```
braingram, line 262:
```suggestion
                        # NOTE: This .name attrib agreed upon by the JWST Cal
```
Is this comment still relevant?

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
+        Parameters
+        ----------
+        imcats : list
+            A list of image catalogs containing source information for alignment.
+
+        Returns
+        -------
+        None
+        """
+        tweakreg.relative_align(
```
braingram, line 386: Nice! This method helps to hide the mapping between the step spec and stcal. Good idea!

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -627,96 +628,6 @@ def test_tweakreg_updates_group_id(tmp_path, base_image):
...
-        (0, 3, 5, True),
-        (1, 1, 5, True),
-        (5, 5, (5**2 + 5**2) ** 0.5, True),
-        (5, 5, 5, False),
-        (5, 5, 1, False),
-        (5, 5, 3, False),
-        (5, 10, 5, False),
-    ],
-)
-def test_tweakreg_correction_magnitude(
```
braingram, line 643: Is this no longer needed? I think it's checking if small wcs corrections are correctly identified. This looks to be handled by stcal: https://github.com/spacetelescope/stcal/blob/c3e133228528f369eb85d34c81e2ceb4c76eaaff/src/stcal/tweakreg/tweakreg.py#L101C20-L101C44
If this is correct, no objection from me for removing this.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -627,96 +628,6 @@ def test_tweakreg_updates_group_id(tmp_path, base_image):
...
-            [
-                "l1-270-66-gaia-2016-sca1_cal.asdf",
-                "l1-270-66-gaia-2016-sca2_cal.asdf",
-                "l1-270-66-gaia-2016-sca3_cal.asdf",
-            ],
-            "l1-270-66-gaia-2016-sca",
-        ),
-    ),
-)
-def test_tweakreg_common_name(filename_list, expected_common_name, request):
```
braingram, line 684: I think this one is no longer needed since the library handles the grouping (there's no longer a need for `_common_name`). Is that right?

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -727,6 +638,8 @@ def test_tweakreg_common_name_raises_error_on_invalid_input(filename_list):
 )
 def test_tweakreg_save_valid_abs_refcat(tmp_path, abs_refcat, request):
     """Test that TweakReg saves the catalog used for absolute astrometry."""
+    os.chdir(tmp_path)
```
braingram, line 641: Is this needed? Will the step respect `output_dir` for writing the catalog?

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -748,6 +658,8 @@ def test_tweakreg_save_valid_abs_refcat(tmp_path, abs_refcat, request):
 )
 def test_tweakreg_defaults_to_valid_abs_refcat(tmp_path, abs_refcat, request):
     """Test that TweakReg defaults to DEFAULT_ABS_REFCAT on invalid values."""
+    os.chdir(tmp_path)
```
braingram, line 661: Same as above.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1023,96 +912,6 @@ def test_tweakreg_handles_multiple_groups(tmp_path, base_image):
             res.shelve(r, modify=False)
 
 
-@pytest.mark.parametrize(
-    "column_names",
-    [("x", "y"), ("xcentroid", "ycentroid")],
-)
-def test_imodel2wcsim_valid_column_names(tmp_path, base_image, column_names):
```
braingram, line 1030: Are these still useful? I think romancal is still handling the column renaming (although it looks like it could be moved to stcal).

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
-
-        # Build the catalogs for input images
+        # build the catalogs for input images
+        imcats = []
         with images:
             for i, image_model in enumerate(images):
-                if image_model.meta.exposure.type != "WFI_IMAGE":
-                    # Check to see if attempt to run tweakreg on non-Image data
+                exposure_type = image_model.meta.exposure.type
+                if exposure_type != "WFI_IMAGE":
```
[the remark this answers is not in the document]
mairanteodoro: The current implementation loops over the input and if the exposure type is a non-WFI_IMAGE we set `meta.cal_step` to `"SKIPPED"` and then returns a `ModelLibrary` object.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
-                searchrad=self.searchrad,
-                separation=self.separation,
-                use2dhist=self.use2dhist,
-                tolerance=self.tolerance,
-                xoffset=0,
-                yoffset=0,
-            )
+                    except AttributeError as e:
+                        self.log.error(f"Failed to retrieve tweakreg_catalog: {e}")
+                        images.shelve(image_model, i, modify=False)
```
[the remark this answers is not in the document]
mairanteodoro: If an `AttributeError` exception is raised that means it was not possible to find the catalog in the meta at all. In that case, there's nothing to be modified.

romancal/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
braingram: I opened a PR with a test that fails with this PR. I don't know if this is a representative test (it has 2 WFI_IMAGE and 1 WFI_GRISM members) but I hope it illustrates the problem. The test fails on the call to `del image_model.meta.source_detection["tweakreg_catalog"]` because it attempts to delete `tweakreg_catalog` from the grism observation (and it didn't add `tweakreg_catalog` earlier in the step).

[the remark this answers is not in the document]
schlafly: Can you remind me, Mairan, if we need to support passing spectroscopic data to tweakreg at present?  e.g., because the regtests send it, or because of the issue you identified in the ELP where it's passing more than it needs to.  I agree with Brett that morally we should probably just error out immediately if we ever see spectroscopic data passed to tweakreg, but we don't need that in this PR necessarily.

romancal/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
schlafly: Brett, if we leave modify=False here, but do modify the file, what happens?  In this case where we've failed to get the catalog we probably don't want to modify the file.

Mairan, it looks to me like that tweakreg_catalog entry is using an old format we don't want to support (guessing, based on the Table(asarray(...)) wrapping!) so an option would be to just delete this bit if that doesn't have too many other knock-on consequences.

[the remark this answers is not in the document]
braingram: > Brett, if we leave modify=False here, but do modify the file, what happens? In this case where we've failed to get the catalog we probably don't want to modify the file.

It depends on if the `ModelLibrary` is `on_disk=True` or `on_disk=False`. For `on_disk=True` passing `modify=False` will prevent writing a temporary file (so if the model was modified a later borrow won't have the modifications). For `on_disk=False` passing `modify=False` does nothing (since the model is already in memory). The general advice is to only use `modify=False` if you're 100% sure there was no modification otherwise it introduces a change in behavior depending on the state of `on_disk`. I agree with @mairanteodoro that the modification doesn't look reachable so `modify=False` is ok here.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
-                        # Large corrections are typically a result of source
-                        # mis-matching or poorly-conditioned fit. Skip such models.
-                        self.log.warning(
-                            "WCS has been tweaked by more than"
-                            f" {10 * self.tolerance} arcsec"
+                    if self.save_abs_catalog:
+                        output_name = os.path.join(
+                            self.catalog_path, f"fit_{self.abs_refcat.lower()}_ref.ecsv"
+                        )
+                        catalog.write(
```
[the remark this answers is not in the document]
mairanteodoro: Yes, it does.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
-                searchrad=self.abs_searchrad,
-                separation=self.abs_separation,
-                use2dhist=self.abs_use2dhist,
-                tolerance=self.abs_tolerance,
-                xoffset=self.abs_xoffset,
-                yoffset=self.abs_yoffset,
-            )
+        # run alignment only if it was possible to build image catalogs
+        if len(imcats):
+            if getattr(images, "group_indices", None) and len(images.group_indices) > 1:
```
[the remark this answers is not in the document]
mairanteodoro: Now that we're using ModelLibrary, I don't think so; `group_indices` is a property of the class.

romancal/tweakreg/tweakreg_step.py
mairanteodoro: I don't know, but I'll ask around to see if it still is.

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Yes, since we moved it to `stcal`, there's no need to keep it around anymore.
The current `stcal` test for that is here: https://github.com/spacetelescope/stcal/blob/c3e133228528f369eb85d34c81e2ceb4c76eaaff/tests/test_tweakreg.py#L136-L148

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Correct.

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Not needed anymore since the step will respect `output_dir`.

romancal/tweakreg/tests/test_tweakreg.py
mairanteodoro: Since we moved the validation of the column names to a helper, a corresponding unit test has been implemented to replace the old unit test.

romancal/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
mairanteodoro: > Can you remind me, Mairan, if we need to support passing spectroscopic data to tweakreg at present?

Not to my knowledge. For example, the ELP pipeline behavior is to check for the exposure type, and if it is not a `"WFI_IMAGE"` then it sets `meta.cal_step.tweakreg` to `"SKIPPED"` and then added to a list that will be passed to `TweakRegStep` at the end. There's a [note in the ELP code](https://github.com/spacetelescope/romancal/blob/96af8bbc6a6dafd48d1beb251b5209d9da875ca1/romancal/pipeline/exposure_pipeline.py#L183-L186) that reads:
```
# Now that all the exposures are collated, run tweakreg
# Note: this does not cover the case where the asn mixes imaging and spectral
```

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +141,331 @@ def process(self, input):
...
-                        "fitmask",
-                    ]:
-                        del image_model.meta["wcs_fit_results"][k]
-
-                    image_model.meta.wcs = imcat.wcs
-                images.shelve(image_model, i)
+            # finalize step
+            with images:
+                for i, imcat in enumerate(imcats):
+                    image_model = images.borrow(i)
```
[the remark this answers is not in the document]
mairanteodoro: For now, I just used a `dict` to map the model index to the corresponding catalog.

romancal/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
mairanteodoro: For now, `TweakRegStep` will always return all the datamodels passed on to it (even mixed exposure types). However, if the exposure is of type "WFI_IMAGE", the step will process it normally. Otherwise, it will set `meta.cal_step="SKIPPED"` and nothing else. But all the updated input files will still be returned to the caller.

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1185,3 +991,80 @@ def test_parse_catfile_raises_error_on_invalid_content(tmp_path, catfile_line_co
...
+
+    img3 = base_image(shift_1=1000, shift_2=1000)
+    img3.meta.exposure.type = "WFI_GRISM"
+
+    res = trs.TweakRegStep.call([img1, img2, img3])
+
+    assert len(res) == 3
+    assert img1.meta.cal_step.tweakreg == "COMPLETE"
+    assert img2.meta.cal_step.tweakreg == "COMPLETE"
+    assert img3.meta.cal_step.tweakreg == "SKIPPED"
```
braingram, line 1070:
```suggestion
    assert img1.meta.cal_step.tweakreg == "COMPLETE"
    assert img2.meta.cal_step.tweakreg == "SKIPPED"
    assert img3.meta.cal_step.tweakreg == "COMPLETE"
```

romancal/tweakreg/tests/test_tweakreg.py
```diff
@@ -1185,3 +991,80 @@ def test_parse_catfile_raises_error_on_invalid_content(tmp_path, catfile_line_co
...
+    img1 = base_image(shift_1=1000, shift_2=1000)
+    add_tweakreg_catalog_attribute(tmp_path, img1, catalog_filename="img1")
+    img1.meta.exposure.type = "WFI_IMAGE"
+
+    img2 = base_image(shift_1=1000, shift_2=1000)
+    add_tweakreg_catalog_attribute(tmp_path, img2, catalog_filename="img2")
+    img2.meta.exposure.type = "WFI_IMAGE"
+
+    img3 = base_image(shift_1=1000, shift_2=1000)
+    img3.meta.exposure.type = "WFI_GRISM"
```
braingram, line 1063:
```suggestion
    img2 = base_image(shift_1=1000, shift_2=1000)
    img2.meta.exposure.type = "WFI_GRISM"

    img3 = base_image(shift_1=1000, shift_2=1000)
    add_tweakreg_catalog_attribute(tmp_path, img3, catalog_filename="img3")
    img3.meta.exposure.type = "WFI_IMAGE"
```

Thanks! I think the previous bug might not have been caught by this test if the `WFI_GRISM` exposure was last. I added 2 suggestions here to move the `WFI_GRISM` exposure to second-to-last. The test (should) pass both as it is now and with the suggestion.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -6,35 +6,16 @@
...
 
 import numpy as np
-from astropy import units as u
-from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from roman_datamodels import datamodels as rdm
-from tweakwcs.correctors import JWSTWCSCorrector
-from tweakwcs.imalign import align_wcs
-from tweakwcs.matchutils import XYXYMatch
+from stcal.tweakreg import tweakreg
```
braingram, line 11:
```suggestion
from stcal.tweakreg import tweakreg
from stcal.tweakreg.tweakreg import _SINGLE_GROUP_REFCAT_STR, SINGLE_GROUP_REFCAT
```

romancal/tweakreg/tweakreg_step.py
```diff
@@ -6,35 +6,16 @@
...
-    elif nelem == 2:
-        return f"{str_list[0]} or {str_list[1]}"
-    else:
-        return ", ".join(map(repr, str_list[:-1])) + ", or " + repr(str_list[-1])
-
-
-SINGLE_GROUP_REFCAT = ["GAIADR3", "GAIADR2", "GAIADR1"]
-_SINGLE_GROUP_REFCAT_STR = _oxford_or_str_join(SINGLE_GROUP_REFCAT)
+SINGLE_GROUP_REFCAT = tweakreg.SINGLE_GROUP_REFCAT
+_SINGLE_GROUP_REFCAT_STR = tweakreg._SINGLE_GROUP_REFCAT_STR
```
braingram, line 18:
```suggestion
```
I added a suggestion above to import these directly from stcal instead of re-assigning them to variables here.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +128,325 @@ def process(self, input):
...
+                        {
+                            "model_index": i,
+                            "imcat": tweakreg.construct_wcs_corrector(
+                                wcs=image_model.meta.wcs,
+                                refang=image_model.meta.wcsinfo,
+                                catalog=catalog_table,
+                                group_id=image_model.meta.group_id,
+                            ),
+                        }
+                    )
```
braingram, line 217:
```suggestion
                    imcat = tweakreg.construct_wcs_corrector(
                                wcs=image_model.meta.wcs,
                                refang=image_model.meta.wcsinfo,
                                catalog=catalog_table,
                                group_id=image_model.meta.group_id,
                            )
                    imcat.meta["model_index"] = i
                    imcats.append(imcat)
```
The correctors have a `meta` attribute (sort of like datamodels) that can be used here to store the `model_index`.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +128,325 @@ def process(self, input):
...
-                separation=self.abs_separation,
-                use2dhist=self.abs_use2dhist,
-                tolerance=self.abs_tolerance,
-                xoffset=self.abs_xoffset,
-                yoffset=self.abs_yoffset,
-            )
+        # run alignment only if it was possible to build image catalogs
+        if len(imcats):
+            # extract WCS correctors to use for image alignment
+            correctors = [x["imcat"] for x in imcats]
```
braingram, line 223:
```suggestion
```

romancal/tweakreg/tweakreg_step.py
```diff
@@ -134,460 +128,325 @@ def process(self, input):
...
+            if len(images.group_indices) > 1:
+                self.do_relative_alignment(correctors)
```
braingram, line 225:
```suggestion
                self.do_relative_alignment(imcats)
```

braingram (commented):
Thanks for addressing my comments. I added a few more (final) ones:
- a minor change to one of the tests
- a minor change to the imports from stcal
- using `imcat.meta["model_index"]` instead of a dict containing an imcat and the model index

braingram (approved):
As the remaining comments are more style than substance I'll mark this as approved and feel free to make the changes if you agree or forge ahead otherwise.

Thanks!

schlafly (approved):
Thanks Mairan, looks good to me!

Issue #1325, opened 2024-07-22
`test_level2_image_processing_pipeline` and `test_level2_grism_processing_pipeline` lead to unexpected "okify" results


braingram:
The [test_level2_image_processing_pipeline](https://github.com/spacetelescope/romancal/blob/a8d1007decf805d3e1b7ce293f37c0700f856fb4/romancal/regtest/test_wfi_pipeline.py#L30) and [test_level2_grism_processing_pipeline](https://github.com/spacetelescope/romancal/blob/a8d1007decf805d3e1b7ce293f37c0700f856fb4/romancal/regtest/test_wfi_pipeline.py#L277) tests are very long, containing many asserts and more than 1 call to `compare_asdf`.

This is an issue for "okifying" results. For example, if `test_level2_image_processing_pipeline` fails on the first `compare_asdf`:
https://github.com/spacetelescope/romancal/blob/a8d1007decf805d3e1b7ce293f37c0700f856fb4/romancal/regtest/test_wfi_pipeline.py#L46
and the result is "oked". The next regression tests will likely continue to fail because the second `compare_asdf`:
https://github.com/spacetelescope/romancal/blob/a8d1007decf805d3e1b7ce293f37c0700f856fb4/romancal/regtest/test_wfi_pipeline.py#L258
is never reached and the second file is never produced for "okifying".

It would be beneficial to break up these long tests to avoid needing to okify several regression test runs (and to make the test output more informative).

PR #1426, opened 2024-09-30
Split long regtests into reusable fixtures and smaller named tests

braingram:
Closes https://github.com/spacetelescope/romancal/issues/1325

This PR refactors some large regtests:
- moving the pipeline run to a module scoped fixture (the result will be reused for any tests that use the fixture in the module)
- splitting asserts into separate named tests that use the new fixture
- removing the multiple `passfail` functions and their use (the test results provide information about if the test passed or failed)

There are a few items to resolve for this PR:
- [x] does the new `dms_requirement_tests.json` need parametrization values (and are they supported) for parameterized tests that correspond to dms requirements? (for example `romancal.regtest.test_mos_pipeline.test_steps_ran` now tests DMS400 but technically only using the `skymatch` parametrization) EDIT: @zacharyburnett confirmed the non-parametrized one is the one to use.
- [ ] many log messages contained mentions of dms requirements that are not in `dms_requirement_tests.json` (and did not previously have a decorator). These log messages aren't in a format that the metrics collector understands. If these were added to verify a requirement those requirements are not being verified. I suspect we want to convert the log messages into entries into `dms_requirement_tests.json` and will add comments for those dms requirements.

Regression tests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/11071508240

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/regtest/test_wfi_grism_16resultants.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.fixture(scope="module")
+def input_model(input_filename):
+    with rdm.open(input_filename) as model:
+        yield model
+
+
+def test_output_is_image_model(output_model):
+    # DMS280  FIXME comment says 280
+    # DMS414  FIXME log message said 414
+    # FIXME was not an assert
```
braingram, line 59: The current code:
https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/regtest/test_wfi_16resultants.py#L98-L102
Mentions DMS280 in the comment but DMS414 in the log message and contains no assert. What is the expected DMS requirement here?

romancal/regtest/test_wfi_grism_pipeline.py
```diff
@@ -0,0 +1,156 @@
...
+    assert hasattr(output_model, arr_name)
+
+
+def test_has_exposure_time(output_model):
+    # DMS88 total exposure time exists
+    assert "exposure_time" in output_model.meta.exposure
+
+
+def test_wcs_has_bounding_box(output_model):
+    # DMS93 WCS tests  TODO isn't listed
```
braingram, line 132: DMS93 isn't listed in the requirements. Is it satisfied and should it be?

romancal/regtest/test_wfi_grism_pipeline.py
```diff
@@ -0,0 +1,156 @@
...
+    # DMS90
+    repointed_filename, _ = repointed_filename_and_delta
+
+    rtdata.get_truth(f"truth/WFI/grism/{Path(repointed_filename).name}")
+    diff = compare_asdf(repointed_filename, rtdata.truth, **ignore_asdf_paths)
+    assert diff.identical, diff.report()
+
+
+def test_repointed_wcs_differs(repointed_filename_and_delta, output_model):
+    # DMS93  TODO isn't listed
```
braingram, line 148: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781144017

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+@pytest.mark.soctests
+def test_jump_in_uneven_ramp(output_model):
+    # DMS361 jump detection detected jumps in uneven ramp
+    uneven = len({len(x) for x in output_model.meta.exposure.read_pattern}) > 1
+    assert uneven & np.any(output_model.dq & pixel.JUMP_DET)
+
+
+@pytest.mark.soctests
+def test_has_a_wcs(output_model):
+    # DMS129  TODO not listed
```
braingram, line 127: DMS129 isn't listed in the requirements. Is it satisfied?

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+        "dq_init",
+        "jump",
+        "linearity",
+        "ramp_fit",
+        "saturation",
+    ),
+)
+def test_steps_ran(output_model, step_name):
+    # DMS86
+    # also DMS129 for assign_wcs
```
braingram, line 114: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781145752

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+
+@pytest.mark.soctests
+def test_has_a_wcs(output_model):
+    # DMS129  TODO not listed
+    assert output_model.meta.wcs is not None
+
+
+@pytest.mark.soctests
+def test_wcs_has_distortion_information(output_model):
+    # DMS129  TODO not listed
```
braingram, line 133: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781145752

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+    assert "v2v3" in output_model.meta.wcs.available_frames
+
+
+@pytest.mark.soctests
+def test_wcs_applies_distortion_correction(output_model):
+    # DMS129  TODO not listed
```
braingram, line 139: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781145752

romancal/regtest/test_wfi_grism_pipeline.py
```diff
@@ -0,0 +1,156 @@
...
+        ("flat_field", "SKIPPED"),
+        ("photom", "SKIPPED"),
+        ("source_detection", "SKIPPED"),
+        ("tweakreg", "SKIPPED"),
+    ),
+)
+def test_step_status(output_model, step_name, status):
+    # DMS90
+    # DMS278
+    # also DMS129 for assign_wcs
```
braingram, line 110: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781145752

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+@pytest.mark.soctests
+def test_has_exposure_time(output_model):
+    # DMS88 total exposure time exists
+    assert "exposure_time" in output_model.meta.exposure
+
+
+@pytest.mark.soctests
+@pytest.mark.parametrize("meta_attribute", ("detector", "optical_element"))
+def test_instrument_meta(output_model, meta_attribute):
+    # DMS-136 PSF tests  TODO not listed
```
braingram, line 171: DMS136 isn't listed in the requirements. Is it satisfied?

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+    assert meta_attribute in output_model.meta.instrument
+
+
+@pytest.mark.soctests
+def test_wcs_has_bounding_box(output_model):
+    # DMS89 WCS tests  TODO not listed
```
braingram, line 177: DMS89 is not listed in the requirements. Is it satisfied?

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+    assert len(output_model.meta.wcs.bounding_box) == 2
+
+
+@pytest.mark.soctests
+def test_repointed_matches_truth(
+    repointed_filename_and_delta, rtdata, ignore_asdf_paths
+):
+    # DMS89  TODO not listed
```
braingram, line 185: Same as https://github.com/spacetelescope/romancal/pull/1426/files#r1781149604

romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+    repointed_filename, _ = repointed_filename_and_delta
+
+    rtdata.get_truth(f"truth/WFI/image/{Path(repointed_filename).name}")
+    diff = compare_asdf(repointed_filename, rtdata.truth, **ignore_asdf_paths)
+    assert diff.identical, diff.report()
+
+
+@pytest.mark.soctests
+def test_repointed_wcs_differs(repointed_filename_and_delta, output_model):
+    # DMS89  TODO not listed
```
braingram, line 195: https://github.com/spacetelescope/romancal/pull/1426/files#r1781149604

romancal/regtest/test_mos_pipeline.py
```diff
@@ -32,123 +27,93 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
-    model = rdm.open(rtdata.output, lazy_load=False)
 
-    pipeline.log.info(
-        "DMS373 MSG: Testing the creation of a Level 3 mosaic image resampled to a skycell"
-        + passfail(model.meta.cal_step.resample == "COMPLETE")
-    )
+def test_thumbnail_exists(thumbnail_filename):
+    # DMS356
+    # FIXME was not an assert before
+    assert os.path.isfile(thumbnail_filename)
```
braingram, line 74: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_mos_pipeline.py
```diff
@@ -32,123 +27,93 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+
+
+def test_preview_exists(preview_filename):
+    # DMS356
+    # FIXME was not an assert before
+    assert os.path.isfile(preview_filename)
```
braingram, line 80: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_mos_pipeline.py
```diff
@@ -32,123 +27,93 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+
+
+@pytest.mark.parametrize("suffix", ("cat", "segm"))
+def test_file_exists(output_filename, suffix):
+    # DMS374 for catalog and segm
+    # FIXME was not an assert before
+    expected_filename = output_filename.rsplit("_", 1)[0] + f"_{suffix}.asdf"
+    assert os.path.isfile(expected_filename)
```
braingram, line 88: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_mos_pipeline.py
```diff
@@ -32,123 +27,93 @@ def test_level3_mos_pipeline(rtdata, ignore_asdf_paths):
...
+
+
+def test_output_is_mosaic(output_model):
+    # DMS356
+    # FIXME comment says result is an ImageModel... it's not
+    assert isinstance(output_model, rdm.datamodels.MosaicModel)
```
braingram, line 94: The previous code:
https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/regtest/test_mos_pipeline.py#L60-L64
Says that DMS356 is "result is an ImageModel" that's not the case here. Does this fail the requirement?

romancal/regtest/test_mos_skycell_pipeline.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_output_matches_truth(output_filename, truth_filename, ignore_asdf_paths):
+    diff = compare_asdf(output_filename, truth_filename, **ignore_asdf_paths)
+    assert diff.identical, diff.report()
+
+
+def test_resample_ran(output_model):
+    # DMS373 test output is resampled to a skycell
+    # FIXME this doesn't test if the output is a skyceell
+    assert output_model.meta.cal_step.resample == "COMPLETE"
```
braingram, line 53: The previous code:
https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/regtest/test_mos_pipeline.py#L151-L154
mentions that the requirement is that the output is resampled to a skycell. The test doesn't appear to test that (and was not previously an assert).

romancal/regtest/test_wfi_grism_16resultants.py
```diff
@@ -0,0 +1,91 @@
...
+    assert isinstance(output_model, rdm.datamodels.ImageModel)
+
+
+def test_input_has_16_resultants(input_model):
+    # DMS414
+    # FIXME was not an assert
+    assert len(input_model.meta.exposure.read_pattern) == 16
```
braingram, line 66: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_wfi_grism_16resultants.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_output_has_16_resultants(output_model):
+    # DMS414
+    # FIXME was not an assert
+    assert len(output_model.meta.exposure.read_pattern) == 16
```
braingram, line 72: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_wfi_image_16resultants.py
```diff
@@ -0,0 +1,90 @@
...
+    # DMS280  FIXME comment says 280
+    # DMS413  FIXME log message said 413
+    # FIXME was not an assert
+    assert isinstance(output_model, rdm.datamodels.ImageModel)
+
+
+def test_input_has_16_resultants(input_model):
+    # DMS413
+    # FIXME was not an assert
+    assert len(input_model.meta.exposure.read_pattern) == 16
```
braingram, line 66: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_wfi_image_16resultants.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_output_has_16_resultants(output_model):
+    # DMS413
+    # FIXME was not an assert
+    assert len(output_model.meta.exposure.read_pattern) == 16
```
braingram, line 72: This was not previously as assert. It passes so I'm assuming it's ok to make it one.

romancal/regtest/test_wfi_image_16resultants.py
```diff
@@ -0,0 +1,90 @@
...
+def input_model(input_filename):
+    with rdm.open(input_filename) as model:
+        yield model
+
+
+def test_output_is_image_model(output_model):
+    # DMS280  FIXME comment says 280
+    # DMS413  FIXME log message said 413
+    # FIXME was not an assert
+    assert isinstance(output_model, rdm.datamodels.ImageModel)
```
braingram, line 60: The previous code:
https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/regtest/test_wfi_16resultants.py#L43-L47
Mentions DMS280 in the comment but DMS413 in the log message and was not an assert. Which requirement does this satisfy?

braingram:
I'm going to open this for review to try to get some help on the questions about the DMS requirements. There are a number of inconsistencies and missing requirements (ones mentioned in the code but not in the requirement file, for the ones I investigated they also did not previously have a decorator).

braingram:
I also marked this as `no-changelog-needed` as it's all regtest changes. I somewhat think this deserves a changelog fragment (perhaps under `general`) given the size of the changes so please let me know if one makes sense.

braingram:
Thanks for taking a look.

> The tests are designed to not keep reporting the same errors, this is confusing, thus they often stop at the error.

There shouldn't be any duplicate checks (that weren't already in the code). There also should not be any errors (given the regtests are passing). What errors are you seeing and what duplicates?

> The tests are not just to test the formal DMS requirements.

Understood. The tests were reorganized to fix https://github.com/spacetelescope/romancal/issues/1325 (which is a direct result of having long regtests that generate multiple files with multiple checks). It's my understanding that the regtests are expected to demonstrate that the code satsifies the listed DMS requirements (hence the need for the requirements file and associated infrastructure).

zacharyburnett:
> I also marked this as `no-changelog-needed` as it's all regtest changes. I somewhat think this deserves a changelog fragment (perhaps under `general`) given the size of the changes so please let me know if one makes sense.

Since this is a pretty big change to regtests and affects file generation and DMS reqs outside, perhaps there should be a changelog entry so that future people coming here can look in `CHANGES.rst` to see when the change happened

romancal/regtest/test_wfi_grism_16resultants.py
schlafly: 414 is right.  413 in the ~docstring and 280 are wrong.  This is also correct (414) in https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/tests/dms_requirement_tests.json .

romancal/regtest/test_wfi_grism_pipeline.py
schlafly: I think the issue here is that DMS9 is spuriously listed as being tested in this step, and we should have done DMS93.

romancal/regtest/test_wfi_grism_pipeline.py
schlafly: I think this is the DMS9 <-> DMS93 issue.

romancal/regtest/test_wfi_image_pipeline.py
schlafly: Yes, we should mark this as done.  I can't figure out why it's not picked up by John's page, though.

romancal/regtest/test_wfi_image_pipeline.py
schlafly: Yes, we should go ahead and mark it.  I don't understand why it's not tracked on John's "missing requirements" page.

romancal/regtest/test_wfi_image_pipeline.py
schlafly: This is likewise a requirement that should be added to the requirements.json list; we've delivered this requirement and this was the test that did that.

romancal/regtest/test_mos_pipeline.py
schlafly: That seems reasonable to me.

romancal/regtest/test_mos_pipeline.py
schlafly: That seems reasonable to me.

romancal/regtest/test_mos_pipeline.py
schlafly: This is a copy/paste error, we want the mosaic pipeline to make mosaic models.  The comment should be updated.

romancal/regtest/test_mos_skycell_pipeline.py
schlafly: I agree that the comment is misleading.  Let's leave this for now.  Perhaps the test should be updated to look for https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/mosaic_basic-1.0.0.yaml#L133 , but I'm not sure that we actually populate that right now.  I'll spin that into a separate issue.

romancal/regtest/test_wfi_image_16resultants.py
schlafly: 413 is correct here.

schlafly:
Thanks Brett.  I agree that it's very helpful to have split the two compare_asdfs apart, so that one doesn't need to okify twice to fix the relevant regtests.

I think Dave is concerned that when one slices and dices the tests very finely one often sees that many tests fail when there is "really" only one problem, and that can be confusing.  But it's also beneficial in that you get to see all the problems at once.  I tend to be ambivalent there---at the end of the day I usually end up fixing things one at a time anyway, and then rerun tests---but your reorganization is clean and looks good to me.

braingram:
> I think Dave is concerned that when one slices and dices the tests very finely one often sees that many tests fail when there is "really" only one problem, and that can be confusing. But it's also beneficial in that you get to see all the problems at once. I tend to be ambivalent there---at the end of the day I usually end up fixing things one at a time anyway, and then rerun tests---but your reorganization is clean and looks good to me.

Thanks for the description. That's helpful to understand the goal.

To work through one possible example for the current `test_level2_image_processing_pipeline` the first assert is the `compare_asdf` against truth. I don't think this alone can be used to say the code satisfies most (if not all) DMS requirements (since the truth file might not satisfy the requirement).

So let's say we need additional asserts for DMS requirements. If we put them in the same test as `compare_asdf` anything that causes this test to fail will show as failed DMS requirements. I understand that the goal of the regtests is not to test the DMS requirements but I think at the moment that is what we have to work with.

Ignoring all details related to the DMS requirements, I think the structure of having a fixture that runs a pipeline/step then separate tests that check the outputs is helpful. The mosaic pipeline produces 3 files and currently they are all checked in the same test `test_level3_mos_pipeline`. This means that a change to the pipeline might cause the first assert to fail, the developer fixes the issue, then re-runs the regtests, finds that a later assert fails, fixes the issue, etc... This PR puts those checks in separate tests so the developer knows more from the first run.

romancal/regtest/test_wfi_grism_pipeline.py
braingram: Thanks! Clear as mud :)

I removed DMS9 and added DMS93 in 03ed9b4

romancal/regtest/test_wfi_grism_pipeline.py
braingram: Thanks, did 03ed9b4 fix this?

romancal/regtest/test_wfi_image_pipeline.py
braingram: Added DMS129 in 4c0ad1d

romancal/regtest/test_wfi_image_pipeline.py
braingram: Added DMS129 in 4c0ad1d

romancal/regtest/test_wfi_image_pipeline.py
braingram: Thanks! Added DMS136 in 229b164

romancal/regtest/test_wfi_image_pipeline.py
braingram: Added DMS89 in 03ce0ad

romancal/tests/test_dms_requirements.py
```diff
@@ -20,7 +20,7 @@ def test_requirements():
     }
 
     existing_tests = set()
-    test_regex = re.compile(r"def (test_[^\(]+)\(.*\):")
+    test_regex = re.compile(r"def (test_[^\(]+)\(")
```
braingram, line 23: @zacharyburnett I modified the regex here to catch multi-line function signatures which won't end in `):`. Does this look good to you?

romancal/regtest/test_wfi_grism_pipeline.py
schlafly: Yes, thanks!

schlafly:
Thanks Brett.  I think we do need to include the additional asserts, and I agree it's important to only need to okify once.  Following the discussion on the telecon yesterday, I'm going ahead and approving this---thanks!

schlafly (approved):

romancal/tests/test_dms_requirements.py
zacharyburnett: Good catch! Yeah that looks good, it only needs the function name

PR #1437, opened 2024-10-03
rcal-920 script to create associations based on skycells

### LINKED ISSUE #1421: Utility to create asn files based on skycells
_Issue [RCAL-920](https://jira.stsci.edu/browse/RCAL-920) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

ddavis-stsci:
Resolves [RCAL-920](https://jira.stsci.edu/browse/RCAL-920)

Closes #1421 

This PR adds code to take an input list of calibrated WFI exposures and creates associations based on the 
skycells that they overlap. 

## Tasks

Regression test results are at 
https://github.com/spacetelescope/RegressionTests/actions/runs/11168259268

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/regtest/test_skycell_generation.py
```diff
@@ -0,0 +1,47 @@
...
+from romancal.associations import skycell_asn
+
+
+def passfail(bool_expr):
+    """set pass fail"""
+    if bool_expr:
+        return "Pass"
+    return "Fail"
+
+
```
braingram, line 16:
```suggestion
```

romancal/regtest/test_skycell_generation.py
```diff
@@ -0,0 +1,47 @@
...
+        "r512_r274dp63x32y80_visit_F158_prompt_i2d_asn.json",
+        "r512_r274dp63x32y81_visit_F158_prompt_i2d_asn.json",
+        "r512_r274dp63x33y80_visit_F158_prompt_i2d_asn.json",
+        "r512_r274dp63x33y81_visit_F158_prompt_i2d_asn.json",
+    ]
+    # Test that the json files exist
+    for file in output_files:
+        skycell_asn.logger.info(
+            "Check that the json file exists   " + passfail(os.path.isfile(file))
+        )
```
braingram, line 47:
```suggestion
    missing_files = [fn for fn in output_files if not os.path.isfile(fn)]
    assert not missing_files, f"Missing files: {missing_files}"
```
How does this look for an `assert` to add to this test? Without an assert that checks there are no missing files the test can silently pass if one or all of the files are missing.

romancal/associations/skycell_asn.py
```diff
@@ -0,0 +1,172 @@
...
+            + suffix
+        )
+        with open(asn_file_name + "_asn.json", "w") as outfile:
+            prompt_product_asn = asn_from_list.asn_from_list(
+                member_list, product_name=asn_file_name
+            )
+            prompt_product_asn["asn_type"] = "image"
+            prompt_product_asn["program"] = program_id
+            prompt_product_asn["target"] = patch_name
+            prompt_product_asn["skycell_wcs_info"] = json.dumps(projcell_info)
```
schlafly, line 80: I think I would have imagined leaving this as a dictionary, rather than converting it to a string via dumps?  What's the motivation for stringifying this?

schlafly (approved):
Thanks Dave.  That looks good.  I left one inline comment re whether or not to convert projcell_info to a string.

As follow-up work, could we add the following features?
- Let's extend the projcell info to include the orientat_cent from orientat_projection_center.  This is presently always zero, and for now we can just throw an error if it's not equal to that, but it's the one bit of information we're not tracking.
- Let's write the code to actually use this wcs info in resample.
- Presently this code generates the "visit" products, though it kind of just assumes that you pass it a visit.  It would be good if you just passed it a list of files and it figured out how to make the various pass / visit / full combinations.  I think you could generate those with an additional loop over these, something like the following pseudocode:
```
obsid_list = []
for file_name in filelist:
    obsid_list.append(cal_file.meta.observation.obs_id)
visitid = [o[:19] for o in obsid_list]
passid = [o[:10] for o in obsid_list]
fullid = [o[:5] for o in obsid_list]
# generate visit products
for visit in unique(visitid):
    generate_products([f for f, v in zip(filelist, visitid) if v == visit], type='visit')
# generate pass products
for passno in unique(passid):
    generate_products([f for f, p in zip(filelist, passid) if p == passno], type='pass')
for progno in np.unique(fullid):
    generate_products([f for f, p in zip(filelist, fullid) if p == progno), type='full')
```
The other thing that we would need to loop over would be the filter names, so that generate_products only sees one filter at a time.

romancal/associations/skycell_asn.py
ddavis-stsci: Keeping this a string  allows the other unit tests to pass since we're using one json schema file at the moment. Since a dictionary is more convenient we should add schema for the skycell asn - at a later date.

PR #1438, opened 2024-10-04
Make var_flat optional.

schlafly:
This PR makes the ELP not compute var_flat by default for individual exposures.  It adds a new keyword "include_var_flat" to the flat field step spec to enable including the flat field variances, with a default value of False---we will not by default compute var_flat.  It adds tests to make sure that this keyword is being followed.

Together with https://github.com/spacetelescope/rad/pull/462, it makes the mosaic pipeline either create a var_flat extension or not, depending on whether the input exposures contain var_flat.  If not all of the input exposures include var_flat, it will not compute var_flat in the output.

This will lead to changes in the regression tests, since we will not be computing var_flat by default.  The derived errors will also be different.  If we update the exposures used as input for the mosaic pipeline to not include flat variances, the derived mosaics will also be different.

"err" will no longer include a contribution from var_flat by default, in either the exposures or eventually the mosaics.  In principle we could include var_flat in "err" but still delete var_flat by default.  This seems a little perverse to me, but I mention it as a possibility.

It makes a handful of forward-looking changes to the regression tests, essentially not requiring that var_flat exist.

@tddesjardins , @hcferguson, you've expressed opinions about removing this in the past; care to weigh in?

## Tasks

### DIFF
--- a/docs/roman/flatfield/main.rst
+++ b/docs/roman/flatfield/main.rst
@@ -59,3 +59,10 @@ and finally the error that is associated with the science data is given by,
 
 The total ERR array in the science exposure is updated as the square root
 of the quadratic sum of VAR_POISSON, VAR_RNOISE, and VAR_FLAT.
+
+Note that by default we do not compute VAR_FLAT nor include its
+contribution to ERR, unless the "include_var_flat" is specified.  This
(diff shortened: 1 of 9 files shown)


ddavis-stsci (approved):
LGTM

PR #1457, opened 2024-10-15
Copy input datamodel in source_catalog step

larrybradley:
With this PR the input datamodel to the `source_catalog` step is copied before being operated on so that it is left completely unaltered.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/source_catalog/source_catalog_step.py
+++ b/romancal/source_catalog/source_catalog_step.py
@@ -6,6 +6,7 @@ import numpy as np
 from astropy.table import Table
 from roman_datamodels import datamodels, maker_utils
 from roman_datamodels.datamodels import ImageModel, MosaicModel
+from roman_datamodels.maker_utils import mk_datamodel
 
 from romancal.source_catalog.background import RomanBackground
 from romancal.source_catalog.detection import convolve_data, make_segmentation_image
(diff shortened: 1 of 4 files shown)


larrybradley:
The model.err array is also modified in-place in `RomanSourceCatalog` to convert from MJy/sr to uJy.  But I can can update this to only copy those 2 arrays.

schlafly:
If the source_catalog code is responsible for all of the modifications and so we can safely save a couple of copies in it and restore them at the end, let's do that.  If photutils is "allowed" to modify the arrays we're passing it in the model then let's go ahead and do the full model copy.

Maybe let's also document in the docstring to some function that source_catalog is allowed to change fields data & err by a scale factor (numerical precision differences only).  Thanks!

larrybradley:
@schlafly I've updated this PR to copy *only* the `data` and `err` arrays (that are modified in place in the step).

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -324,8 +324,8 @@ def test_l3_input_model_unchanged(mosaic_model, tmp_path):
         save_results=False,
     )
 
-    assert_allclose(original_data, mosaic_model.data, atol=5.0e-5)
-    assert_allclose(original_err, mosaic_model.err, atol=5.0e-5)
+    assert_equal(original_data, mosaic_model.data)
+    assert_equal(original_err, mosaic_model.err)
```
schlafly, line 328: Thanks for this test!

romancal/source_catalog/source_catalog_step.py
```diff
@@ -111,40 +136,34 @@ def process(self, input_model):
...
+            # set the suffix to something else to prevent the step from
+            # overwriting the source catalog file with a datamodel
             self.suffix = "sourcecatalog"
-            # return DataModel
-            result = output_model
+
+            if isinstance(input_model, ImageModel):
+                update_metadata(input_model, output_catalog_name)
+
+            result = input_model
```
schlafly, line 159: I think this looks right.  We'll see if the regtests for the pipeline complain.

schlafly (approved):
This looks good to me, pending successful completion of the regression tests.

romancal/source_catalog/source_catalog_step.py
larrybradley: The only reason this is necessary is because for L2 we modify the input model metadata with the catalog filename.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -300,8 +300,8 @@ def test_l2_input_model_unchanged(image_model, tmp_path):
         save_results=False,
     )
 
-    assert_allclose(original_data, image_model.data, atol=5.0e-5)
-    assert_allclose(original_err, image_model.err, atol=5.0e-5)
+    assert_equal(original_data, image_model.data)
+    assert_equal(original_err, image_model.err)
```
mairanteodoro, line 304: I think it would be safer to use `assert_allclose()` because both arrays are of type `float32`.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -324,8 +324,8 @@ def test_l3_input_model_unchanged(mosaic_model, tmp_path):
         save_results=False,
     )
 
-    assert_allclose(original_data, mosaic_model.data, atol=5.0e-5)
-    assert_allclose(original_err, mosaic_model.err, atol=5.0e-5)
+    assert_equal(original_data, mosaic_model.data)
```
mairanteodoro, line 327: I think it would be safer to use `assert_allclose()` because both arrays are of type `float32`.

mairanteodoro (commented):
Looks good to me.
I only have one concern about using `assert_equal()` to compare `float32` arrays.

schlafly:
This "fails" regtests but I think that's just saying that this is working and it's a numerical noise thing that is now fixed that we should okify.  The only failures are pipeline tests that run source cataloging with small numerical differences.
https://github.com/spacetelescope/RegressionTests/actions/runs/11839507319/job/32991232674

larrybradley:
This was out-of-sync with main, so I just rebased.  I'll merge when CI passes again.

romancal/source_catalog/tests/test_source_catalog.py
larrybradley: With this PR the arrays really are identical, so `assert_equal` should always work.

romancal/source_catalog/tests/test_source_catalog.py
larrybradley: same as above

PR #1471, opened 2024-10-23
rcal-936 Update skycell_asn docs and add skycell_asn as a script at install time

### LINKED ISSUE #1458: Install skycell_asn as a script and update docs
_Issue [RCAL-936](https://jira.stsci.edu/browse/RCAL-936) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Add skycell_asn to pyproject.toml so that it will be installed as a script.

ddavis-stsci:
Resolves [RCAL-936](https://jira.stsci.edu/browse/RCAL-936)

Closes #1458

This PR updates the read the docs to add information about the skycell_asn script to generate associations based on sky cells in the patch table from INS. 

The also updates the pyproject.toml file to install skycell_asn as script. 

## Tasks
  - [N/A] update or add relevant tests
  - [N/A] update relevant docstrings and / or `docs/` page
  - [N/A] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job ([click here for instructions](https://github.com/spacetelescope/RegressionTests/blob/main/docs/running_regression_tests.md))

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/docs/roman/associations/skycell_asn.rst
+++ b/docs/roman/associations/skycell_asn.rst
@@ -83,7 +83,7 @@ where the skycell name can vary based on the location on the celestial sphere an
 that this is resampled 2-d imaging data. The release product name can be changed from the default
 by adding the optional argument --release-product <new name> to the command line. 
 	
-An analogous command to generate the pass level products, again setting segment to 003 to only select
+An analogous command to generate the pass level products, again setting observation to 003 to only select
 the F158 filter. 
 
 .. code-block:: text
--- a/docs/roman/associations/skycell_asn.rst
+++ b/docs/roman/associations/skycell_asn.rst
@@ -26,11 +26,11 @@ for the output associations. To get a complete list of options you can run the c
 
 .. code-block:: python
 
-		skycell_asn -h 
+		skycell_asn -h
 
 
 By knowing the structure of the conventional file name you can generate custom associations based
-on the information included in the file names. 
+on the information included in the file names.
 The current naming convention for the input files consists of a Visit_ID + Exposure_ID + detector + suffix
 Where the Visit_ID is a 19 digit string
 PPPPPCCAAASSSOOOVVV, with
@@ -38,7 +38,7 @@ PPPPPCCAAASSSOOOVVV, with
 PPPPP = Program number
 CC = Execution plan number
 AAA = Pass number (within execution plan)
-SSS = Segment Number (within pass)  
+SSS = Segment Number (within pass)
 OOO = Observation number (within the leg, not to be confused with the Observation ID)
 VVV = Visit Number (within the observation)
 
@@ -81,10 +81,10 @@ or for the selections above
 
 where the skycell name can vary based on the location on the celestial sphere and the i2d indicates
 that this is resampled 2-d imaging data. The release product name can be changed from the default
-by adding the optional argument --release-product <new name> to the command line. 
-	
+by adding the optional argument --release-product <new name> to the command line.
+
 An analogous command to generate the pass level products, again setting observation to 003 to only select
-the F158 filter. 
+the F158 filter.
 
 .. code-block:: text
 
--- a/docs/roman/associations/skycell_asn.rst
+++ b/docs/roman/associations/skycell_asn.rst
@@ -57,13 +57,13 @@ rPPPPPCCAAASSSOOOVVV_eeee_detector_suffix.asdf
 To give a more concrete example we'll use the APT example program for the
 High Latitude Wide Angle Survey (hlwas). The program is 00991, the execution plan number is 01,
 there are 12 passes, 169 segments, 5 observations and a various number of visits.
-For this example we want to select a single filter, say F158, and that is segment 003.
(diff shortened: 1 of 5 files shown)


docs/roman/associations/skycell_asn.rst
```diff
@@ -0,0 +1,91 @@
...
+The suffix indicates the processing level of the file and to create products based on the
+skycells the suffix should generally be 'cal'.
+
+and the file name is constructed as,
+rPPPPPCCAAASSSOOOVVV_eeee_detector_suffix.asdf
+
+To give a more concrete example we'll use the APT example program for the
+High Latitude Wide Angle Survey (hlwas). The program is 00991, the execution plan number is 01,
+there are 12 passes, 5 segments, 4 observations and a various number of visits.
+For this example we want to select a single filter, say F158, and that is segment 003.
```
schlafly, line 60: I'm not sure I follow this example.  When I look at program 991 and pass 1 segment 3, I see a lot of exposures in different filters.  Specifically, each visit is a linegap 4 dither, and those are grouped into observations of 32 pointings in a mosaic, and those observations are grouped into a segment with the F106, F129, F158, F184, and GRISM elements.

So I think the HLWAS example program has:
- 12 passes (but note there are really only 2 passes of imaging, which are the ones we care about and the other passes are purely spectroscopic)
- 169 segments per spectroscopic pass (basically, a big mosaic covering a lot of the sky)
- 5 observations per segment (F106, F129, F158, F184, GRISM)
- 32 visits per observation (a smaller mosaic covering a patch)
- 4 exposures per visit (linegap 4 dither)
So to select out only F158 I think you need to get observation 3.

It does look to me like it's required that all exposures in the same "observation" use the same filter, which I hadn't appreciated before.

docs/roman/associations/skycell_asn.rst
```diff
@@ -0,0 +1,91 @@
...
+where the skycell name can vary based on the location on the celestial sphere and the i2d indicates
+that this is resampled 2-d imaging data. The release product name can be changed from the default
+by adding the optional argument --release-product <new name> to the command line.
+
+An analogous command to generate the pass level products, again setting segment to 003 to only select
+the F158 filter.
+
+.. code-block:: text
+
+		skycell_asn r0099101???003*_*_cal.asdf  -o  r0099101001003001001 --product-type pass
```
schlafly, line 91: I think the right example here is more like:
skycell_asn r0099101001???003*_*_cal.asdf -o r0099101001 --product-type pass
i.e., making the path specify both the execution plan and the pass (01001) and then using the three ??? to go over all segments before specifying observation 003 to get a particular filter.  Then the output root should include the activity / pass information but nothing else, and the script figures out the filter?

docs/roman/associations/skycell_asn.rst
ddavis-stsci: I've updated the numbers, I was trying to simplify this for the docs. Is there another program you know of that might be clearer?

docs/roman/associations/skycell_asn.rst
ddavis-stsci: good catch, I updated this from the visit and forgot to change that.

docs/roman/associations/skycell_asn.rst
schlafly: The HLWAS makes sense because it's one people will have heard of.  There are much smaller programs that one could use but then I just think they will be unknown to people.  In general, I am not sure how well publicized the example APT programs are, but I did think it was useful to have a "worked example"; this was a reminder for me that the "execution plan" stuff is a weird detail that is pretty extraneous to our planning.  It's possible we would want to ?? over that as well, but then I don't know what base name I would use, so I didn't suggest anything there.

schlafly: The new version still selects segment 3, but I think you need observation 3.  i.e., the names need to be more like r0099101001001003001_*_cal.asdf, likewise the base names.

ddavis-stsci:
updated segment to observation

docs/roman/associations/skycell_asn.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. code-block:: text
+
+	r0099101001001003001_r257dp63x98y83_visit_F158_prompt_i2d_asn.json
+
+where the skycell name can vary based on the location on the celestial sphere and the i2d indicates
+that this is resampled 2-d imaging data. The release product name can be changed from the default
+by adding the optional argument --release-product <new name> to the command line.
+
+An analogous command to generate the pass level products, again setting segment to 003 to only select
```
schlafly, line 86:
```suggestion
An analogous command to generate the pass level products, again setting observation to 003 to only select
```

schlafly (approved):
I marked one more case of segment -> observation in line, otherwise looks good.

PR #1484, opened 2024-10-31
RCAL-937: update s_region after running TweakRegStep.

### LINKED ISSUE #1455: Update s_region following tweakreg
I think tweakreg currently updates only the image WCS but not the s_region.  The s_region is used to determine the boundaries of the CCDs, e.g., in the MAST viewer, and we should update it when we update the WCS.

I think the WCS gets replaced here:
https://github.com/spacetelescope/romancal/blob/main/romancal/tweakreg/tweakreg_step.py#L273
but I think that's it.

Vaguely relatedly, we don't update other WCS keywords like ra/dec_ref, roll_ref.  I don't think there is a good way for us to update those in the current scheme, though, and still be able to reproduce the WCS, so I don't think we should tackle that at present.

mairanteodoro:
Resolves [RCAL-937](https://jira.stsci.edu/browse/RCAL-937)

Closes #1455 

This PR updates the `meta.wcsinfo.s_region` attribute after running `TweakRegStep`, which so far was only updating the WCS but not the footprint.

### Regression tests
- https://github.com/spacetelescope/RegressionTests/actions/runs/11687286223

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/changes/1484.tweakreg.rst
+++ b/changes/1484.tweakreg.rst
@@ -1,0 +1,1 @@
+Updates s_region after running TweakRegStep successfully.
--- a/romancal/tweakreg/tests/test_tweakreg.py
+++ b/romancal/tweakreg/tests/test_tweakreg.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from typing import Tuple
 
 import numpy as np
-import pysiaf
 import pytest
 import requests
 from astropy import coordinates as coord
@@ -27,7 +26,6 @@ from stcal.tweakreg.astrometric_utils import get_catalog
 from romancal.datamodels import ModelLibrary
 from romancal.tweakreg import tweakreg_step as trs
 from romancal.tweakreg.tweakreg_step import _validate_catalog_columns
-from romancal.assign_wcs import AssignWcsStep
 
 
 class MockConnectionError:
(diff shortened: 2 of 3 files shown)


schlafly (commented):
Thanks, this looks good.  I was a little surprised that the mosaic regtests show some changes (a handful of pixels).  I think this is saying that drizzle / resample are using the s_region to compute which input pixels touch which output pixels, and maybe they don't use quite a large enough margin, so a few boundary pixels were getting slightly clipped and now are not?  I don't think we care, and resample is migrating to stcal soon, but @mcara might be able to say whether we expect the resampled images to change slightly when s_region is very slightly perturbed.

The contexts also all changed; I guess this just means that there was a CRDS context update and we need to okify everything.  Feel free to do that, thanks!

braingram:
FWIW depending on the timing of the regtest runs I merged https://github.com/spacetelescope/stcal/pull/311 and okified those changes.

The run that produced the files for okifying was: https://github.com/spacetelescope/RegressionTests/actions/runs/11682867663
and the run I triggered afterwards showed no failures: https://github.com/spacetelescope/RegressionTests/actions/runs/11684844516

Is it possible the previous run was started after the stcal merge (~3 hrs ago) but before the okify finished which was shortly after the run fished?

I see the latest run for this PR https://github.com/spacetelescope/RegressionTests/actions/runs/11687286223 only shows `s_region` differences.

If none of this is relevant just ignore me :-P

mcara:
> I don't think we care, and resample is migrating to stcal soon, but @mcara might be able to say whether we expect the resampled images to change slightly when s_region is very slightly perturbed.

I do not see how `s_region` can have any _direct_ effect on resample. An indirect effect can happen via a direct effect on computation of the output WCS which in turn can have an effect on resample.

One way to test this is to create a "custom" reference WCS with the old code and modify tests to use that WCS. There should not be any differences then. If you restore the code to again re-compute the output WCS (from `s_region`) and you get test failures, then it is a clear indication that test failures are due to changes in the output WCS.

schlafly:
The current regtest run shows no pixel level data changes, consistent with Mihai's and my expectations.  I reviewed before ~4 days ago, so I don't think the stcal changes should be relevant.  But I guess as the regtests look good now I'm not inclined to try to do too much archaeology to figure out the old small thing I didn't understand.  So LGTM unless anyone else has ideas here!

braingram:
> The current regtest run shows no pixel level data changes, consistent with Mihai's and my expectations. I reviewed before ~4 days ago, so I don't think the stcal changes should be relevant. But I guess as the regtests look good now I'm not inclined to try to do too much archaeology to figure out the old small thing I didn't understand. So LGTM unless anyone else has ideas here!

Do you think it was https://github.com/spacetelescope/stcal/pull/315

schlafly:
Right, that's plausible!  Then it's on me for not noticing that main was also failing with the same issue.  Okay, let's go ahead.

schlafly:
Mairan, please okify the regtests after merging, thanks!

mairanteodoro:
Sorry I missed all the fun here while I was at a doctor's office...

The latest regtest run shows only two failures that are related to the updates to `TweakRegStep` in this PR. I believe that Brett's right, and reverting stcal PR#135 fixed the issue.

I will go ahead and merge this and then okify the regtest run listed in the description of this PR.

PR #1494, opened 2024-11-05
New catalog tests for B16 (aperture photometry, deblending, kernel source detection)

schlafly:
Resolves [RCAL-946](https://jira.stsci.edu/browse/RCAL-946)

This PR addresses adds new tests of the aperture fluxes, deblending, and kernel source detection to the catalog regression tests.  It takes a calibrated L2 file from artifactory as input, and checks that derived output catalogs have aperture fluxes measured.  It also runs that calibrated L2 image with a different detection kernel and source deblending.  It checks the deblended source catalog to verify that it contains more records than the un-deblended source catalog.

It also updates the regtestdata script to make the new files with deblending and a larger kernel.  I uploaded the corresponding new artifacts to artifactory for I&T though the regtest script doesn't currently use those.

These changes were limited to regression tests, and didn't change any public API, so I have not updated any docs / change logs for the moment.

A problem with this PR is that the logs from running pytest currently don't include the relevant "pass" messages.  It looks like when webbpsf goes to build a PSF that it does something to the logging from which the logging does not recover; some kind of pytest / stpipe / webbpsf interaction?  But the actual tests pass.  In a pinch we can probably get the logs to print if we turn off PSF fitting in this source catalog run, though that is a hack.

## Tasks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,7 @@ test = [
     "pytest >8.0.0",
     "pytest-astropy >= 0.11.0",
     "deepdiff",
+    "stpreview>=0.5.1",
 ]
 dev = [
     "romancal[docs,test]",
@@ -69,8 +70,6 @@ dev = [
 ]
 sdp = [
     "stpreview>=0.5.1",
-    "typer",
-    "typing_extensions",
 ]
 
 [project.urls]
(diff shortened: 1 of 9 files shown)


schlafly:
I merged this work into Larry's multiband catalog PR; closing.

PR #1510, opened 2024-11-12
Fix crds keywords

schlafly:
We missed updating the CRDS version & context keywords as part of the B16 L1/L2 metadata refactoring.  This PR updates their names.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     # "rad>=0.21.0,<0.22.0",
     "rad @ git+https://github.com/spacetelescope/rad.git",
     # "roman_datamodels>=0.21.0,<0.22.0",
-    "roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git",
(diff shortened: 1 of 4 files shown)


schlafly:
I did a regression test run here:
https://github.com/spacetelescope/RegressionTests/actions/runs/11803097124
The failures are all ~expected changes in crds.sw_version and crds.context_used.  But okifying won't do the right thing because the L1s also have these fields and so the new files that are being made have both the new and old keywords.  We need to use a new set of L1s that I've prepared in /grp/roman/eschlafly/romanisim/20241112/regtestl1s to fix this.

mairanteodoro (approved):
Looks good to me.

mairanteodoro:
> We need to use a new set of L1s that I've prepared in /grp/roman/eschlafly/romanisim/20241112/regtestl1s to fix this.

I ran all the regtests locally and they all passed, except for one that was expected and will be fixed after okifying it.

ddavis-stsci (approved):
LGTM

mairanteodoro (approved):
Looks good.

PR #1525, opened 2024-11-15
Switch exposure pipeline to use ModelLibrary instead of a list of models

### LINKED ISSUE #1523: exposure pipeline stops processing association on first fully saturated input
If I generate 2 associations using 1 fully saturated and 1 non-saturated input:
- saturated: https://bytesalad.stsci.edu/ui/repos/tree/General/roman-pipeline/dev/WFI/image/r0000101001001001001_0001_wfi01_ALL_SATURATED_uncal.asdf
- unsaturated: https://bytesalad.stsci.edu/ui/repos/tree/General/roman-pipeline/dev/WFI/image/r0000101001001001001_0001_wfi01_uncal.asdf

in 2 different orders and run each through the `ExposurePipeline` I get 2 different results:

- for the association with the saturated model first, elp returns a list with a single model corresponding to the saturated input
- for the association with the saturated model second, elp return a list with 2 models corresponding to the unsaturated and then saturated input

I failed to find a description of this in the documentation.

Is it expected that the elp will stop processing inputs when it encounters the first saturated input?

braingram:
The PR makes a few updates to the exposure pipeline. The major feature is to use a `ModelLibrary` instead of a list of models. This allows:
- removing the custom asn parsing in the pipeline
- the tweakreg (and other step) cal_step status to be correctly marked by stpipe when steps are skipped

This PR also fixes a bug uncovered while working on this PR (#1523) where an all saturated input resulted in:
- the pipeline skipping processing of all subsequence models in the input association
- failing to return an all zeroed ImageModel

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/11861738024
show 2 expected failures.
The first:
```
AssertionError: 
  Diff report for:
      result file: /runner/_work/_temp/pytest_basetemp/popen-gw0/test_pipeline_suffix0/r0000101001001001001_0001_wfi01_star.asdf
          model type: ImageModel
      truth file: /runner/_work/_temp/pytest_basetemp/popen-gw0/test_pipeline_suffix0/truth/r0000101001001001001_0001_wfi01_star.asdf
          model type: ImageModel

{'dictionary_item_added': ["root['roman']['meta']['asn']", "root['roman']['meta']['exptype']", "root['roman']['meta']['group_id']"],
'values_changed': {"root['roman']['meta']['cal_step']['tweakreg']": {'new_value': 'SKIPPED',
'old_value': 'INCOMPLETE'}}}
assert False
```
is a byproduct of using ModelLibrary. The output files for elp now contain asn information and have tweakreg correctly marked as 'SKIPPED' for [test_pipeline_suffix](https://github.com/spacetelescope/romancal/blob/281aa556a388363f98ed8a69ad1de46ef4b78297/romancal/regtest/test_wfi_image_pipeline.py#L283).

The second failure is due to this PR fixing #1523. `test_all_saturated_against_truth` fails since this PR restores the old behavior where an all saturated input produces a cal files with all zeros and all subsequent steps skipped. As described in #1523 several updates to the elp resulted in this behavior being lost and `create_fully_saturated_zeroed_image` (which was added in https://github.com/spacetelescope/romancal/pull/541) being unused. With this PR `create_fully_saturated_zeroed_image` is once again used. To hopefully avoid a future regression I added some additional tests for the "all saturated" input file that:
- checks that the output is an `ImageModel`
- checks that the output arrays are all zeros

Finally I added a few unit tests that run the elp with all steps skipped and no reference files. These are relatively fast (good candidates for unit tests) and check that when provided an input model the pipeline returns a model etc (and not a list as it does on main).

Closes #1523

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/regtest/test_wfi_image_pipeline.py
```diff
@@ -300,5 +300,5 @@ def test_pipeline_suffix(rtdata, ignore_asdf_paths):
         assert model.meta.cal_step.flat_field == "COMPLETE"
         assert model.meta.cal_step.photom == "COMPLETE"
         assert model.meta.cal_step.source_detection == "COMPLETE"
-        assert model.meta.cal_step.tweakreg == "INCOMPLETE"
+        assert model.meta.cal_step.tweakreg == "SKIPPED"
```
braingram, line 303: This `assert` isn't currently reached in the regtests due to the `compare_against_truth` failing:
```
FAILED romancal/regtest/test_wfi_image_pipeline.py::test_pipeline_suffix - AssertionError: 
  Diff report for:
      result file: /runner/_work/_temp/pytest_basetemp/popen-gw2/test_pipeline_suffix0/r0000101001001001001_0001_wfi01_star.asdf
          model type: ImageModel
      truth file: /runner/_work/_temp/pytest_basetemp/popen-gw2/test_pipeline_suffix0/truth/r0000101001001001001_0001_wfi01_star.asdf
          model type: ImageModel
  
  {'dictionary_item_added': ["root['roman']['meta']['asn']", "root['roman']['meta']['exptype']", "root['roman']['meta']['group_id']"],
   'values_changed': {"root['roman']['meta']['cal_step']['tweakreg']": {'new_value': 'SKIPPED',
                                                                        'old_value': 'INCOMPLETE'}}}
```
But when 'okified' this change to the assert will be needed.

schlafly:
Thanks.  This looks good.

You also deleted some orphaned fully saturated exposure handling here.  I think following your investigation we probably do want to keep and start using that code again, rather than deleting it?
https://github.com/spacetelescope/romancal/issues/1523
Maybe just to say that deleting create_fully_saturated_zeroed_image needn't be part of this PR.

@ddavis-stsci , this refactors some of the pipeline handling to use ModelLibrary more uniformly, so you may want to weigh in here.

braingram:
Converted to draft to attempt to also address #1523 and some other elp cleanup.

ddavis-stsci:
Besides the all saturated case not working it seems fine.

braingram:
> Besides the all saturated case not working it seems fine.

Thanks for giving this a look. Unfortunately it was draft at the time and not yet finished. The major changes that occurred after your comment were:
- fixing the pipeline handling of all saturated input (this results in regression test differences that will need to be okified)
- adding unit tests for the elp input/output types

I re-requested your review in light of these changes.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -73,123 +68,89 @@ def process(self, input):
...
+                    # also can't run tweakreg.
+                    result.meta.cal_step.tweakreg = "SKIPPED"
+                lib.shelve(result, model_index)
 
         # Now that all the exposures are collated, run tweakreg
         # Note: this does not cover the case where the asn mixes imaging and spectral
         #          observations. This should not occur on-prem
-        result = self.tweakreg.run(results)
+        if not any_saturated:
+            self.tweakreg.run(lib)
```
schlafly, line 141: I don't like thinking about the fully saturated case because I expect it to never occur or to be so bad, should it occur, that we don't really care about what we do.

But conceptually if I try to imagine a case where one detector is fully saturated but the others are all fine the "right thing to do" would be to run tweakreg using the source catalogs from the other detectors.  The model that gets fit right now is very rigid and would get well constrained even if only a few detectors were available.

i.e., I guess that I would have imagined the flow being more like: tweakreg loops over the association gathering the catalogs.  If it doesn't find one, it emits some warning message.  If it doesn't find _any_, it gives up and skips or something.  Then the pipeline doesn't need to specially track the any_saturated case here.

schlafly (commented):
This looks good to me.  I left some comments about the tweakreg / all saturated interaction.

Ultimately I think this case is rare enough and weird enough that probably we should be suspect of the whole exposure if any individual SCA is fully saturated, so I don't think it's a big deal.  However, it surprised me a little to skip tweakreg for the whole array because of a single detector.

romancal/pipeline/exposure_pipeline.py
braingram: That sounds good to me.

Given that `source_catalog` is skipped for a fully saturated image I think one option (if it sounds good to you) would be to update tweakreg to check if the image is fully saturated, if so, don't check for the source catalog entry and skip that model. Currently tweakreg throws an exception if an input model doesn't have a source catalog:
https://github.com/spacetelescope/romancal/blob/78c7dc41093f8fb76fe049682ce56de3cad4c488/romancal/tweakreg/tweakreg_step.py#L159
so some sort of change to that will be needed (either turn that into a warning and skip that model or check if it's all saturated).

schlafly: Yeah, I certainly don't consider this a regression, just trying to think about what we want.

Maybe best to go with your current approach to avoid having tweakreg know about full saturation, keeping all of the fully saturated logic in one place.  I hope this case just never actually occurs in practice, so it doesn't matter.

FWIW, I did discuss this, and we could not come up with a situation where this would actually occur for a science exposure.

schlafly (approved):
Thanks Brett, approving!

braingram:
Thanks! I rebased and started a new regtest run here:
https://github.com/spacetelescope/RegressionTests/actions/runs/12188643274
@schlafly feel free to merge this whenever the regtests finish and you're ready to test the new okify script (or let me know if you'd like me to do that).

PR #1527, opened 2024-11-18
Clean up bounding box assignment

WilliamJamieson:
spacetelescope/gwcs#522 adds a warning to flag instances where bounding boxes that GWCS uses are ordered using the `C` (y, x) ordering rather than the `F` ordering the GWCS defaults to. GWCS handles the alternate ordering by default however tuple assignment to the forward transform before creating the GWCS causes astropy to assume a `C` ordering while if one assigns the same tuple directly to the wcs, GWCS will assume a `F` order. Due to this source of issues GWCS now throws a warning when it discovers an assigned bounding box with the `C` ordering as GWCS will always attach and handle bounding boxes specifically in the `F` ordering.

This PR updates `romancal` so that bounding boxes are always attached with the tuple correctly ordered and flagged as `F` order.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/assign_wcs/assign_wcs_step.py
+++ b/romancal/assign_wcs/assign_wcs_step.py
@@ -8,6 +8,7 @@ import logging
 import gwcs.coordinate_frames as cf
 from astropy import coordinates as coord
 from astropy import units as u
+from astropy.modeling import bind_bounding_box
 from gwcs.wcs import WCS, Step
 from roman_datamodels import datamodels as rdm
 
@@ -141,7 +142,7 @@ def wfi_distortion(model, reference_files):
     transform = dist.coordinate_distortion_transform
 
     try:
-        bbox = transform.bounding_box
+        bbox = transform.bounding_box.bounding_box(order="F")
     except NotImplementedError:
         # Check if the transform in the reference file has a ``bounding_box``.
         # If not set a ``bounding_box`` equal to the size of the image after
@@ -149,9 +150,10 @@ def wfi_distortion(model, reference_files):
         bbox = None
     dist.close()
 
-    if bbox is None:
-        transform.bounding_box = wcs_bbox_from_shape(model.data.shape)
-    else:
-        transform.bounding_box = bbox
+    bind_bounding_box(
+        transform,
+        wcs_bbox_from_shape(model.data.shape) if bbox is None else bbox,
+        order="F",
+    )
 
     return transform
--- a/changes/1527.general.rst
+++ b/changes/1527.general.rst
@@ -1,0 +1,1 @@
+Perform bounding box assignment inline with the ordering that GWCS prefers.
--- a/romancal/conftest.py
+++ b/romancal/conftest.py
@@ -198,7 +198,6 @@ def _create_wcs(input_dm, shift_1=0, shift_2=0):
 
     # create necessary transformations
     distortion = Shift(-shift_1) & Shift(-shift_2)
-    distortion.bounding_box = ((-0.5, shape[-1] + 0.5), (-0.5, shape[-2] + 0.5))
     tel2sky = pointing.v23tosky(input_dm)
 
     # create required frames
@@ -219,6 +218,7 @@ def _create_wcs(input_dm, shift_1=0, shift_2=0):
     ]
 
     wcs_obj = wcs.WCS(pipeline)
+    wcs_obj.bounding_box = ((-0.5, shape[-2] + 0.5), (-0.5, shape[-1] + 0.5))
 
     input_dm.meta["wcs"] = wcs_obj
 
--- a/romancal/tweakreg/tests/test_astrometric_utils.py
+++ b/romancal/tweakreg/tests/test_astrometric_utils.py
@@ -254,7 +254,6 @@ def create_wcs_for_tweakreg_pipeline(input_dm, shift_1=0, shift_2=0):
 
     # create necessary transformations
     distortion = Shift(-shift_1) & Shift(-shift_2)
-    distortion.bounding_box = ((-0.5, shape[-1] + 0.5), (-0.5, shape[-2] + 0.5))
     tel2sky = _create_tel2sky_model(input_dm)
 
     # create required frames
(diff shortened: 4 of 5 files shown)


WilliamJamieson:
Regression tests pass. Note that these tests were specifically run with pytest turning all the bounding box warnings from GWCS into errors. https://github.com/spacetelescope/RegressionTests/actions/runs/11916504320

schlafly (approved):
Looks good to me.

FWIW, bind_bounding_box is new to me.  Would you mind saying a few words about when this is needed?  i.e., in parts of this PR you do wcs_obj.bounding_box = box, but for adding it to the transform you use the bind_bounding_box routine rather than the previous simpler-looking direct assignment.  Relatedly, it looks like when you assign to the wcs object you don't specify an order (presumably 'F' is assumed), but you are more explicit when assigning to the transform.  What motivates being more explicit in one case than in the other?

WilliamJamieson:
The use of tuples of tuples to represent bounding boxes for 2D+ circumstances has annoying consequences due to choices within `astropy`. For `astropy` models, when setting a bounding box via the tuple of tuples method, bounding boxes are assumed to be ordered in **reverse** input order, i.e. if the model has inputs ordered `x, y` then then it must be the case that for assignment
```python
model.bounding_box = ((y_min, y_max), (x_min, x_max))
```
This is the so called `"C"` ordering. By contrast, GWCS assumes the bounding box is ordered in the coordinate input ordering, so if they are say `ra, dec` then for assignment
```python
wcs.bounding_box = ((ra_min, ra_max), (dec_min, dec_max))
```
This is the so called `"F"` ordering. Hopefully it is clear that when building a WCS using GWCS it can be very very easy to accidentally reverse a bounding box from what you meant. There are two possible mistakes

1. Building the bounding box using the GWCS assumed ordering, but attaching it to the forward transform **before** passing the transform to the WCS.
2. Building the bounding box using the assumed `astropy` ordering, but attaching it to the final WCS.

In either case, one ends up with the bounding box in reverse order to what it should be. The bugs resulting from this can be rather difficult to find (ask me how I know), especially when the box is nearly but not total square. Hence, GWCS's new attempts to warn users that there maybe a possible issue.

Note that GWCS uses the "first" `forward_transform` to store the WCS's bounding box. Since this is an `astropy` model, everything ultimately gets passed into `astropy`. `astropy` internally handles bounding boxes not as tuples of tuples, but as abstracted objects which associate an "interval" (min/max) with each input variable. This means that programmatically once a bounding box is attached to a model, the "ordering" does not matter; however, it does matter when using tuples of tuples to set the bounding box. There are several methods for actively setting a bounding box to an `astropy` model using the `F` ordering, they just do not involve the `model.bounding_box =`. Once such method is the `bind_bounding_box` function, which exists as a simple way to accomplish using a `F` ordered tuple to attach a bounding box to them model and indicate that it was an `F` ordered box to start with, which is how GWCS implements the `wcs.bounding_box=` operation.

The method GWCS uses for issuing the warning for a possible bounding box issue is interrogation of the bounding box's order when it grabs said bounding box from the first forward transform. If that order is '"C"' (is dimension > 2, and not a default bounding box from the model class itself), GWCS knows that the bounding box was set outside of its control meaning the user had to remember to reverse it.

> Looks good to me.
> 
> FWIW, bind_bounding_box is new to me. Would you mind saying a few words about when this is needed? i.e., in parts of this PR you do wcs_obj.bounding_box = box, but for adding it to the transform you use the bind_bounding_box routine rather than the previous simpler-looking direct assignment. Relatedly, it looks like when you assign to the wcs object you don't specify an order (presumably 'F' is assumed), but you are more explicit when assigning to the transform. What motivates being more explicit in one case than in the other?

As to why `bind_bounding_box` was used in some places while in others I moved the bounding box attachment around? If you look closely `bind_bounding_box` is used in cases where the forward transform is being built along side the bounding box separate from where the WCS is built, ie in a function that generates the forward transform, which is then passed to the WCS. In these cases, it is rather awkward to set the bounding box directly on the WCS instead, so the `bind_bounding_box` is used to explicitly set the bounding box using the `"F"` ordering. In the cases, where I moved bounding box assignment around, were those where the bounding box object was generated in the same context as the WCS was being built, which means I could simply reverse the bounding box tuple order and then attach it to the WCS directly (from which GWCS would then do all the necessary stuff under the hood to make sure it was attached as an order `"F"` bounding box.

Ideally, the bounding box should always be computed in such a way we can pass directly to the transform; however, that was not always the easiest thing to do without more significant changes. I am planning to clean up all the WCS stuff in `romancal` once we have finalized the GWCS API for GWCS release 1.0, so making significant code changes now to do stuff more cleanly does not really make sense to do significant changes in this PR.

romancal/assign_wcs/assign_wcs_step.py
```diff
@@ -141,17 +142,18 @@ def wfi_distortion(model, reference_files):
...
 
-    if bbox is None:
-        transform.bounding_box = wcs_bbox_from_shape(model.data.shape)
-    else:
-        transform.bounding_box = bbox
+    bind_bounding_box(
+        transform,
+        wcs_bbox_from_shape(model.data.shape) if bbox is None else bbox,
+        order="F",
+    )
```
WilliamJamieson, line 157: This is a case where the GWCS warning identified a bug. It just did not get picked up by any tests...

Note https://github.com/spacetelescope/romancal/blob/281aa556a388363f98ed8a69ad1de46ef4b78297/romancal/assign_wcs/utils.py#L20-L35

Reports to create a bounding box in the `x, y` ordering **but** the `transform.bounding_box =` was making the assumption this was a `y, x` box.

With this change it is now being assigned correctly with an `x, y` ordering.

schlafly:
Ah, good, astropy.modeling uses the other convention and has this box convention---thanks, that makes sense.

> As to why `bind_bounding_box` was used in some places while in others I moved the bounding box attachment around? If you look closely `bind_bounding_box` is used in cases where the forward transform is being built along side the bounding box separate from where the WCS is built, ie in a function that generates the forward transform, which is then passed to the WCS. 

Yup, great, this makes sense, thanks!

PR #1556, opened 2024-12-09
Hackday mteodoro sky variance readnoise

mairanteodoro:

Closes #

This PR addresses ...

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/resample/resample_utils.py
```diff
@@ -401,3 +416,19 @@ def resample_range(data_shape, bbox=None):
...
+        else ModelLibrary([input_models])
+    )
+    with input_models:
+        ref_img = input_models.borrow(index=0)
+        var_sky = np.zeros_like(ref_img.data)
+        input_models.shelve(model=ref_img, index=0)
+        for i, img in enumerate(input_models):
+            var_sky += img.var_rnoise + img.var_poisson / img.data * np.median(img.data)
+            img["var_sky"] = var_sky
+            input_models.shelve(img, i, modify=True)
```
braingram, line 434: I think this pattern will result in a different `var_sky` if this step is run with an on_disk vs in_memory ModelLibrary.
Let's say we have 3 models in the library. The first time through this loop `var_sky` will be set as in the code above and then assigned to the first model under the `var_sky` attribute. The `shelve` will write out this array if the library is "on_disk" freezing the contents of `var_sky` so that for the first model it's `var_sky` will only include contributions from the first model. If the library is in memory the `var_sky` array for model 1 will stay in memory and be updated the 2nd (and 3rd) times through the loop. So for an in memory library the `var_sky` for each model will be determined by all models but for an on disk library the `var_sky` will only be determined by that model (and any previous).
If the goal is to have `var_sky` be a combination of all models the assignment to `img["var_sky"]` could be broken into a separate loop or alternatively just kept in memory (since it would be the same for all models).

mairanteodoro: Thanks for the reminder that we also have an "on disk" mode. ;)
I will reformat this to account for that.

mairanteodoro: Actually, `add_var_sky_array()` is meant to add the `var_sky` array to the L2 images (in memory) so that they can be combined later on by drizzle using `weight_type="ivsky"`.

braingram:
I follow the addition of the new weight type but I'm not sure I see the connection between the ticket and resampling each computed `var_sky` into a `var_sky` added to the mosaic model.

If the goal is to save `var_sky` in the mosaic model would you open a rad and roman_datamodels PRs to update the schema to include `var_sky`?

romancal/resample/resample_utils.py
```diff
@@ -401,3 +416,31 @@ def resample_range(data_shape, bbox=None):
...
+
+    Returns
+    -------
+    None
+    """
+    input_models = (
+        input_models
+        if isinstance(input_models, ModelLibrary)
+        else ModelLibrary([input_models])
+    )
```
braingram, line 438:
```suggestion
    input_models = (
        input_models
        if isinstance(input_models, ModelLibrary)
        else ModelLibrary([input_models])
    )
```
If the function only accepts a `ModelLibrary` (based on the type annotation) then is this necessary?

romancal/resample/resample_utils.py
```diff
@@ -401,3 +416,31 @@ def resample_range(data_shape, bbox=None):
...
+    with input_models:
+        ref_img = input_models.borrow(index=0)
+        input_models.shelve(model=ref_img, index=0)
+        for i, img in enumerate(input_models):
+            img["var_sky"] = img.var_rnoise + img.var_poisson / img.data * np.median(
+                img.data
```
schlafly, line 444: Let's protect against the image.data == 0 case here.

schlafly (commented):
I think this looks conceptually right; thanks!  Can you put a couple of example files on /grp/roman somewhere, together with an invocation (run strun roman::resample blah.asn --weight_type skyivm ??) so that I can easily poke at some example input and output files?

mairanteodoro:
> If the goal is to save `var_sky` in the mosaic model would you open a rad and roman_datamodels PRs to update the schema to include `var_sky`?

Yes, that's the idea.

romancal/resample/resample_utils.py
```diff
@@ -163,6 +164,20 @@ def build_driz_weight(
...
+            and model.var_sky.shape == model.data.shape
+        ):
+            with np.errstate(divide="ignore", invalid="ignore"):
+                inv_sky_variance = model.var_sky**-1
+            inv_sky_variance[~np.isfinite(inv_sky_variance)] = 1
+        else:
+            warnings.warn(
+                "var_rnoise and var_poisson arrays are not available. Setting drizzle weight map to 1"
+            )
+        result = inv_sky_variance * dqmask
```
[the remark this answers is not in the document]
mairanteodoro: Nice catch! Thanks!
I completely forgot to add the default `inv_sky_variance` array.

romancal/resample/resample_utils.py
mairanteodoro: No, it is not.

Initially, the method was supposed to handle both `ModelLibrary` and single datamodels, but eventually I thought that it would be better to sticking with the former only.

docs/roman/resample/arguments.rst
```diff
@@ -78,9 +78,13 @@ image.
...
     will be determined per-pixel using the inverse of the read noise
     (VAR_RNOISE) array stored in each input image. If the VAR_RNOISE array does
-    not exist, the variance is set to 1 for all pixels (equal weighting).
+    not exist, the weight is set to 1 for all pixels (equal weighting).
     If `weight_type=exptime`, the scaling value will be set equal to the
     exposure time found in the image header.
+    If `weight_type=ivsky`, the scaling value will be determined per-pixel
+    using the inverse of the sky variance (VAR_SKY) array calculated in the
+    resample step for each input image. If the VAR_SKY array does
+    not exist, the weight is set to 1 for all pixels (equal weighting).
```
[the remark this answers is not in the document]
mairanteodoro: Done.

docs/roman/resample/arguments.rst
```diff
@@ -77,10 +77,20 @@ image.
...
+    If `weight_type=ivsky`, the scaling value will be determined per-pixel
+    using the inverse of the sky variance (``VAR_SKY``) array calculated in the
+    resample step for each input image. ``VAR_SKY`` is given by the following equation:
+
+    .. math::
+
+        \text{VAR\_SKY} = \text{VAR\_RNOISE} + \text{VAR\_POISSON} \, \frac{ med(\text{DATA}) }{ \text{DATA} },
+
+    where :math:`\text{DATA}` and :math:`med(\text{DATA})` correspond to the data array and its median, respectively.
+    If the ``VAR_SKY`` array does not exist, the weight is set to 1 for all pixels (equal weighting).
```
braingram, line 93: Should this say if the `VAR_RNOISE` and/or `VAR_POISSON` array does not exist?

romancal/resample/resample_utils.py
```diff
@@ -402,3 +419,29 @@ def resample_range(data_shape, bbox=None):
...
+
+    Returns
+    -------
+    None
+    """
+    with input_models:
+        ref_img = input_models.borrow(index=0)
+        input_models.shelve(model=ref_img, index=0)
+        for i, img in enumerate(input_models):
+            if np.all(img.data != 0):
```
schlafly, line 441: Sorry, my fault.  This is good protection for the img.data = 0 case, but probably too good.

In an average image we should expect some pixels with flux == 0; say they're masked or something and are marked as DO_NOT_USE.  Then var_poisson will also be 0 or NaN or something.  We need to handle this case without giving up entirely.  Let's handle this for now via something like:
```
m = img.data != 0
img['var_sky'][m] = img.var_rnoise[m] + img.var_poisson[m] / img.data[m] * np.median(img.data)
img['var_sky'][~m] = img.var_rnoise[~m]
```
If we wanted to do better we should probably do only a median over the pixels that are "good", so something like
np.median(img.data[(img.dq & DO_NOT_USE) != 0]), for example.

romancal/resample/resample_utils.py
```diff
@@ -163,6 +164,22 @@ def build_driz_weight(
...
         result = exptime * dqmask
+    elif weight_type == "ivsky":
+        if (
+            hasattr(model, "var_sky")
+            and model.var_sky is not None
+            and model.var_sky.shape == model.data.shape
+        ):
+            with np.errstate(divide="ignore", invalid="ignore"):
+                inv_sky_variance = model.var_sky**-1
+            inv_sky_variance[~np.isfinite(inv_sky_variance)] = 1
```
schlafly, line 175: Let's make the inv_sky_variance for these pixels 0 rather than 1.
inverse_variance = 0 -> no weight on these pixels
inverse_variance = 1 -> some amount of weight that depends on units...

docs/roman/resample/arguments.rst
braingram: Also I'm not sure the escapes are needed for the underscores in the variable names. They may be causing the rendering to look odd in the docs:
https://roman-pipeline--1556.org.readthedocs.build/en/1556/roman/resample/arguments.html#step-arguments

romancal/resample/resample.py
```diff
@@ -396,25 +397,25 @@ def resample_many_to_one(self):
...
         exptime_tot = self.resample_exposure_time(output_model)
 
-        # TODO: fix unit here
         output_model.err = np.sqrt(
             np.nansum(
                 [
                     output_model.var_rnoise,
                     output_model.var_poisson,
                     output_model.var_flat,
+                    output_model.var_sky,
```
braingram, line 411: Isn't `var_sky` highly correlated with `var_rnoise` and `var_poisson which violates the assumptions for this addition in quadrature?

schlafly: Oh, wow, good catch, thanks.  Let's just not include var_sky in the err sum; i.e., delete the added line 411.

The alternative would be to delete the var_rnoise / var_poisson lines, but that would not be backward compatible.  I agree that we don't want to double-count the var_rnoise & var_poisson from the sky.

braingram: Is it still useful to include a resampled `var_sky` in the output mosaic? If not, this PR could be simplified to only compute `var_sky` for each input (during the weight calculation) if:
- sky weighting is selected
- the input doesn't contain a `var_sky` (I'm not sure under what case the inputs would have one)

schlafly: It's still helpful.  IMO Harry and I think it may make sense to remove var_rnoise & var_poisson and only have something like var_sky, but we're not ready to make that step for this PR.

braingram: Thanks. Remove `var_rnoise` from the `err` computation or remove the `var_rnoise` array from the output mosaic (or both)?

I'm looking at this PR with an eye towards the consolidation of resample code into stcal. The current PR for that https://github.com/spacetelescope/stcal/pull/320/ doesn't contain options for `var_sky` or for easily adjusting the `err` computation by using a different combination of var components.

mairanteodoro: Yup, that was one thing that I noticed after adding it; it shouldn't be there.

schlafly: Ah, understood.  If we were to make a change to remove var_rnoise & var_poisson, the proposal would be to remove them from the output mosaic and to replace err with sqrt(var_sky).  Note that this is not a formal proposal, obviously!

Here is the thinking: a lot of people use exposure time sampling in preference to inverse-variance-readnoise weighting because the images are background limited (i.e., var_sky dominates var_rnoise).  That's still not right in cases where something like a CR has hit a pixel; exptime weighting would weight CR-affected pixels and "pristine" pixels equally, but var_sky should appropriately deweight the CR-affected pixels.

var_rnoise is used for weighting because it is sensitive to things like whether a CR affected a pixel, but it's naturally of course also sensitive to variations in the read noise from pixel to pixel, which are not important if the read noise is a small part of the total variance.

var_poisson is rarely used for weighting or fitting because it depends on the fit to the flux in each pixel, and you can get biases in your derived photometry if you, e.g., downweight pixels that fluctuate high due to having larger var_poisson.

var_sky has both var_poisson & var_rnoise contributions and is sensitive to things like CRs.  It doesn't have problematic dependence on the per-pixel fluxes since the median is over a big chunk of the image.  So it's in a few ways a nice improvement over what we're doing now.

Note that I think Webb would benefit equally to Roman for all of this, but I appreciate that that is another fish to fry.

docs/roman/resample/arguments.rst
braingram: Weird, RTD is showing something different. Do you see something like the following:
Screenshot 2025-01-29 at 5 35 32 PM
at 
https://roman-pipeline--1556.org.readthedocs.build/en/1556/roman/resample/arguments.html#step-arguments

romancal/resample/resample_utils.py
mairanteodoro: Done.

romancal/resample/resample_utils.py
mairanteodoro: Done.

romancal/resample/resample.py
mairanteodoro: Removed. Thanks for catching that, @braingram!

docs/roman/resample/arguments.rst
mairanteodoro: Yeah, I noticed that too.

mairanteodoro: Maybe it has something to do with the build command?
- locally, `make html` runs `sphinx-build -b html -d _build/doctrees   . _build/html`
- during CI we run `sphinx -T -W --keep-going -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html`

romancal/resample/resample_utils.py
```diff
@@ -402,3 +419,32 @@ def resample_range(data_shape, bbox=None):
...
+        for i, img in enumerate(input_models):
+            try:
+                ok_data = img.data != 0
+                img["var_sky"] = np.empty_like(img.data)
+                img["var_sky"][ok_data] = img.var_rnoise[ok_data] + img.var_poisson[
+                    ok_data
+                ] / img.data[ok_data] * np.median(img.data)
+                img["var_sky"][~ok_data] = img.var_rnoise[~ok_data]
+            except (AttributeError, KeyError, TypeError, ValueError) as e:
+                raise ValueError("Input model contains invalid data array.") from e
```
braingram, line 449: Since this raises an exception (if for instance `var_rnoise` isn't available won't this stop resample from proceeding if `ivsky` is selected and `var_rnoise` isn't available? That behavior makes sense to me but is at odds with the documentation (which says thew weight will be 1 for all pixels in this case).

Also I suggest checking for `var_rnoise` and `var_poisson` instead of using the try/except (since the try/except will also trigger for other reasons.

romancal/resample/resample_utils.py
```diff
@@ -402,3 +419,32 @@ def resample_range(data_shape, bbox=None):
...
+    input_models : ModelLibrary
+        A library of models to which the sky variance array will be added.
+
+    Returns
+    -------
+    None
+    """
+    with input_models:
+        ref_img = input_models.borrow(index=0)
+        input_models.shelve(model=ref_img, index=0)
```
braingram, line 439:
```suggestion
```
Is ref_img used?

romancal/resample/resample_utils.py
```diff
@@ -402,3 +419,32 @@ def resample_range(data_shape, bbox=None):
...
+        for i, img in enumerate(input_models):
+            try:
+                ok_data = img.data != 0
+                img["var_sky"] = np.empty_like(img.data)
+                img["var_sky"][ok_data] = img.var_rnoise[ok_data] + img.var_poisson[
+                    ok_data
+                ] / img.data[ok_data] * np.median(img.data)
```
braingram, line 446: If `data` contains a `nan` won't this cause every `var_sky` pixel (for each index wither `data` != 0) to be `nan`?

romancal/resample/resample_utils.py
```diff
@@ -402,3 +419,32 @@ def resample_range(data_shape, bbox=None):
...
+    -------
+    None
+    """
+    with input_models:
+        ref_img = input_models.borrow(index=0)
+        input_models.shelve(model=ref_img, index=0)
+        for i, img in enumerate(input_models):
+            try:
+                ok_data = img.data != 0
+                img["var_sky"] = np.empty_like(img.data)
```
braingram, line 443: I think this could be simplified by starting with something like:
```python
img["var_sky"] = img.var_rnoise.copy()
```
since all pixels are either:
- copied from `var_sky` if ~(img.data != 0)
- computed below

romancal/resample/resample.py
```diff
@@ -86,13 +86,15 @@ def __init__(
             )
 
         self.input_models = input_models
+        # add sky variance array
+        resample_utils.add_var_sky_array(self.input_models)
```
braingram, line 90: This will always re-compute `var_sky` (even if the input contains a `var_sky`). If that's the goal (always have resample compute `var_sky`) what about something like:
- if `weight` is `ivsky` compute `var_sky` for each model
- when resampling `var_sky` compute it (if it wasn't computed before)

That way for non`ivsky` weighting resample won't need to make a separate pass through all the models here (in `add_var_sky_array`).

braingram (commented):
Thanks for updating this.

I left a few more comments and questions.

Would you open a RAD PR to add `var_sky` to the mosaic schema?

romancal/resample/tests/test_resample.py
```diff
@@ -694,15 +695,19 @@ def get_footprint(model, index):
...
+@pytest.mark.parametrize("weight_type", ["ivm", "exptime", "ivsky"])
 def test_resampledata_do_drizzle_default_single_exposure_weight_array(
     exposure_1,
     weight_type,
 ):
     """Test that resample methods return non-empty weight arrays."""
 
+    # adding a few zero flux pixels
+    for i, model in enumerate(exposure_1):
+        model.data[10 + i, 40 - i] = 0
```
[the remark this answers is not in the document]
mairanteodoro: I just added a random pixel with zero flux. I was not interested in the position of the pixel, just that we did have a pixel with zero flux.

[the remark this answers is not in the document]
braingram: I'm still confused by this. What's the need for a random zero flux pixel per image? If I comment out this code the test doesn't fail.

romancal/resample/tests/test_resample.py
```diff
@@ -712,6 +717,8 @@ def test_resampledata_do_drizzle_default_single_exposure_weight_array(
         many_to_one_model = output_models_many_to_one.borrow(0)
         assert np.any(many_to_one_model.weight > 0)
         assert np.any(many_to_many_model.weight > 0)
+        assert many_to_one_model.data[10, 40] == 0
```
braingram, line 720: What is this line testing? The referenced pixel at 10, 40 (seen as the single highlighted pixel in the upper left) falls in the "fill" region of the output image.
Screenshot 2025-02-03 at 1 47 59 PM

PR #1567, opened 2024-12-19
Adjust crval in resample tests to get overlap with input

mcara:
This PR adjusts `crval` value in `test_populate_mosaic_basic_single_exposure` and `test_populate_mosaic_basic_different_observations` unit tests. The original value was 10 degrees off  due to which input images do not have any overlap with output image created with custom WCS parameters. With new changes in GWCS (enabling bounding box for inverse + a fix in stcal) this would lead to tests failing.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/resample/tests/test_resample.py
+++ b/romancal/resample/tests/test_resample.py
@@ -1039,7 +1039,11 @@ def test_populate_mosaic_basic_different_observations(
         rotation=0,
         shape=None,
         crpix=(0, 0),
+<<<<<<< HEAD
         crval=(30, 45),
+=======
+        crval=(10, 0),
+>>>>>>> 001c238 (Adjust crval in resample tests to get overlap with input)
     )
     output_model = maker_utils.mk_datamodel(
         datamodels.MosaicModel, shape=tuple(output_wcs.array_shape)


schlafly:
Thanks.  Mostly for my reference, the issue is that the wfi_scas were populated at (ra, dec) = (10, 0), but the corresponding crvals were populated at (0, 0), as Mihai says.
https://github.com/spacetelescope/romancal/pull/1567/files#diff-3263dc8226f60646b134491bd47fad24fecde7e8904e07bb43253e695e8c385eL174

schlafly (approved):

braingram:
It looks like both of those tests are only looking at the metdata of the resampled result. Would they work with the upcoming changes if `crval` was removed? I don't think that would have any negative impact on the test. It looks like both are only computing the `output_wcs` to get an array shape to use to make an output model and then calling `populate_mosaic_basic`. Perhaps the array shape can just be hard coded?

mcara:
> It looks like both of those tests are only looking at the metadata for a blended model. Would they work with the upcoming changes if `crval` was removed? I don't think that would have any negative impact on the test. It looks like both are only computing the `output_wcs` to get an array shape to use to make an output model and then calling `populate_mosaic_basic`. Perhaps the array shape can just be hard coded?
> 
> EDIT: I was writing this while Eddie was reviewing. The above changes could be a follow-up PR (if suitable).

There are different options:
1.  array shape can be hard-coded;
2. `crpix` can be removed;
3. or have `crpix` "agree" with "crval" so that there are at least some valid data;
4. or just accept this PR.

braingram:
Thanks @mcara I opened an issue to track the follow up changes https://github.com/spacetelescope/romancal/issues/1569

mairanteodoro:
Running the changes in this PR against `https://github.com/mcara/stcal/tree/fix-bbox-shift-make-wcs` still causes the two tests to fail. Removing `crpix` and leaving `crval` (regardless of its value) works.

Since the idea behind those tests is to check for the functionality of `populate_mosaic_basic()`, I suggest that we drop both `crpix` and `crval` (they are computed automatically by `make_output_wcs()` anyways).

romancal/resample/tests/test_resample.py
```diff
@@ -715,7 +715,7 @@ def test_populate_mosaic_basic_single_exposure(exposure_1):
             rotation=0,
             shape=None,
             crpix=(0, 0),
-            crval=(0, 0),
+            crval=(10, 0),
```
mairanteodoro, line 718: Tests are still failing with `crpix=(0,0)` + `crval=(10,0)`. It does work when removing `crpix` alone (regardless of `crval`'s value).

romancal/resample/tests/test_resample.py
```diff
@@ -1014,7 +1014,7 @@ def test_populate_mosaic_basic_different_observations(
         rotation=0,
         shape=None,
         crpix=(0, 0),
-        crval=(0, 0),
+        crval=(10, 0),
```
mairanteodoro, line 1017: Tests are still failing with `crpix=(0,0)` + `crval=(10,0)`. It does work when removing `crpix` alone (regardless of `crval`'s value).

romancal/resample/tests/test_resample.py
```diff
@@ -715,7 +715,7 @@ def test_populate_mosaic_basic_single_exposure(exposure_1):
             rotation=0,
             shape=None,
             crpix=(0, 0),
```
mairanteodoro, line 717:
```suggestion

```

romancal/resample/tests/test_resample.py
```diff
@@ -1014,7 +1014,7 @@ def test_populate_mosaic_basic_different_observations(
         rotation=0,
         shape=None,
         crpix=(0, 0),
```
mairanteodoro, line 1016:
```suggestion

```

mairanteodoro (changes requested):
I suggest we drop both `crpix` and `crval`.

romancal/resample/tests/test_resample.py
mcara: @mairanteodoro Sorry for the delay. Which tests are failing?

romancal/resample/tests/test_resample.py
mcara: I do not see which tests are failing due to this change. Yes, removing `crpix` is a different solution but it also will no longer test a specific branch in the code. I think it is important to get these tests working as is even though the tests themselves do not care how the WCS was "made".

schlafly:
Merging as this PR looks good, and the test failures are related to the ongoing unit-removal from romancal reference files.

PR #1574, opened 2025-01-08
Use mask when making segmentation image.

schlafly:
This PR adds some DQ bits to the mask used for computing the segmentation image.  Those bits were also used for the PSF fitting.  Because the two masks did not agree, it was possible for sources to be created consisting of fully masked pixels, which would cause problems when fitting PSFs to the fully masked regions.

This PR as currently constructed is a bit of a hack.
* the masks used in the segmentation image and the PSF fitting are not identical
* we should compute those masks in only one place
* I have added a hack to ignore the DEAD bit which is set for all pixels in the example files I am using.
so we should not merge this PR.  But I am hoping it is a useful starting point for discussions of the problem.


schlafly:
@tddesjardins , can you see if this addresses your issue?

schlafly:
@larrybradley , if you could take a look at my use of the mask image here and bless it, I'd appreciate it.

larrybradley:
@schlafly I've updated this PR:

* moved the psf model from lib to source_catalog
* creates a single mask to be used for both segmentation and PSF photometry

larrybradley (approved):
I approve :grin:

schlafly:
This looks good to me.  I'm running regtests now.  After I merge I'll update the other catalog PR to have the multiband catalog use the same mask infrastructure.  Thanks!
https://github.com/spacetelescope/RegressionTests/actions/runs/13267547607

schlafly:
@larrybradley , this has an issue where the MultibandSourceCatalog regression test seems to hit some NaN values and fail.
https://github.com/spacetelescope/RegressionTests/actions/runs/13267547607/job/37038824068#step:30:1383
Do you have time to take a look?  Thanks!

schlafly:
I will try to address this now.

schlafly:
Regression tests now pass except for changes in the WCS fit results, that look like tiny improvements.  I will merge and okify.

larrybradley:
Thanks, @schlafly!

PR #1578, opened 2025-01-10
Use PSF astrometry in tweakreg.

schlafly:
This PR fixes two issues with tweakreg's accuracy.  The first is that we at some point unintentionally switched from using the PSF astrometry (x_psf, y_psf) to the segment centroid astrometry (xcentroid, ycentroid) in tweakreg.  This leads to much worse astrometric accuracy.  The second is that I had left in a fudge factor in the tweakreg regression test, so that we did not notice when this occurred.  This PR switches back to using the PSF astrometry and removes the fudge factor.

Regression tests will fail due to changes in the astrometry.  Regression tests are here:
https://github.com/spacetelescope/RegressionTests/actions/runs/12713885744

Three failures in steps that use tweakreg.  Note that in these cases the median absolute errors decrease and the shifts become closer to zero arcseconds (when the simulated WCS is correct) or to 1" in dec (when the simulated WCS is intentionally offset by 1 arcsecond in ra / dec, in the tweakreg regression test).

We will need to okify these regression differences when merging.

## Tasks
  - [X] Regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/12713885744    - [ ] Do truth files need to be updated ("okified")?


ddavis-stsci (approved):
Looks reasonable to me

larrybradley (approved):

mairanteodoro (approved):
Looks good to me.
I haven't checked it yet, but don't we have to update the docs too?

schlafly:
Thanks, yes, I should update this line of the docs.
https://github.com/spacetelescope/romancal/blob/f538ba0734dc9dcbe73ef66768713a54d5457a7e/docs/roman/tweakreg/README.rst?plain=1#L23
I'll do that this evening and merge.

PR #1585, opened 2025-01-19
RCAL-930 Roundtrip L3 wcsinfo especially when skycell specifications are used

### LINKED ISSUE #1431: Verify that we can reproduce the L3 WCS from the L3 metadata
We should verify that we can reproduce the L3 WCS from the L3 metadata, when creating L3 files on sky cells and on inferred projections (i.e., no explicit sky cell, letting resample pick what WCS it wants to resample onto).

@mairanteodoro  found that this block of code
https://github.com/spacetelescope/romancal/blob/e741d1963e163d8fbbeff714cb6a2acea662ae8a/romancal/resample/resample.py#L820-L829
produces a different rotation matrix in the two blocks of the try statement, differing by the sign of rotation_matrix[0][0], which suggests to me that we may currently be encoding whether the RA direction increases to the +x or -x direction in the rotation matrix.  That hadn't previously been obvious to me but doesn't seem crazy.

The goal is that we should be able to feed the L3 metadata to something like
https://github.com/spacetelescope/romancal/blob/b652ef8ba98d0c6b4c135e66ee7eee21eb885308/romancal/pipeline/mosaic_pipeline.py#L154-L191
after appropriate translation of the metadata keywords into record keywords that that function expects, and get out a WCS that is a close match for the actual WCS.

stscieisenhamer:
Resolves [RCAL-930](https://jira.stsci.edu/browse/RCAL-930)

Closes #1431 

This PR addresses ensures that the L3 `meta.wcsinfo` can be used to accurately reproduce the originating WCS from which the wcsinfo block had been derived. In particular, when the L3 product has been created based on a skycell instead of the input images, this update ensures that roundtripping of the wcsinfo is still valid.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/changes/1585.mosaic_pipeline.rst
+++ b/changes/1585.mosaic_pipeline.rst
@@ -1,0 +1,1 @@
+Roundtrip L3 wcsinfo especially when skycell specifications are used
--- a/romancal/pipeline/mosaic_pipeline.py
+++ b/romancal/pipeline/mosaic_pipeline.py
@@ -14,6 +14,8 @@ from astropy import units as u
 from astropy.modeling import models
 from gwcs import WCS, coordinate_frames
 
+from stcal.alignment import util as wcs_util
+
 import romancal.datamodels.filetype as filetype
 from romancal.datamodels import ModelLibrary
 
@@ -114,13 +116,13 @@ class MosaicPipeline(RomanPipeline):
 
                     # check to see if there exists a skycell on disk if not create it
                     if not isfile(skycell_file_name):
-                        # extract the wcs info from the record for generate_tan_wcs
+                        # extract the wcs info from the record for skycell_to_wcs
                         log.info(
                             "Creating skycell image at ra: %f  dec %f",
                             float(skycell_record["ra_center"]),
                             float(skycell_record["dec_center"]),
                         )
-                        skycell_wcs = generate_tan_wcs(skycell_record)
+                        skycell_wcs = skycell_to_wcs(skycell_record)
                         # skycell_wcs.bounding_box = bounding_box
 
                         # For resample to use an external grid we need to pass it the skycell gwcs object
@@ -128,6 +130,7 @@ class MosaicPipeline(RomanPipeline):
                         wcs_tree = {"wcs": skycell_wcs}
                         wcs_file = asdf.AsdfFile(wcs_tree)
                         wcs_file.write_to("skycell_wcs.asdf")
+
                         self.resample.output_wcs = "skycell_wcs.asdf"
                         self.resample.output_shape = (
                             int(skycell_record["nx"]),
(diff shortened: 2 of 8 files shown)


stscieisenhamer:
[Regression test 1](https://github.com/spacetelescope/RegressionTests/actions/runs/12851081091)

schlafly (commented):
This looks good.  My chief concern is whether we're populating the "real" L3 WCS correctly, though, and I'm not 100% sure that this captures that case?  I would suggest one of the following two tests:
- regression test comparing wcsinfo_to_wcs of one of the resampled L3 files in the regression tests to l3.meta.wcs
- unit test comparing something like 
```
wcs1 = skycell_to_wcs(skycell_record)
l3 = dummy_l3()
gwcs_into_l3(l3, wcs1)
wcs2 = wcsinfo_to_wcs(l3.meta.wcsinfo)
# compare wcs1, wcs2
```

I'm also interested in knowing which of these two cases
https://github.com/spacetelescope/romancal/pull/1585/files#diff-15cbca0fd660d7d5eac167ece3633d2a1a665d776bf180b07617f06b5fb40a7bR229-R231
is the usual branch we take for normal files we produce; I hope that it's the one on line 230 and not the one on line 232-236?

Did you end up being able to reproduce the WCS nearly perfectly?

stscieisenhamer:
I have pushed the wcs comparison tests. The dec differences are < 1.e-5 arcsec. The ra differences are < 5e-4 arcsec.  If one looks at the wcs info itself, in particular the scaling, they are similar to the 6th decimal place. Without making the wcs to wcsinfo func very knowledgeable about the exact info in a wcs (right now it is fairly general), this is the best, due to calculation roundoff.

Concerning the question of whether the matrix is used directly or not. Methinks this is the wrong question to be asking. There is really only one situation where this routine would normally be called by MosaicPipeline and that is when a skycell has been defined. As best as I can tell, the patches file does not define a rotation matrix, but only "orientation". I have also not done a search, but I presume most skycells are not defined with 0 orientation anyways. Hence, it is mute.

Otherwise, the wcsinfo_to_wcs is just a convenience function. If MosaicPipeline does not have a skycell, it will be computing the wcs from the input images' wcs' anyways.

schlafly:
Thanks.  It looks to me like the mad_std are actually 10^-5 _mas_, rather than arcseconds, which is negligible.  The biggest difference in the regtest I looked at was the median overall offset rather than the standard deviation, coming from the scale factor times the distance from the center.  This was ~3e-3 mas, which is small but not utterly negligible.  If I artificially forced the scale factor to match exactly, this all went away and things agreed perfectly.

I think we have two options if we want to get that extra precision:
1. we can explicitly copy over factor_3 and factor_4 kind of as we do here 
https://github.com/spacetelescope/romancal/blob/main/romancal/resample/resample.py#L835-L836
That's fragile but will work for our usual skycells and so is maybe okay.
2. we can try changing https://github.com/spacetelescope/stcal/blob/main/src/stcal/alignment/util.py#L361 to have a parameter that lets one specify a larger than one pixel step, so that we get less numerical error there.

Do you have a sense here?  I'm okay with either of those options, but do think we should try to get the median shift down to the <1 uas range.

schlafly (approved):
Looks great, thanks!

stscieisenhamer:
For the record, here are some plots of the footprints of a few adjacent skycells and all skycells.

Screenshot 2025-01-28 at 10 12 31

Screenshot 2025-01-28 at 10 12 44

PR #1596, opened 2025-01-30
RCAL-965 Invoke converter from_tvac_raw to enable processing of TVAC/FPS data

stscieisenhamer:
Resolves [RCAL-965](https://jira.stsci.edu/browse/RCAL-965)

This PR addresses the issue that TVAC/FPS data cannot be run through the pipeline, mostly due to the fact that the TVAC-related datamodels are frozen. The method `ScienceRawModel.from_tvac_raw` is utilized to ensure that both science and TVAC/FPS data can be processed by the ELP.

The conversion happens in the DQ Init step at the same point that the RampModel consistency is done.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


stscieisenhamer:
[Initial regression run](https://github.com/spacetelescope/RegressionTests/actions/runs/13159285995)

braingram:
Would you add a regression test that runes the exposure pipeline with one of the tvac files?

romancal/regtest/test_elp_tvac.py
```diff
@@ -0,0 +1,166 @@
...
+    return run_elp.truth
+
+
+@pytest.mark.soctests
+def test_output_matches_truth(output_filename, truth_filename, ignore_asdf_paths):
+    diff = compare_asdf(output_filename, truth_filename, **ignore_asdf_paths)
+    assert diff.identical, diff.report()
+
+
+@pytest.mark.soctests
```
braingram, line 64: This looks like mostly a copy of [test_wfi_image_pipeline.py](https://github.com/spacetelescope/romancal/blob/826a972ebe8d01a6727921bfd248f029cb62255c/romancal/regtest/test_wfi_image_pipeline.py#L207). If the tests beyond this point in the file are not needed to satisfy the commented DMS requirements I'd say removing everything in this file after this commented line (64) makes sense.

If there is reason to keep the additional tests in this file at least `test_elp_input_dm` should be removed as it doesn't process tvac data and looks to be an exact duplicate of the one in `test_wfi_image_pipeline.py`.

stscieisenhamer: Though looking now, I would suggest keeping the tests but removing the dms requirement markings. Reason is that tvac/fps are taken in a different way, and having all the test to verify what is produced I actually like. Thoughts?

stscieisenhamer:
For all reviewers: I believe all the comments have been addressed. Ready for full review

braingram:
Please remove `test_elp_input_dm` it is an exact copy of https://github.com/spacetelescope/romancal/blob/826a972ebe8d01a6727921bfd248f029cb62255c/romancal/regtest/test_wfi_image_pipeline.py#L207

romancal/regtest/test_elp_tvac.py
stscieisenhamer: as noted below, my cutting missed it, even after calling it out specifically 😞

braingram:
Regtests with this PR show 1 error in one of the new tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/13316220160
@stscieisenhamer is this expected?

stscieisenhamer:
Hmmm. So, wrt that fact that the meta is fine, that is what is relevant to this pr.

With respect to the data differences, this runs fine on my mac so it may be a mac vs linux issue, though the differences seem large for that.

So, I'm going with expected and to okayify the results.

pyproject.toml
```diff
@@ -23,8 +23,8 @@ dependencies = [
     "photutils >=1.13.0",
     "pyparsing >=2.4.7",
     "requests >=2.26",
-    "roman_datamodels>=0.23.0,<0.24.0",
-    #"roman_datamodels @ git+https://github.com/spacetelescope/roman_datamodels.git",
+    #"roman_datamodels>=0.23.0,<0.24.0",
+    "roman_datamodels @ git+https://github.com/stscieisenhamer/roman_datamodels.git@rcal-965-tvac",
```
braingram, line 27: Note to update prior to merge.

braingram (approved):
LGTM (pending merge of https://github.com/spacetelescope/roman_datamodels/pull/455 and update to the roman_datamodels pin).

schlafly (commented):
I agree that the amount of duplication between test_wfi_image_pipeline and test_elp_tvac is unfortunate; if one does something horrible like from test_wfi_image_pipeline import a, b, c, d, e, f does that end up doing the right thing?  The tests do have value, as you say, but I don't think there's enough motivation to do a significant refactoring to accommodate these weird files and can go along with the copying if there's not an easy alternative.

As Brett says, the regtest is currently failing and so I might avoid merging for B17 unless this is expected.

braingram:
The CI is in a bad state unrelated to this PR (something is up with the webbpsf data and cache, hopefully https://github.com/spacetelescope/romancal/pull/1617 will work around the issue for now).

schlafly (approved):

PR #1611, opened 2025-02-11
Forced source catalog

schlafly:
Resolves [RCAL-995](https://jira.stsci.edu/browse/RCAL-995)

This PR adds a new forced-photometry mode to the source catalog step, enabling one to take a segmentation image and detection image from one epoch and apply it to derive photometry and shapes at another epoch.  It also makes some more fields part of the set of fields that the multiband source catalogs compute.

Note that this makes a fair number of changes to the contents of the source catalogs and so regression tests fail.  Tests pass locally using the updated make_regtestdata.sh included with this PR.  I will update the regression test data when we merge this.

## Tasks
- [X] if a JIRA ticket exists, [make sure it is resolved properly](https://github.com/spacetelescope/romancal/wiki/How-to-resolve-JIRA-issues)

### DIFF
--- a/changes/1574.source_catalog.rst
+++ b/changes/1574.source_catalog.rst
@@ -1,0 +1,1 @@
+Create a common mask array for both segmentation and PSF photometry.
--- a/changes/1585.mosaic_pipeline.rst
+++ b/changes/1585.mosaic_pipeline.rst
@@ -1,0 +1,1 @@
+Roundtrip L3 wcsinfo especially when skycell specifications are used
--- a/changes/1592.associations.rst
+++ b/changes/1592.associations.rst
@@ -1,0 +1,1 @@
+Fix bug where skycell_wcs_info was double json encoded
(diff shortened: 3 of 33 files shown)


schlafly:
@larrybradley , if you have a chance to look at this in the coming day or two, great.  This PR implements the forced photometry / multiple epoch source catalog along the lines we discussed.  The most significant conceptual change is that I changed the way we were dividing columns into "columns that get per-filter multiband photometry and not" into "columns that are computed from the direct images" and "columns that are computed from the detection image."  That led to me computing a few more quantities of marginal value on the multiband catalogs.

schlafly:
I'm running regression tests, but they will require some updates and so are not terribly informative.  An initial set is here:
https://github.com/spacetelescope/RegressionTests/actions/runs/13293265191/job/37119010359
showing that some catalog fields change name and that I hadn't had all of the needed inputs in place in artifactory for the new test when I ran this.  But @larrybradley , if you're able, you could take a quick look, though any significant changes should wait for the next build.

schlafly:
Regression tests are passing; merging this to fix the tests.
https://github.com/spacetelescope/RegressionTests/actions/runs/13298377857

PR #1623, opened 2025-02-14
release metadata for `0.18.0` (`25Q2_B17`)

zacharyburnett:


README.md
```diff
@@ -172,6 +172,7 @@ an [issue](https://github.com/spacetelescope/romancal/issues).
 | 0.16.2    | 24Q4_B15  | 063          | Aug 2024  | Release for Build 24Q3_B15 (Build 15) |
 | 0.16.3    | 24Q4_B15  | 063          | Aug 2024  | Release for Build 24Q3_B15 (Build 15) |
 | 0.17.0    | 25Q1_B16  | 065          | Nov 2024  | Release for Build 25Q1_B16 (Build 16) |
+| 0.18.0    | 25Q2_B17  | 072          | Feb 2025  | Release for Build 25Q2_B17 (Build 17) |
```
zacharyburnett, line 175: @schlafly to confirm, I should use CRDS pmap 0072 here, right?

braingram:
The latest stcal release doesn't support python 3.13:
https://github.com/spacetelescope/stcal/blob/fc7183a367466ed711fb1cc916a1b1fa54943569/pyproject.toml#L5
causing the check dependencies to fail. Releasing stcal would be one way to fix that.

README.md
zacharyburnett: It looks like this is the tested reference, so I'm gonna go ahead with this release

Issue #1633, opened 2025-02-20
s_region is off by 1/2 pixel


braingram:
Following the JWST convention the `meta.wcsinfo.s_region` is the spatial footprint using the bounding box of the wcs (so 4 corners at the edges of pixels including a point at -0.5, -0.5). However roman files are produced with an s_region that uses the center pixel coordinates (including a point 0, 0).

Opening a cal file `roman-pipeline/dev/WFI/image/r0000101001001001001_0001_wfi01_cal.asdf` yields:
```python
>> m.meta.wcs.bounding_box
ModelBoundingBox(
    intervals={
        x0: Interval(lower=-0.5, upper=4087.5)
        x1: Interval(lower=-0.5, upper=4087.5)
    }
    model=CompoundModel(inputs=('x0', 'x1'))
    order='F'
)
>> m.meta.wcsinfo['s_region']
'POLYGON ICRS  269.986902672 65.974283813 269.986696178 66.097247504 269.676738633 66.097214425 269.680070749 65.974360690'
>> m.meta.wcs(-0.5, -0.5)  # not in the s_region
(269.9869402588003, 65.97426889277803)
>> m.meta.wcs(0, 0)  # incorrectly in the s_region
(269.9869026716723, 65.97428381257568)
```
This is likely true (but I haven't confirmed) for:
- l2 files before and after tweakreg which use [create_footprint](https://github.com/spacetelescope/romancal/blob/6de6898df8f728e6dba00a5d39d23777683d3ef7/romancal/assign_wcs/utils.py#L361) with `center=True`
- l3 files for similar use of [create_footprint](https://github.com/spacetelescope/romancal/blob/6de6898df8f728e6dba00a5d39d23777683d3ef7/romancal/resample/resample.py#L798)

schlafly:
Thanks.  I agree that we should follow the Webb convention and go with the corners of the edge pixels.  I'm not terribly worried about the difference but in the interest of being able to share code we should make the same choices.

Issue #1630, opened 2025-02-18
Change default fill value for resampling to nan


braingram:
Currently the default fill value for resample and outlier detection is `indef`:
https://github.com/spacetelescope/romancal/blob/be5e0eac829171c55840b0041971f0fa73fb91c2/romancal/resample/resample_step.py#L60
https://github.com/spacetelescope/romancal/blob/be5e0eac829171c55840b0041971f0fa73fb91c2/romancal/outlier_detection/outlier_detection_step.py#L35

Since `indef` doesn't have a defined starting value and neither step starts from an existing array the use of `indef` seems misleading.

`nan` seems appropriate for outlier detection (since unsampled pixels will be assigned nans which won't contribue to the computed median).

Using `nan` for resample will mean that the unsampled pixels will be `nan` instead of 0.

JWST made this change in https://github.com/spacetelescope/jwst/pull/8488

braingram:
Pinging @mcara and @schlafly for comments.

mcara:
> Since `indef` doesn't have a defined starting value and neither step starts from an existing array the use of `indef` seems misleading.

That is correct. It should be `NaN` or `0` or whatever is convenient. Some prefer `Nan` because it is easier to create a mask with `np.isfinite` but the mask can also be created from the context array (if context is enabled).

mcara:
`INDEF` would make sense only if continuing resampling on an existing output image. That is how it was intended to work but I do not think it ever worked like that. See, e.g., https://github.com/spacetelescope/jwst/issues/7664#issuecomment-1894015768

mcara:
It is a relic from IRAF times.

schlafly:
NaN seems reasonable to me to follow Webb.  I'm pretty sure we handle both NaN and 0 cases in the source catalog step but we can update to do that if this change breaks that.  For some smoothing operations NaN is particularly frustrating but those steps can filter the NaNs out as appropriate; I think the most important thing would be to align these choices between Webb and Roman as much as possible.

nden:
@stscijgbot-rstdms

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RCAL-1022](https://jira.stsci.edu/browse/RCAL-1022)._

stscijgbot-rstdms:
[...]

Issue #1600, opened 2025-02-05
Resample output pixel area contains invalid value


braingram:
The truth file roman-pipeline/dev/truth/WFI/image/r0099101001001001001_r274dp63x31y81_prompt_F158_coadd.asdf contains a pixel area of -99999.0 even though the input pixel areas are `[2.80833899537275e-13, 2.80833899537275e-13, 2.80833899537275e-13]`.

This is likely due to resample never assigning to `meta.photometry.pixel_area` and instead only scaling it by `pixel_scale_ratio`.

nden:
@stscijgbot-rstdms

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RCAL-1005](https://jira.stsci.edu/browse/RCAL-1005)._

braingram:
I suggest we handle this after:
- https://github.com/spacetelescope/romancal/issues/1637 (so we can use the resample calculated pixel area)
- https://github.com/spacetelescope/romancal/issues/1598 (which is on hold until L3 metadata updates)

One reason is that this is L3 metadata that might be changed with the other L3 metadata updates.

braingram:
Should have closed with: https://github.com/spacetelescope/romancal/pull/1838

Issue #1598, opened 2025-02-04
populate_mosaic_basic overwritten in step


braingram:
The metadata blending performed by `populate_mosaic_basic`:
https://github.com/spacetelescope/romancal/blob/826a972ebe8d01a6727921bfd248f029cb62255c/romancal/resample/resample.py#L875
is later overwritten by:
- update_exposure_times: https://github.com/spacetelescope/romancal/blob/826a972ebe8d01a6727921bfd248f029cb62255c/romancal/resample/resample.py#L533
- l2_into_l3_meta: https://github.com/spacetelescope/romancal/blob/826a972ebe8d01a6727921bfd248f029cb62255c/romancal/resample/resample.py#L734

The test PR https://github.com/spacetelescope/romancal/pull/1597 adds a test that runs the resample step on 2 inputs with different metadata:
```python
                {
                    "meta.observation.visit": 1,
                    "meta.observation.pass": 1,
                    "meta.observation.segment": 1,
                    "meta.observation.program": 1,
                    "meta.instrument.optical_element": "F158",
                    "meta.instrument.name": "WFI",
                },
                {
                    "meta.observation.visit": 2,
                    "meta.observation.pass": 2,
                    "meta.observation.segment": 2,
                    "meta.observation.program": 2,
                    "meta.instrument.optical_element": "F062",
                    "meta.instrument.name": "WFI",
                },
            ),
```
The mosaic model contains:
```python
            {
                "visit": 1,
                "pass": 1,
                "segment": 1,
                "optical_element": "F158",
                "instrument": "WFI",
            },
```
This shows that the metadata assigned by `populate_mosaic_basic` is being overwritten since it would set:
- visit: -1
- pass: -1
- segment: -1

schlafly:
Thanks.  Looking at the code here---the update_exposure_times bit is redundant but at least the calculations look the same.  Meanwhile l2_into_l3_meta "contributes" only coordinates and so should probably get deleted with the "coordinates" contribution folded into populate_mosaic_basic; the other bits it does are inferior to what's happening in populate_mosaic_basic.

Anyway, I agree that overwriting without the blending is wrong and the redundant calculations are problematic---just making sure we agree on the scope of the bug here.  Thanks!

nden:
@stscijgbot-rstdms

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RCAL-1006](https://jira.stsci.edu/browse/RCAL-1006)._

[...]

PR #1634, opened 2025-02-20
RCAL-1014: Update resample to use stcal

### LINKED ISSUE #1610: resample should use common code from stcal
_Issue [RCAL-1014](https://jira.stsci.edu/browse/RCAL-1014) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

spacetelescope/stcal#320 adds common code from resample to stcal. romancal.resample should start using it.

braingram:
Resolves [RCAL-1014](https://jira.stsci.edu/browse/RCAL-1014)

Closes #1610 

This PR updates the resample step to use code from stcal.

Regression tests:
https://github.com/spacetelescope/RegressionTests/actions/runs/13398069422
Show differences for:
- err and var_* arrays. Differences are expected since stcal uses the current jwst method for resampling variance and error (whereas romancal main uses an older method)
- a lack of `exptype` and `group_id` in the output model. I think this is an improvement as these metadata attributes are properties of the level 2 files that leaked into the level 3 files

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


docs/0.5.0
braingram, line 1: While updating the docs I noticed this stray file added in https://github.com/spacetelescope/romancal/pull/596

romancal/resample/exptime_resampler.py
```diff
@@ -0,0 +1,74 @@
...
+                "data": model.data,
+                "dq": model.dq,
+            },
+            weight_type=None,
+            good_bits=self.good_bits,
+            flag_name_map=dqflags.pixel,
+        )
+
+        xmin, xmax, ymin, ymax = resample_range(data.shape, model.meta.wcs.bounding_box)
+        # FIXME reuse pixmap
```
braingram, line 44: Opened an issue on stcal: https://github.com/spacetelescope/stcal/issues/341

romancal/resample/meta_blender.py
```diff
@@ -0,0 +1,66 @@
...
+
+        self._model["individual_image_cal_logs"] = []
+
+        self._mid_mjds = []
+
+        self._meta.basic.time_first_mjd = model.meta.exposure.start_time.mjd
+        self._meta.basic.time_last_mjd = model.meta.exposure.end_time.mjd
+
+        # copy some metadata from the first input model
+        # FIXME this is incorrect and these values should be
```
braingram, line 21: This reproduces the current behavior on main. See https://github.com/spacetelescope/romancal/issues/1598

romancal/resample/meta_blender.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+            self._meta.basic.time_last_mjd = max(
+                self._meta.basic.time_last_mjd, model.meta.exposure.end_time.mjd
+            )
+
+        self._model["individual_image_cal_logs"].append(model.meta.cal_logs)
+
+        self._mid_mjds.append(model.meta.exposure.mid_time.mjd)
+
+        # FIXME this is an existing hack that we have to reproduce here
```
braingram, line 55: See: https://github.com/spacetelescope/roman_datamodels/issues/461

romancal/resample/resample.py
```diff
@@ -1,318 +1,220 @@
...
-        """
-        if (input_models is None) or (len(input_models) == 0):
-            raise ValueError(
-                "No input has been provided. Input must be a non-empty ModelLibrary"
-            )
-
+        # fillval indef doesn't define a starting value
+        # since we're not resampling onto an existing array
+        # we overwrite this to 0 (to match the old behavior)
+        if fillval.lower() == "indef":
```
braingram, line 40: See: https://github.com/spacetelescope/romancal/issues/1630

romancal/resample/resample.py
```diff
@@ -1,318 +1,220 @@
...
-
-        log.debug(f"Output mosaic size: {self.output_wcs.array_shape}")
-
-        # n_images sets the number of context image planes.
-        # This should be 1 to start (not the default of 2).
-        self.blank_output = maker_utils.mk_datamodel(
-            datamodels.MosaicModel, n_images=1, shape=tuple(self.output_wcs.array_shape)
+                    input_models.shelve(m, i, modify=False)
+
+            # FIXME use s_regions here, but this causes data differences
```
braingram, line 59: See https://github.com/spacetelescope/romancal/issues/1633

romancal/resample/resample.py
```diff
@@ -1,318 +1,220 @@
...
+        pixel_scale_ratio = self.output_model["pixel_scale_ratio"]
+        if pixel_scale_ratio is not None:
+            output_model.meta.resample.pixel_scale_ratio = pixel_scale_ratio
 
-        # cast context array to uint32
-        output_model.context = output_model.context.astype("uint32")
-        return output_model
+        # record the actual filenames (the expname from the association)
+        # for each file used to generate the output_model
+        # FIXME this is incorrect when resample_group is called with a subset of models
```
braingram, line 208: Needs romancal issue

romancal/resample/resample.py
```diff
@@ -1,318 +1,220 @@
...
+            "exposure_time": exposure_time,
+            "start_time": model.meta.exposure.start_time,
+            "end_time": model.meta.exposure.end_time,
+            "duration": None,  # unused
+            "level": level,
+            "subtracted": subtracted,
+        }
+
+    def _get_intensity_scale(self, model):
+        # FIXME we lie about this to retain the old behavior
```
braingram, line 169: Needs romancal issue

romancal/resample/tests/test_resample_step.py
```diff
@@ -370,3 +241,141 @@ def test_build_driz_weight_different_weight_type(base_image, weight_type):
...
+    )
+
+    for key, value in expected_basic.items():
+        assert getattr(output_model.meta.basic, key) == value
+
+
+@pytest.mark.parametrize(
+    "input_pixel_area, pixel_scale_ratio, expected_pixel_area",
+    [
+        # (None, 1.0, None), # this cannot be tested since it causes the step to crash
```
braingram, line 367: See https://github.com/spacetelescope/romancal/issues/1600

mcara (changes requested):
`build_driz_weight` is available in `stcal` - https://github.com/spacetelescope/stcal/blob/main/src/stcal/resample/utils.py#L80 Is this version different from the `stcal` version? Also, `decode_context` is available from `drizzle.utils`.

No methods in resample.py are documented.

I do not think this PR should be merged with `exptime_resampler.py`. Instead merge https://github.com/spacetelescope/stcal/pull/342 and move some of the code to `update_time`.

braingram:
> `build_driz_weight` is available in `stcal` - https://github.com/spacetelescope/stcal/blob/main/src/stcal/resample/utils.py#L80 Is this version different from the `stcal` version? Also, `decode_context` is available from `drizzle.utils`.

Thanks. I originally left them in to try and limit the scope of this PR to resample changes. I pushed a few new commits that removes both of these functions (and updates outlier detection to use `build_driz_weight` from stcal).

> No methods in resample.py are documented.

I pushed docstrings in https://github.com/spacetelescope/romancal/pull/1634/commits/cb8a1962242f27f98838a930f41ebccfdc8f3ccf
I don't believe anyone outside of romancal developers will ever use that class so the docs are aimed at that audience. It's expected that users will use `ResampleStep`.

> I do not think this PR should be merged with `exptime_resampler.py`. Instead merge [spacetelescope/stcal#342](https://github.com/spacetelescope/stcal/pull/342) and move some of the code to `update_time`.

If that's the case the stcal PR will need to be updated as it's not compatible with the needs here. I don't think stcal changes are needed for this PR. Yes we should eventually make changes to stcal to reuse the pixmap for exposure time resampling but as I noted on https://github.com/spacetelescope/stcal/pull/342 there are other stcal resample changes that will be needed to support https://github.com/spacetelescope/romancal/pull/1556. I believe a structure similar to what's in `exptime_resampler.py` may work but that can be handled in follow-up PRs.

mcara:
I can't see how `exptime_total` can be different from an all-zeros array: https://github.com/spacetelescope/romancal/blob/1aff84af0962a3e28a85589e7f588f0acbc44b17/romancal/resample/exptime_resampler.py#L67 It is just not using resampled image at all.

mcara:
@braingram I wanted to make a PR against your branch but obviously I did something wrong. Please rebase and drop two commits that came from my mistake: 9737677482fb3b340e3178bcedb14ed56efd8534 and be61fa20764836fdf2198004a64e81a736665caf

mcara:
Here is my suggestion on how to make resample time work with the `stcal` resample and current PR to `stcal` (which I modified sligthly).

[resample.py.txt](https://github.com/user-attachments/files/18935596/resample.py.txt)

Also remove romancal/resample/exptime_resampler.py

braingram:
> Here is my suggestion on how to make resample time work with the `stcal` resample and current PR to `stcal` (which I modified sligthly).
> 
> [resample.py.txt](https://github.com/user-attachments/files/18935596/resample.py.txt)
> 
> Also remove romancal/resample/exptime_resampler.py

I attempted to apply your updated code and tried to run https://github.com/spacetelescope/romancal/blob/main/romancal/regtest/test_resample.py
It's failing because the produced `meta.resample.product_exposure_time` is 0.0 (the expected value is 329.840087890625).

This PR as it stands produces the expected results without stcal changes. Why is this a problem?

mcara:
[resample.py.txt](https://github.com/user-attachments/files/18954667/resample.py.txt)

I have attached a new version that implements exactly the same method as in your code. I do see `<1e-6` relative differences but I am not sure if this is because of OS differences or is it something else.

> It's failing because the produced meta.resample.product_exposure_time is 0.0 (the expected value is 329.840087890625).

I can't see this happening unless regtests did not use the updated code from https://github.com/spacetelescope/stcal/pull/342.

braingram:
From offline discussion with @mcara and @nden the pixmap optimizations can be handled in follow-up PRs to stcal, romancal (and possibly jwst).

schlafly (commented):
Awesome.  Just to be clear, the _data_ are an exact match in your regression tests.  The uncertainty arrays are all different, presumably because of the change in the way stcal resample propagated the uncertainties from the L2s.

A little mysteriously to me, the new var_rnoises tend to be 5-10% smaller than the old ones, while the new var_poisson's tend to be 5-10% larger than the old ones.  I don't really have a picture for what's going on there.  The other big difference is that CRs and outliers have a much smaller impact on the variances than they used to.  
image

The new estimated uncertainties look clearly better in those cases; the data don't show as large of a variance as expected from the earlier variance estimates.

So this all looks good to me.

romancal/resample/resample.py
braingram: Yeah that's the plan.
- update the s_regions
- use the s_regions here
- remove `make_output_wcs` (and other code no longer needed)

braingram:
> Awesome. Just to be clear, the _data_ are an exact match in your regression tests.
Yeah the data arrays are unchanged with this PR (within the tolerance of the regtest comparisons).
 
> A little mysteriously to me, the new var_rnoises tend to be 5-10% smaller than the old ones, while the new var_poisson's tend to be 5-10% larger than the old ones. I don't really have a picture for what's going on there. The other big difference is that CRs and outliers have a much smaller impact on the variances than they used to.

One likely significant change is that stcal uses a non-unity weight map for resampling the variance arrays. This is not the case on romancal main:
https://github.com/spacetelescope/romancal/blob/9b860a868c66270499dd56bac1741494cb2b0ecb/romancal/resample/resample.py#L430
so if the outliers have a high var_rnoise they will be subdued with this PR (and not so on main). I'm not sure about the var_poisson difference though.

schlafly:
Ah, perfect.  Yes, using a different weight map for the data and the variance sounds bad; removing that will do something along the lines of what I'm seeing here.  Great.

schlafly (approved):

braingram:
Started a fresh regtest run for me to okify after merge: https://github.com/spacetelescope/RegressionTests/actions/runs/13573462160

PR #1642, opened 2025-02-25
RCAL-1016 move skycell wcs reading into resample

### LINKED ISSUE #1615: Use skycell WCS in outlier detection
_Issue [RCAL-1016](https://jira.stsci.edu/browse/RCAL-1016) was created on JIRA by [Eddie Schlafly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eschlafly):_

For Roman our usual mode for creating coadded products is to create them on a common set of predefined WCSes which we call skycells.  We support coadding onto skycells in resample but we do not currently support it in outlier detection.  This means that we make larger coadds during outlier detection than we need, and detect outliers in those larger coadds, and then throw away those computed values.  

We should instead use the skycell WCSes during the outlier detection step as well in order to save time and memory.  An argument could be made that we should in fact use a WCS that is one or two pixels larger than the skycell WCS so that we guarantee the exact same treatment of outliers in the boundary region between skycells.  If that's easy to support, we should do it, but if it's annoying, we should pass for now since it's certainly harder to think about.

### LINKED ISSUE #1643: MosaicPipeline fails with UnboundLocalError when provided a ModelLibrary
```python
from romancal.pipeline import MosaicPipeline
from romancal.datamodels import ModelLibrary

lib = ModelLibrary("L3_mosaic_asn.json")  # from regtest files
p = MosaicPipeline()
p(lib)
```
produces an exception:
```
File ~/.pyenv/versions/3.12.4/envs/romancal/lib/python3.12/site-packages/stpipe/step.py:585, in Step.__call__(self, *args)
    579 def __call__(self, *args):
    580     warnings.warn(
    581         "Step.__call__ is deprecated. It is equivalent to Step.run "
    582         "and is not recommended.",
    583         UserWarning,
    584     )
--> 585     return self.run(*args)

File ~/.pyenv/versions/3.12.4/envs/romancal/lib/python3.12/site-packages/stpipe/step.py:507, in Step.run(self, *args)
    505     self.prefetch(*args)
    506 try:
--> 507     step_result = self.process(*args)
    508 except TypeError as e:
    509     if "process() takes exactly" in str(e):

File ~/projects/src/romancal/romancal/pipeline/mosaic_pipeline.py:166, in MosaicPipeline.process(self, input)
    163         if input_filename:
    164             result.meta.filename = self.output_file
--> 166 return result

UnboundLocalError: cannot access local variable 'result' where it is not associated with a value
```
Due to the mosaic pipeline failing to handle all filetypes. Providing a `DataModel` causes the same error for the same reason.

braingram:
Resolves https://jira.stsci.edu/browse/RCAL-1016

Closes #1615
Closes #1643

This PR updates resample to read the skycell wcs from the association data associated with the passed ModelLibrary. This allows:
- outlier detection to use the skycell wcs
- the mosaic pipeline to be simplified

A new option is added `resample_on_skycell` for the resample, outlier detection and the mosaic pipeline. This defaults to `True` (where in all cases the skycell wcs from the input association will be used if available) but can be set to `False` to use a "combined" wcs that has a footprint large enough to encompass all footprints of input model wcses (the old behavior for outlier detection).

Regtests:
https://github.com/spacetelescope/RegressionTests/actions/runs/13523271813
show differences for `test_mos_skycell_pipeline` when comparing the result and truth files.

These differences only occur when outlier detection uses the skycell wcs. There are differences in the number of outliers detected (which can be explained by the different wcs which has a different pixel size). The different outliers result in data, variance, error and context differences in the resampled result.

Looking at some of the outlier differences there are some outliers on main that are missed with this PR and some that are detected with this PR that are missed on main.

To provide one example. There is a source near 269.8154487374499, 66.00822803716659. Looking at a 50x50 pixel region in the output shows some pixels that differ:
Screenshot 2025-03-04 at 1 30 05 PM
in part due to main detecting outliers in images 1 and 2 for this region and with this PR none are detected in 1 and 2. Looking at the data (stretched slighly) from image 1 I don't seen any obvious "outlier":
Screenshot 2025-03-04 at 1 38 53 PM
I am suspicious of the `maskpt` setting (on main, unchanged by this PR). It's causing many pixels to be masked in the median image. Here are the medians for that same region (for main and with this PR):
Screenshot 2025-03-04 at 1 42 43 PM
Many of the pixels are `nan` due to the maskpt value:
https://github.com/spacetelescope/romancal/blob/341d333b9e4d037caa09a4c65df9dd454ce395c4/romancal/outlier_detection/utils.py#L82C33-L82C47
(note that the main/combined wcs is not the same orientation as the pr/skycell wcs so look for the best matching rotation about the source).
Since many of these source pixels have weights near the threshold a small change to the threshold (as will happen with this PR) will change what pixels are non-nan (candidates for outlier detection).

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


schlafly:
Re maskpt, it looks like the code is deciding that the lowest X% of pixels are bad and is masking them out.  As soon as pixels start saturating then the ivm weights start dropping and roughly the brightest X% of pixels will also be the lowest X% of the weights.  So I'd speculate that that's what's going on here.  That doesn't explain the smattering of pixels in relatively not-bright regions or the smattering of pixels near the core that are not getting masked, however.

Even among pixels that aren't saturated, the ones with the most flux will have (marginally) the lowest weights because the more flux one has, the more heavily the first and last pixels get weighted, leading to higher read noise and lower total noise (since Poisson noise is dominant here).

I don't have a good suggestion there.  Broadly, the right thing to do is not to just mask things due to low weights, and instead to trust the weights except for things like DO_NOT_USE, etc..  I forget whether we store "partial saturation" flags, but another option would be to not mask low weighted pixels that are also partially saturated---but I think that would just push this problem to the brightest unsaturated pixels rather than the brightest partially saturated pixels.  I have implemented code once that did something similar to this, but it looked more like:
- compute the median weight
- throw away anything less than 1/100 that weight.
i.e., get a typical value and throw away anything that looks really suspiciously much worse than that.  That kind of scheme would probably work fine.

But I agree we shouldn't mess with this for this PR.

romancal/patch_match/patch_match.py
```diff
@@ -252,3 +258,131 @@ def veccoords_to_tangent_plane(vertices, tangent_point_vec):
...
+
+    if "skycell_wcs_info" in input.asn and input.asn["skycell_wcs_info"] != "none":
+        skycell_record = input.asn["skycell_wcs_info"]
+    else:
+        if PATCH_TABLE is None:
+            load_patch_table()
+        skycell_record = PATCH_TABLE[
+            np.where(PATCH_TABLE["name"][:] == skycell_name)[0][0]
+        ]
+    log.info("Skycell record %s:", skycell_record)
```
schlafly, line 380: I feel a little differently about the importance of "target" than implied here.  I think I would have checked first to see if skycell_wcs_info were present, and if so, returned skycell_to_wcs(skycell_record).  Only if it were not present would I look for the target as the skycell name, and then look that up in the patch file.

This is not actually important but would allow people to resample onto whatever fixed grids they wanted using the ELP.  i.e., if someone wants to make a WCS that were exactly centered on some nearby galaxy, they could do that by specifying an appropriate skycell_wcs_info in their association, and things would go, even if it weren't in our file.

That still kind of works now but they have to give it a name that we like or it will fail, and I'd rather they specify a totally different name than a name like our usual names if they're doing a custom "skycell."

schlafly (commented):
This looks great; mosaic pipeline becomes a single path!  One question about whether we should basically ignore "target" if skycell_wcs_info is set, otherwise looks good to me.

romancal/patch_match/patch_match.py
braingram: Sounds good. I'll look into making the code check skycell_wcs_info first (I don't expect it will change any of the tests). I'll update this thread once the PR is updated.

braingram: I updated this PR to read `skycell_wcs_info` first in https://github.com/spacetelescope/romancal/pull/1642/commits/cca45bb9efe2f00eb639c524b2701bda25ff69a7 (and also added a docstring).

I re-ran the regtests and the results look to be the same.

schlafly (approved):
Looks great, thanks!

PR #1668, opened 2025-03-28
Update filenames

schlafly:
This PR updates the regression test file names to use new names for the L1, L2, and L3 files.  L1 and L2 files are updated to include a bandpass component.  L3 files are updated to use the new format, along the lines of r00001_p_v01001001001001_r274dp63x31y81_f158_coadd.asdf.  These L3 names will need another update when the skycell names are updated, but at least this gets them closer to where they need to be.

I also updated the make_regtestdata.sh script to automatically pull and create files needed for the TVAC regression tests, so that this script remains a one-stop-shop for updating all regtest files, and we can just replace an existing regtest directory with one generated from this script.  That's a little circular in that it pulls the TVAC regtest data from artifactory, but that seems okay.

There is an issue with one regression test, test_wfi_dq_init, relating to how make_regtestdata.sh generates truth files for each script from a single run of the ELP, while the regression test directly runs dq init on an uncal file.  The ELP first wraps the datamodel in a ModelLibrary, which adds some keywords to data model that aren't present when the DQ initial step is run directly on a model.  I'm discussing with Brett how best to resolve that issue.

I still need to upload my local regtest files to a new artifactory directory and start a regtest against this.  We expect some failures from the usual numerical issues and the DQ init issue.

Regression tests are here:
https://github.com/spacetelescope/RegressionTests/actions/runs/14179657836
There are a number of failures due to numerical issues that I'll okify when we merge the PR; these are due to my running on a different architecture in creating the new regtest files.  No failures outside compare_asdf.

## Tasks


romancal/regtest/test_match_skycell.py
```diff
@@ -51,13 +51,13 @@ def test_match_files(run_patchlist, expected_match_files):
 
 
 @pytest.fixture(scope="module")
-def run_skycellasn(rtdata_module):
+def run_skycellasn(rtdata_module, run_patchlist):
```
schlafly, line 54: @ddavis-stsci , I somewhat blindly added this argument to run_skycellasn since locally I was getting cases where this failed because the match files didn't exist.  I decided that this was because something wasn't guaranteeing that run_patchlist would happen first, and adding this made this issue go away.  Does this change look sane to you?

romancal/regtest/test_mos_pipeline.py
```diff
@@ -22,7 +22,7 @@ def run_mos(rtdata_module, resource_tracker):
     rtdata.get_asn("WFI/image/L3_regtest_asn.json")
 
     # Test Pipeline
-    output = "r0099101001001001001_F158_visit_coadd.asdf"
+    output = "r0000101001001001001_f158_coadd.asdf"
```
schlafly, line 25: We don't expect to make this kind of product in operations; this is a non-skycell coadd.  So we don't have a defined naming scheme here.  I needed to change it to make F158 lowercase, and elected to drop the _visit_ since it's somewhat implied by the fact that the initial string only goes down to the visit level.

romancal/regtest/test_skycell_generation.py
```diff
@@ -13,12 +13,12 @@
...
-    "r512_p_v01001001001_r274dp63x32y82_f158_asn.json",
-    "r512_p_v01001001001_r274dp63x32y80_f158_asn.json",
-    "r512_p_v01001001001_r274dp63x32y81_f158_asn.json",
-    "r512_p_v01001001001_r274dp63x32y82_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x31y80_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x31y81_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x32y82_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x32y80_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x32y81_f158_asn.json",
+    "r00001_p_v01001001001_r274dp63x32y82_f158_asn.json",
```
schlafly, line 21: I moved the initial program string here to be r00001 to match the input exposures, which are program 00001.

ddavis-stsci (approved):
LGTM, Does this also mean that the level 1 files will have the filter as well?

braingram (approved):
Code changes and regtests look good to me.

schlafly:
Yes.  I updated the L1 files in romanisim here https://github.com/spacetelescope/romanisim/pull/210 , intending to follow the conventions here https://innerspace.stsci.edu/display/RDMSDOC/Roman+Science+File+Naming+Conventions+for+Files+in+the+Archive+Catalog#RomanScienceFileNamingConventionsforFilesintheArchiveCatalog-Level1 .

PR #1680, opened 2025-04-04
RCAL-1025: Write the WfiWcs models after ELP run

stscieisenhamer:
Resolves [RCAL-1025](https://jira.stsci.edu/browse/RCAL-1025)
Depends on:
- [RDM#477](https://github.com/spacetelescope/roman_datamodels/pull/477)
- [RAD#564](https://github.com/spacetelescope/rad/pull/564)

This PR addresses the [RSOCREQ-852](https://jira.stsci.edu/browse/RSOCREQ-852) requirement of producing a file that contains the GAIA-aligned WCS that is modified for the Level 1 detector. This will be provided to the community for those wishing to apply a fully-corrected WCS to their Level 1 products.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/pipeline/exposure_pipeline.py
```diff
@@ -145,6 +146,9 @@ def process(self, input):
         if not any_saturated:
             self.tweakreg.run(lib)
 
+        # Write out the WfiWcs products
+        self.save_wfiwcs(lib)
```
braingram, line 150: How about making this a step? That way users can skip this, control the output, etc using the existing strun infrastructure?

stscieisenhamer: I believe this is an official, archived, output product, much like the resulting models themselves. Hence, if saving output, these should be produced. @schlafly will need to be the final determination.

braingram: Another thing to note is that currently since this is not a step the romancal version won't be correctly recorded in the metadata (since the saving does not pass the model through [finalize_result](https://github.com/spacetelescope/romancal/blob/06a5c19cd9b24d68967497b206cc9f851b5e10e6/romancal/stpipe/core.py#L76)).

stscieisenhamer:
[Regression Test #1](https://github.com/spacetelescope/RegressionTests/actions/runs/14334781965)

romancal/pipeline/exposure_pipeline.py
stscieisenhamer: so, options are? 

1. call `finalize_result` explicitly
2. make a step
3. add to the return

Methinks the ideal is to add to the return list, but I do not believe that the stpipe internals have been updated for that, i.e., multiple return types.

stscieisenhamer: going with calling `finalize_result` explicitly.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -73,6 +75,9 @@ def process(self, input):
         self.source_catalog.return_updated_model = True
         # make sure we update source catalog coordinates afer running TweakRegStep
         self.tweakreg.update_source_catalog_coordinates = True
+        # tweakreg can also save the L1 wcs files. However, ELP takes
+        # on this responsibiility, so disable the tweakreg saving.
+        self.tweakreg.save_l1_wcs = False
```
schlafly, line 80:
```suggestion
        # tweakreg should save the L1 wcs files
        self.tweakreg.save_l1_wcs = True
```

We've had different conversations about this, but if we make this the default we'll do the right thing in operations.

schlafly (commented):
Minor changes requested to change the defaults.  Let's make Azure's job easy.  They need the WCS files created by tweakreg so that they get all 18 detectors in.  The current approach would mean that they have to change the imaging jobs to have save_l1_wcs = False in the ELP (since we don't want to make a bunch of files from the ELP run that doesn't include tweakreg).

For the spectroscopic files, once we have an ephemeris-updating step in some future build, we will probably want to use save_l1_wcs = True to get that updated ephemeris out.  But since we don't have that feature yet let's not worry about it.

schlafly:
CI failures are due to L1/L2 metadata changes for which fixes in romancal are coming in #1698 .  Merging.

PR #1688, opened 2025-04-08
remove uses of MosaicModel.append_individual_meta

braingram:
This PR updates the resample "meta_blender" to build tables out of L2 metadata and insert them in the L3 "individual_image_meta" list. This was previously handled by `MosaicModel.append_individual_meta` which:
- crashes when `cal_logs` are provided
- crashes when a None value is provided
- crashes if the L2 metadata form each input image is not structured in the same way ("outlier_detection" missing from 1 L2 input will result in an exception)

There are a few cases to consider (with decisions made in this PR) that are caused by constructing tables of L2 metadata. These issues would not be present if instead the L2 metadata was copied as-is into the L3 metadata (and tables could be created via MosaicModel methods with user input about these decisions).

- What is the appropriate datatype for a list of floats and a None? For example let's say 3 files are appended with backgrounds of `[1.1, None 0.1]`. This PR sets the `None` to `np.nan` in this case.
- What is the appropriate datatype for a list of ints and a None (`[1, None, 0]`). For this PR the `None` is converted to `np.nan` and the other values are converted to floats.
- What is the appropriate datatype for all `None`s? For this PR these are converted to strings.
- How should missing values be handled? This PR treats them (mostly) as `None` (see above). The one exception is that they are converted to a string of `MISSING_CELL` instead of `None`.

Regtests show no failures: https://github.com/spacetelescope/RegressionTests/actions/runs/14339854988/job/40509434891

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/resample/meta_blender.py
+++ b/romancal/resample/meta_blender.py
@@ -75,7 +75,7 @@ class TableBuilder:
 
 class MetaBlender:
     def __init__(self):
-        self._tables = defaultdict(lambda: TableBuilder())
+        self._tables = defaultdict(TableBuilder)
         self._n_rows = 0
 
     def _blend_first(self, model):
--- a/romancal/resample/meta_blender.py
+++ b/romancal/resample/meta_blender.py
@@ -26,7 +26,7 @@ class TableBuilder:
             self._columns[name] = [MISSING_CELL] * row_index
         return self._columns[name]
 
-    def _santize_value(self, value):
+    def _sanitize_value(self, value):
         if isinstance(value, list | dict | AsdfDictNode | AsdfListNode):
             return str(value)
         return value
--- a/romancal/resample/meta_blender.py
+++ b/romancal/resample/meta_blender.py
@@ -34,7 +34,7 @@ class TableBuilder:
     def add_row(self, data_dict, row_index):
         updated = set()
         for key, value in data_dict.items():
-            svalue = self._santize_value(value)
+            svalue = self._sanitize_value(value)
             if svalue is None:
                 self._flagged_columns.add(key)
             self._get_column(key, row_index).append(svalue)
--- a/romancal/resample/meta_blender.py
+++ b/romancal/resample/meta_blender.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 import numpy as np
 from asdf.lazy_nodes import AsdfDictNode, AsdfListNode
 from asdf.tags.core.ndarray import NDArrayType
-from astropy.table import QTable
+from astropy.table import Table
 from roman_datamodels import datamodels, maker_utils, stnode
 
 
@@ -70,7 +70,7 @@ class TableBuilder:
             # if all else fails, convert everything to a string
             array_columns[name] = [str(v) for v in values]
 
-        return QTable(array_columns)
+        return Table(array_columns)
 
 
 class MetaBlender:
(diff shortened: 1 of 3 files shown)


romancal/resample/meta_blender.py
```diff
@@ -1,8 +1,82 @@
...
+            if arr.dtype.kind in ("f", "i"):
+                array_columns[name] = [
+                    np.nan if v in (MISSING_CELL, None) else v for v in values
+                ]
+                continue
+
+            # if all else fails, convert everything to a string
+            array_columns[name] = [str(v) for v in values]
+
+        return QTable(array_columns)
```
braingram, line 72: `QTable` is used here to match what `append_individual_meta` is returning. I couldn't find any columns with units.

schlafly: Thanks.  We probably (?) had a bunch of units in there before the unit removal push a build or two ago, so these could probably go to Tables now.

romancal/resample/meta_blender.py
```diff
@@ -1,8 +1,82 @@
...
+        self._flagged_columns = set()
+
+    def _get_column(self, name, row_index):
+        if name not in self._columns:
+            if row_index:
+                self._flagged_columns.add(name)
+            self._columns[name] = [MISSING_CELL] * row_index
+        return self._columns[name]
+
+    def _santize_value(self, value):
```
schlafly, line 29:
```suggestion
    def _sanitize_value(self, value):
```
typo?

romancal/resample/meta_blender.py
```diff
@@ -1,8 +1,82 @@
...
+        if isinstance(value, list | dict | AsdfDictNode | AsdfListNode):
+            return str(value)
+        return value
+
+    def add_row(self, data_dict, row_index):
+        updated = set()
+        for key, value in data_dict.items():
+            svalue = self._santize_value(value)
```
schlafly, line 37:
```suggestion
            svalue = self._sanitize_value(value)
```

schlafly (commented):
Thanks Brett.  Do I understand correctly that a consequence of this PR is that we will start populating things like wcs_fit_results in the individual_image_meta?

I think that the choices you've made about handling Nones and different types are good ones.  I agree that they're annoying to make, though we naturally hope that they don't come up in operations (something has gone wrong if there's a None in one image and a value in another image).

The defaultdict(lambda: TableBuilder())-fu is beyond my ken, but the implementation doesn't look too mysterious.

This looks good to me if you're happy with it!

romancal/resample/meta_blender.py
braingram: I switched this to `Table` I expect there will be some regtest differences (rerunning them now).

braingram: I saw no differences. It looks like the [table comparison](https://github.com/spacetelescope/romancal/blob/002371b1f2d5400f8ae79a3fe110e8f1e59e581a/romancal/regtest/regtestdata.py#L608) doesn't distinguish between `Table` and `QTable`.

braingram:
> Thanks Brett. Do I understand correctly that a consequence of this PR is that we will start populating things like wcs_fit_results in the individual_image_meta?

With the current PR `wcs_fit_results` is ignored (same as with `append_individual_meta`). We could look into changing this (I originally had that change in this PR but it was coming through as a dict converted to a string not as a separate table as is done for the other `DNode`s in `meta`).
 
> The defaultdict(lambda: TableBuilder())-fu is beyond my ken, but the implementation doesn't look too mysterious.

Oh I think the lambda there isn't actually needed. I removed it in https://github.com/spacetelescope/romancal/pull/1688/commits/571a06bfe8554ba4bd5c901b0775a142f0ed99e7

schlafly (approved):
Thanks Brett, looks good.

PR #1694, opened 2025-04-09
[RCAL-1018] `patch_match` -> `skycell.match`

### LINKED ISSUE #1625: Modify patch_match.py code to handle new form of skycell reference file
_Issue [RCAL-1018](https://jira.stsci.edu/browse/RCAL-1018) was created on JIRA by [Perry Greenfield](https://jira.stsci.edu/secure/ViewProfile.jspa?name=perry):_

Now that the final form of the skycell reference file has been settled, it now is necessary to modify the patch_match.py file to handle the new organization of the reference file. Changes will include handling the two stage lookup process for overlapping skycells. Previously all skycells were checked to see if they were close to the exposure image. Now there is a table for all projection regions, and each projection region indicates which part of the skycell table it needs to check for matches. In most cases only one projection region must be used, but occasionally an exposure may overlap more than once projection region, so this involves more complex logic.

Furthermore, some field names have changed or have been moved to metadata. The corresponding references to the previous locations must be updated to use the new locations. Given the more complex projection region matching, more tests will be needed to check that exposures that may overlap more than one projection region are handled properly.

It also makes sense to change the name of the task, perhaps to skycell_match to align with the new terminology.

zacharyburnett:
Resolves [RCAL-1018](https://jira.stsci.edu/browse/RCAL-1018)

Closes #1625 

- rename `patch_match` -> ~`proj_match`~ `skycell.match` (what do you think of that name @spacetelescope/romancal-maintainers?)
- handle [new projection regions reference file format](https://roman-crds.stsci.edu/browse/roman_wfi_skycells_0003.asdf) (top-level projection region associated with "sky tiles" comprised of many individual skycells)
- moved skymap reference file loading and skycell enumeration to a new module `skycell.skymap`
  - ~~`SkyCell.wcs` outputs an `astropy.wcs.WCS` instead of a `gwcs.WCS`; is that okay?~~
  - set the `SKYMAP_PATH` environment variable to override the reference file loaded from CRDS in `skycell.skymap`

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


zacharyburnett:
the transition is pretty much complete (pending regtests); however, I could most likely improve the algorithm by
- find intersecting projection regions first (as they are larger), and then drill down into their associated skycells
- offload geometry operations onto `sphergeo`

zacharyburnett:
regtest running at https://github.com/spacetelescope/RegressionTests/actions/runs/14522099967

zacharyburnett:
reran regtests at https://github.com/spacetelescope/RegressionTests/actions/runs/14535803876

ddavis-stsci:
I think we need to look at the timing for this function. With the old patch match code with the reference file in my CRDS cache I get a time of 
file_patch_list = pm.find_patch_matches(cal.meta.wcs)
The difference of time is : 1.4472598750144243 seconds

with the same file and also with the reference file in my cache for skymatch I'm getting
file_patch_list = sm.find_skycell_matches(cal.meta.wcs)
The difference of time is : 476.64390887506306 seconds

or almost 8 minutes an increase of over 300x 
These use the python timeit function and an unloaded cpu.

zacharyburnett:
there are 8 million skycells so it makes sense that it would take a long time to compute intersections; I can optimize this by matching against projection region extents first (there are only 4096) and then matching against their skycells (it looks like ~2000 per region)

ddavis-stsci:
Another item, the script mk_skycell_list works as expected,
```
mk_skycell_list r0000101001001001001_0001_wfi01_cal.asdf
Command-line arguments: Namespace(output_dir='', filelist=['r0000101001001001001_0001_wfi01_cal.asdf'])
Skycell List:r0000101001001001001_0001_wfi01_cal.asdf, (array([1176849, 1176850, 1176909, 1176910, 1176965, 1176966]), np.int64(0))
```
but when trying to use this I get an error 
```
mk_skycell_asn_from_skycell_list r0000101001001001001_0001_wfi01_cal.match
Command-line arguments: Namespace(filelist=['r0000101001001001001_0001_wfi01_cal.match'], release_product='p', optical_element='f158')
Unique Skycells: [1176849 1176850 1176909 1176910 1176965 1176966]
  File "/Users/ddavis/miniconda3/envs/rcal_0.18_skycell/bin/mk_skycell_asn_from_skycell_list", line 8, in <module>
    sys.exit(_cli())
             ^^^^^^
  File "/Users/ddavis/miniconda3/envs/rcal_0.18_skycell/lib/python3.12/site-packages/romancal/associations/mk_skycell_asn_from_skycell_list.py", line 131, in _cli
    mk_skycell_asn_from_skycell_list(
  File "/Users/ddavis/miniconda3/envs/rcal_0.18_skycell/lib/python3.12/site-packages/romancal/associations/mk_skycell_asn_from_skycell_list.py", line 63, in mk_skycell_asn_from_skycell_list
    projcell_info = skycell.wcsinfo
                    ^^^^^^^^^^^^^^^
AttributeError: 'SkyCell' object has no attribute 'wcsinfo
```

I also see this with skycell_asn, maybe these are for the future?

zacharyburnett:
good catch! I removed that property, I'll do a more thorough sweep of usages

zacharyburnett:
> I think we need to look at the timing for this function. With the old patch match code with the reference file in my CRDS cache I get a time of file_patch_list = pm.find_patch_matches(cal.meta.wcs) The difference of time is : 1.4472598750144243 seconds
> 
> with the same file and also with the reference file in my cache for skymatch I'm getting file_patch_list = sm.find_skycell_matches(cal.meta.wcs) The difference of time is : 476.64390887506306 seconds
> 
> or almost 8 minutes an increase of over 300x These use the python timeit function and an unloaded cpu.
> 
> part of this may be that the length of the returned items, for len(file_patch_list[0]) this is 6 but for len(file_patch_list[1]) 503833 I know the skycells have changed but for the old code I'm getting len(file_patch_list[0]) 5 len(file_patch_list[1]) 101 so a much smaller second item

@ddavis-stsci I figured out the issue with the skytile matching (I had the wrong order of points 😅). It should be much faster now; I can do some more optimizations with a kdtree, if necessary

zacharyburnett:
regtest at https://github.com/spacetelescope/RegressionTests/actions/runs/14579205462

zacharyburnett:
running the `run_skycell_asn` fixture in `regtests/test_skycell_generation.py` (on `r0000101001001001001_0002_wfi01_f158_cal.asdf r0000101001001001001_0002_wfi10_f158_cal.asdf`) fails with 
```
E       romancal.associations.exceptions.AssociationNotValidError: Association r00001-a3001_image_010_asn with 1 products
E       Rule=DMS_ELPP_Base
E       No constraints
E       Products:
E       	r00001_p_v01001001001_270p65x48y69_f158 with 1 members is not valid
```

zacharyburnett:
the `skymap_subset.asdf` file is 7.1 MB; is that okay to put in the git tree?

schlafly:
It looks to me like the regtests haven't been updated such that they can pass; we'll need to get those into a state where they can pass before we can merge.  Are you having the same problems you had yesterday there?

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,588 @@
...
+            skycell_name = library.asn["skycell_wcs_info"]["name"]
+        elif "target" in library.asn:
+            # check to see if the product name contains a skycell name & if true get the skycell record
+            skycell_name = library.asn["target"]
+        else:
+            raise ValueError(
+                "cannot extract skycell information from modellibrary association with neither WCS nor target info"
+            )
+
+        return SkyCell(skycell_name)
```
nden, line 103: Should this be `SkyCell.from_name(skycell_name)`? At least that's what works for me when passing a sky cell name. Also it may be good to add a test.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,588 @@
...
+        return sgp.SingleSphericalPolygon(
+            points=sgv.normalize_vector(corner_vectorpoints),
+            inside=sgv.normalize_vector(center_vectorpoint),
+        )
+
+    @cached_property
+    def skytile(self) -> "SkyTile":
+        if self.index is None:
+            raise ValueError("no index provided")
+        return SkyTile.from_skycell_index(self.index)
```
nden, line 158: `SkyTile.from_skycell_index(0)` fails for me with 
`'sky cell index 0 not found in any projection regions' ` Am I running it incorrectly?

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,588 @@
...
+    @property
+    def pixel_scale(self) -> float:
+        return SKYMAP.pixel_scale
+
+    @property
+    def pixel_shape(self) -> tuple[int, int]:
+        return SKYMAP.pixel_shape
+
+    @property
+    def wcsinfo(self) -> dict:
```
nden, line 169: This fails because the `skytile` property for index 0 fails.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,588 @@
...
+        return self.data["roman"]["meta"]["pixel_scale"] / 3600.0
+
+    @property
+    def pixel_shape(self) -> tuple[int, int]:
+        return self.data["roman"]["meta"]["nxy_skycell"], self.data["roman"]["meta"][
+            "nxy_skycell"
+        ]
+
+    def __getitem__(self, index: int) -> SkyCell:
+        return SkyCell(self.data["roman"]["skycells"][index])
```
nden, line 585: Is this supposed to be `SkyCell.from_data(self.data["roman"]["skycells"][index])`?

nden (commented):
This is a partial review but I feel like I need to know the answers to the questions before I continue, to make sure I'm not misunderstanding the intent.

In general docstrings are missing from the code.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,523 @@
...
+    @cached_property
+    def wcs(self) -> WCS:
+        wcs = WCS(naxis=2)
+        wcs.wcs.cdelt = [self.pixel_scale, self.pixel_scale]
+        wcs.wcs.ctype = ["RA---TAN", "DEC--TAN"]
+        wcs.wcs.crval = list(self.skytile.radec_tangent)
+        wcs.wcs.crpix = list(self.xy_tangent)
+        wcs.wcs.crota = [0, self.skytile.orientation]  # CROTA2 is the rotation angle
+        wcs.array_shape = list(self.pixel_shape)
+        return wcs
```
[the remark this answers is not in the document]
zacharyburnett: I guess we don't need to have this here, it's not directly called anywhere (`.wcsinfo` instead)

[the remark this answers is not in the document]
nden: I don't think we should test with FITS WCS since we are not using FITS WCS.  Also `crota` is a long deprecated and FITS keyword and should not be used.

perrygreenfield (commented):
I am worried there is a major issue with the image footprint definition.

romancal/skycell/match.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(
+        self,
+        ra_min: float,
+        dec_min: float,
+        ra_max: float,
+        dec_max: float,
+    ):
+        """
+        :param ra_min: minimum right ascension
```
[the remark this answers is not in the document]
zacharyburnett: oh that's definitely a mistake, I must have been thinking in the image space. I'll fix that

braingram (commented):
I left a few comments about asdf usage, environment variables and the subset file.

Also any clue why the tests are failing for me locally? If I run `pytest romancal/skycell/tests/`

in a local checkout of ec596d45158b856d5a93d3f50e5860710bd74285
I see:
```
FAILED romancal/skycell/tests/test_skycell.py::test_skycell_to_wcs[225p90x30y51] - AssertionError:
FAILED romancal/skycell/tests/test_skycell.py::test_skycell_to_wcs[225p90x31y51] - AssertionError:
=========================================================================================================== 2 failed, 10 passed in 2.29s
```

<details><summary>click to expand</summary>

```
_________________________________________________________________________________________________________ test_skycell_to_wcs[225p90x30y51] _________________________________________________________________________________________________________

name = '225p90x30y51'

    @pytest.mark.parametrize(
        "name",
        ["225p90x30y51", "225p90x31y51"],
    )
    def test_skycell_to_wcs(name):
        """Test integrity of skycell.wcs"""

        skycell = SkyCell.from_name(name)

        wcs = skycell.wcs

>       assert_allclose(
            np.array(
                [
                    wcs(0.0, 0.0),
                    wcs(skycell.pixel_shape[0] - 1, 0.0),
                    wcs(skycell.pixel_shape[0] - 1, skycell.pixel_shape[1] - 1),
                    wcs(0.0, skycell.pixel_shape[1] - 1),
                ]
            ),
            skycell.radec_corners,
        )
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 6 / 8 (75%)
E       Max absolute difference among violations: 179.99970253
E       Max relative difference among violations: 0.57964156
E        ACTUAL: array([[133.662075,  88.495082],
E              [133.590571,  88.571384],
E              [130.536024,  88.567472],
E              [130.76173 ,  88.491369]])
E        DESIRED: array([[313.661777,  88.49509 ],
E              [313.590242,  88.571407],
E              [310.535039,  88.567493],
E              [310.760841,  88.491375]])

romancal/skycell/tests/test_skycell.py:90: AssertionError
_________________________________________________________________________________________________________ test_skycell_to_wcs[225p90x31y51] _________________________________________________________________________________________________________

name = '225p90x31y51'

    @pytest.mark.parametrize(
        "name",
        ["225p90x30y51", "225p90x31y51"],
    )
    def test_skycell_to_wcs(name):
        """Test integrity of skycell.wcs"""

        skycell = SkyCell.from_name(name)

        wcs = skycell.wcs

>       assert_allclose(
            np.array(
                [
                    wcs(0.0, 0.0),
                    wcs(skycell.pixel_shape[0] - 1, 0.0),
                    wcs(skycell.pixel_shape[0] - 1, skycell.pixel_shape[1] - 1),
                    wcs(0.0, skycell.pixel_shape[1] - 1),
                ]
            ),
            skycell.radec_corners,
        )
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 6 / 8 (75%)
E       Max absolute difference among violations: 179.99968694
E       Max relative difference among violations: 0.58009077
E        ACTUAL: array([[133.593563,  88.568346],
E              [133.514333,  88.644651],
E              [130.295514,  88.640528],
E              [130.545467,  88.564443]])
E        DESIRED: array([[313.59325 ,  88.568354],
E              [313.513985,  88.644673],
E              [310.294473,  88.640549],
E              [310.544531,  88.564449]])

romancal/skycell/tests/test_skycell.py:90: AssertionError```
</details>

File: romancal/skycell/tests/skymap_subset.asdf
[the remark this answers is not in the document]
Comment (zacharyburnett) [reply romancal/skycell/tests/skymap_subset.asdf:1]: as it is currently, the k-d tree works off of the indices of the skycells and projection regions as they're given in the array, so to accommodate the cells matched by the tests the subset needs to include every cell before that. To reduce the size I can find a test match closer to the start of the skycells list

**Comment by zacharyburnett** [comment]
> I left a few comments about asdf usage, environment variables and the subset file.
> 
> Also any clue why the tests are failing for me locally? If I run `pytest romancal/skycell/tests/`
> 
> in a local checkout of [ec596d4](https://github.com/spacetelescope/romancal/commit/ec596d45158b856d5a93d3f50e5860710bd74285) I see:
> 
> ```
> FAILED romancal/skycell/tests/test_skycell.py::test_skycell_to_wcs[225p90x30y51] - AssertionError:
> FAILED romancal/skycell/tests/test_skycell.py::test_skycell_to_wcs[225p90x31y51] - AssertionError:
> =========================================================================================================== 2 failed, 10 passed in 2.29s
> ```
> 
> click to expand

these tests fail because I'm not constructing the GWCS object correctly; they're 180° off for some reason, see https://github.com/spacetelescope/romancal/pull/1694#discussion_r2056828807

**Comment by perrygreenfield** [comment]
I'm wondering if the KD tree usage here is much of an advantage. I noticed the comment that it isn't done for all skycells due to the overhead of the construction of the tree, and that made me wonder if that negated any speed advantage even for the subset of skycells. One could construct the tree once and use it as a reference file (or put it in the existing reference file), but that drags in other issues. Then it adds another dependency and the need to learn its interface (I know that isn't an issue for you, but for potential maintainers). The existing partition of the problem into projection regions and corresponding subsets of skycells is in effect a two level approach rather than a deep tree. It significantly reduces the search space for the two iterations. In other words do you have any hard numbers on what speed up KD trees actually provide over the simple approach. And whether any such difference is significant in the more general usage cases; e.g., if the common use case is resampling data onto the skycells, is the savings in time a significant fraction of the time to do the resampling?

File: romancal/skycell/skymap.py [hunk 7]
```diff
@@ -0,0 +1,678 @@
...
+    if radec_coords.shape[0] == 2:
+        radec_coords = radec_coords.transpose()
+
+    # Convert all celestial coordinates to cartesian coordinates.
+    return np.stack(
+        sgv.lonlat_to_vector(radec_coords[:, 0], radec_coords[:, 1]), axis=1
+    )
+
+
+class SkyCell:
```
perrygreenfield, line 57: I'm wondering about the use of a class for this sort of thing. While in the abstract it looks reasonable, in practice it can inhibit performance when it is necessary to create many instances of the class or perform the same operations on many instances. In a general review of the code here it isn't entirely clear (yet anyway) whether this concern applies. I don't see a huge number of such instances created so I may be worrying about nothing. Even so, sometimes the class can be implemented with the arrays of information embedded in the class, and convenience methods to extract the desired attributes, either for a single skycell, or an array for a subset of skycells. If operations are needed on a large number, it is almost always faster to embed arrays in a class rather than have an effective array of instances. (This is generally true for object oriented languages.)

romancal/skycell/match.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Parameters
+        ----------
+        radec_corners: list[tuple[float, float]]
+            four corners of the image in right ascension and declination
+        """
+        radec_corners = np.array(radec_corners)
+        if radec_corners.shape[0] != 4:
+            raise ValueError(f"need 4 corners, not {radec_corners.shape[0]}")
+        self.__radec_corners = radec_corners
```
perrygreenfield, line 38: I have the impression (perhaps mistaken) that not many use the double underscore private attribute scheme in python, partly because it can be bypassed easily. The convention generally seems to be to use a single underscore to indicate it is private, and if you want to violate it by accessing it elsewhere, go ahead, but "you were warned"

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,672 @@
...
+        retrieve a sky cell from the sky map by its name
+
+        Parameters
+        ----------
+        name : str
+            name of a sky cell, for instance `315p86x50y75`
+        """
+        if not re.match(r"\d{3}\w\d{2}x\d{2}y\d{2}", name):
+            raise ValueError(f"invalid skycell name {name}")
+        return SkyCell(np.where(SKYMAP.skycells["name"] == name)[0][0])
```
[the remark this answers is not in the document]
zacharyburnett: good point, added a `KeyError` with a message in d1fbd39de040a3bf8b4587972e25f69e4682aa68 if the given skycell doesn't exist in the skymap. I can also add checking if the actual skycell location is valid, but I figured that is self-explanatory if it doesn't exist in the global skymap

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,678 @@
...
+        cls,
+        skytile_center: tuple[float, float],
+        skycell_coordinates: tuple[int, int],
+    ) -> "SkyCell":
+        """
+        retrieve a sky cell from the sky map using its location within its containins sky tile
+
+        Parameters
+        ----------
+        skytile_center : tuple[float, float]
```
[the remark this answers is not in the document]
zacharyburnett: this is referencing the sky cell name, which is constructed from the sky tile center. I can add a link to the handbook in the docstring

romancal/skycell/match.py
zacharyburnett: In this case, this attribute should never be modified, since it would misalign the cached properties relying on it. It doesn't make much sense to modify it after creating the object either, at least I can't think of a use case where that would be needed instead of just making a new object

romancal/skycell/skymap.py
zacharyburnett: I guess this depends on how large image mosaics can grow to be. I could definitely make a `SkyCells` that vectorizes operations

zacharyburnett:
> I'm wondering if the KD tree usage here is much of an advantage. I noticed the comment that it isn't done for all skycells due to the overhead of the construction of the tree, and that made me wonder if that negated any speed advantage even for the subset of skycells. One could construct the tree once and use it as a reference file (or put it in the existing reference file), but that drags in other issues. Then it adds another dependency and the need to learn its interface (I know that isn't an issue for you, but for potential maintainers). The existing partition of the problem into projection regions and corresponding subsets of skycells is in effect a two level approach rather than a deep tree. It significantly reduces the search space for the two iterations. In other words do you have any hard numbers on what speed up KD trees actually provide over the simple approach. And whether any such difference is significant in the more general usage cases; e.g., if the common use case is resampling data onto the skycells, is the savings in time a significant fraction of the time to do the resampling?

I did notice a significant speedup when matching against the entire skymap with brute force (even in the two-step scheme) vs constructing KD trees first. I haven't benchmarked it, but I think it's worth the build time, especially with the number of skycells per projection region. Unless the skymap structure completely changes, I don't anticipate we'll have to refactor that code again. It is using `scipy`, which is already in the environment

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,678 @@
...
+        return self.data[0]
+
+    @property
+    def radec_center(self) -> tuple[float, float]:
+        """center point in right ascension and declination"""
+        return self.data["ra_center"], self.data["dec_center"]
+
+    @property
+    def orientation(self) -> float:
+        """orientation from zenith"""
```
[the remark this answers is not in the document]
zacharyburnett: that's just an assumption from what I can tell of the data. I couldn't find an explanation of it in the schema; I'll take out the comment

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,672 @@
...
+                {
+                    "roman.meta.instrument.name": "WFI",
+                    "roman.meta.exposure.start_time": f"{datetime.now():%Y-%m-%d %H:%M:%S}",
+                },
+                reftypes=["skycells"],
+                observatory="roman",
+            )
+            self.__path = Path(rmap["skycells"])
+        with asdf.open(self.__path, memmap=True) as file:
+            output = file.copy()
```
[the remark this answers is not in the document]
zacharyburnett: resolved in 488d7c076c35c4db9f25dd20cc09f6011edafe81

romancal/skycell/tests/skymap_subset.asdf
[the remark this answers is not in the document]
zacharyburnett: was able to shrink the subset to 964K, though this still seems large

romancal/skycell/match.py
perrygreenfield: If the goal is to prevent modifying the attribute, the usual approach is to make it a property without a setter. The main difference between `self.__radec_corners` and `self._radec_corners` is that the latter indicates that the attribute is private and relies on convention not to use it, where as the former is a clumsy way of making inaccessible to code outside the class, though it is a very simple manner to access it by understanding how Python is mangling the name. My impression is few rely on this. Instead you can expose the value as a property but not provide a setter so no one can assign a new value to the property.

zacharyburnett: well I still don't see much functional difference between the two, but it's a trivial change so I'll update the classes

zacharyburnett: c07a61fc052ed039a2c34a3d4c097391b5a66085

perrygreenfield: It's not so much the functional difference than the fact when it is little used it may confuse other maintainers. It's basically a style consistency issue.

zacharyburnett:
here's a rough comparison of the parametrized `test_skycell_match` using the entire sky map table:
| test case | brute force time | k-d tree time |
| --- | --- | --- |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset6-45-0.001-expected_skycell_names6]` | `3.9s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset1-45-0.001-expected_skycell_names1]` | `3.9s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs_with_bbox[1]` | `5.2s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset5-45-0.001-expected_skycell_names5]` | `3.9s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset3-45-0.001-expected_skycell_names3]` | `3.9s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset4-45-0.001-expected_skycell_names4]` | `3.9s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset7-45-0.3-expected_skycell_names7]` | `4.0s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs[1]` | `5.2s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset0-45-0.001-expected_skycell_names0]` | `4.0s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs_with_imshape[1]` | `5.2s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset8-45-0.5-expected_skycell_names8]` | `5.3s` | `0.5s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset2-45-0.001-expected_skycell_names2]` | `4.0s` | `0.1s` |

romancal/skycell/match.py
```diff
@@ -0,0 +1,225 @@
...
+            inside=self.vectorpoint_center,
+        )
+
+    @cached_property
+    def area(self) -> float:
+        """area of this footprint on the sphere in degrees squared"""
+        return self.polygon.area()
+
+    @cached_property
+    def possibly_intersecting_projregions(self) -> int:
```
perrygreenfield, line 123: I wonder if this is unnecessary. All this machinery is hinged on skycells, which are never larger than a projection region.

romancal/skycell/match.py
```diff
@@ -0,0 +1,225 @@
...
+                np.ceil(self.polygon.area() / sc.SkyCell.area)
+                # plus multiplier for partial intersections on the perimeter
+                * 8
+            )
+        else:
+            # 4 foundational intersections
+            return 4
+
+
+def find_skycell_matches(
```
perrygreenfield, line 151: Is there a use case for this? I was under the impression that all data images will be resampled onto sky cells. This code suggests that arbitrary mosaics may be so resampled, but I though that had been ruled out.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,731 @@
...
+
+class ProjectionRegion:
+    """projection region in the sky map, corresponding to a single sky tile on the sky map"""
+
+    _index: int | None
+    _data: np.void
+
+    MIN_AREA = 0.002791388883915502
+    MAX_AREA = 0.003099159706138721
+    MIN_LENGTH = 0.031405005216643245
```
perrygreenfield, line 302: Is this used?

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,731 @@
...
+
+
+class ProjectionRegion:
+    """projection region in the sky map, corresponding to a single sky tile on the sky map"""
+
+    _index: int | None
+    _data: np.void
+
+    MIN_AREA = 0.002791388883915502
+    MAX_AREA = 0.003099159706138721
```
perrygreenfield, line 301: Is this used?

zacharyburnett:
I realize now that I was neglecting to implement Cartesian distance culling in the reimplementation of the old algorithm; that appears to be the better method: 
| test case | brute force time | k-d tree time | distance culling time |
| --- | --- | --- | --- |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset5-45-0.001-expected_skycell_names5]` | `3.9s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset8-45-0.5-expected_skycell_names8]` | `5.3s` | `0.5s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs_with_imshape[1]` | `5.2s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset2-45-0.001-expected_skycell_names2]` | `4.0s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset7-45-0.3-expected_skycell_names7]` | `4.0s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs_with_bbox[1]` | `5.2s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset6-45-0.001-expected_skycell_names6]` | `3.9s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset0-45-0.001-expected_skycell_names0]` | `4.0s` | `0.1s` | `0.4s` |
| `romancal.skycell.tests.test_skycell_match::test_match_from_wcs[1]` | `5.2s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset4-45-0.001-expected_skycell_names4]` | `3.9s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[1-offset3-45-0.001-expected_skycell_names3]` | `3.9s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell_match::test_skycell_match[0-offset1-45-0.001-expected_skycell_names1]` | `3.9s` | `0.1s` | `0.1s` |
| `romancal.skycell.tests.test_skycell::test_skycell_init` | `0.4s` | `0.4s` | `0.2s` |

zacharyburnett:
I realize now that I neglected to implement Cartesian distance culling in the reimplementation of the old algorithm; I'll explore running that over the regression tests as well

zacharyburnett:
it must be a rotation issue, I'll work on this more later tonight
file_vs_wcs

zacharyburnett:
aha! Found it (mostly)
file_vs_wcs

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,731 @@
...
+                (self.data["ra_corn2"], self.data["dec_corn2"]),
+                (self.data["ra_corn3"], self.data["dec_corn3"]),
+                (self.data["ra_corn4"], self.data["dec_corn4"]),
+            )
+        )
+
+    @cached_property
+    def vectorpoint_corners(self) -> NDArray[float]:
+        """corners in 3D Cartesian space on the unit sphere"""
+        return sgv.normalize_vector(
```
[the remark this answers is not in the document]
zacharyburnett: I don't think it does:
https://github.com/spacetelescope/spherical_geometry/blob/db71c0fe8ff4d4cbcdc48269325f69b4d0186db0/spherical_geometry/vector.py#L73-L76

zacharyburnett:
regtest running at https://github.com/spacetelescope/RegressionTests/actions/runs/14732583142/job/41365223423

zacharyburnett:
@nden @perrygreenfield @schlafly @ddavis-stsci this should be ready now, thanks for being so patient! 

The regtests have the following failures:
https://github.com/spacetelescope/RegressionTests/actions/runs/14732583142/attempts/2#summary-41365223423
```
>       raise AssociationNotValidError(f"Association {self} is not valid")
E       romancal.associations.exceptions.AssociationNotValidError: Association r00001-a3001_image_002_asn with 1 products
E       Rule=DMS_ELPP_Base
E       No constraints
E       Products:
E       	r00001_p_v01001001001_270p65x48y69_f158 with 1 members is not valid
```

and some slight differences in WCS parameters:
```
"root['roman']['meta']['wcsinfo']['orientat']": {'new_value': 359.84667934510236,
'old_value': 0.0},
"root['roman']['meta']['wcsinfo']['orientat_local']": {'new_value': 359.8467197346996,
'old_value': 355.9788004287368},
"root['roman']['meta']['wcsinfo']['pixel_scale']": {'new_value': 3.0363092873859018e-05,
'old_value': 2.777777819169892e-05},
"root['roman']['meta']['wcsinfo']['pixel_scale_local']": {'new_value': 3.0363092865617452e-05,
'old_value': 2.769508258922419e-05},
...
"root['roman']['meta']['wcsinfo']['rotation_matrix'][0][0]": {'new_value': -0.9999964196482088,
'old_value': 1.0},
"root['roman']['meta']['wcsinfo']['rotation_matrix'][0][1]": {'new_value': -0.002675946704136785,
'old_value': 0.0},
"root['roman']['meta']['wcsinfo']['rotation_matrix'][1][0]": {'new_value': -0.002675946704136785,
'old_value': -0.0},
"root['roman']['meta']['wcsinfo']['rotation_matrix'][1][1]": {'new_value': 0.9999964196482088,
'old_value': 1.0},
```
which I assume can be okified.

nden:
Association with one member should be a valid one and should run through HLPP. Please file an RCAL JIRA ticket for the first error.

schlafly:
Our regtests have one member associations, so this error is a bit weird to me.  I don't immediately see how changes to the patch match code can cause it.
https://github.com/spacetelescope/romancal/blob/b05555317d235f933079938596bbeab338a2b151/romancal/scripts/make_regtestdata.sh#L236

The case with different data
https://github.com/spacetelescope/RegressionTests/actions/runs/14732583142/job/41365223423#step:35:1971
is one that we don't want to okify.  I think what's going on here is that some code recognizes the skycell names and uses them to determine whether to coadd on skycells.  With your new module, it no longer recognizes the old names.  Rather than erroring out, it defaults to a "normal" automatic tangent plane, rather than the intended skycell.  That's why the new size isn't square.  The regression tests will need to be updated to use the new skycell names, e.g., by editing the target here:
https://github.com/spacetelescope/romancal/blob/b05555317d235f933079938596bbeab338a2b151/romancal/scripts/make_regtestdata.sh#L236
https://github.com/spacetelescope/romancal/blob/b05555317d235f933079938596bbeab338a2b151/romancal/scripts/make_regtestdata.sh#L206
with a new skycell name that touches the existing L1s.

braingram:
I think the association jsonschema needs to be updated for the new `skycell_wcs_info` structure:
https://github.com/spacetelescope/romancal/blob/b05555317d235f933079938596bbeab338a2b151/romancal/associations/lib/asn_schema_jw_level3.json#L33
Here's the `skycell_wcs_info` section for the `test_file_exists[r00001_p_v01001001001_r274dp63x31y80_f158_asn.json]`
```json
    "skycell_wcs_info": {
        "ra_ref": 270.0,
        "dec_ref": 64.6023730065,
        "x_ref": 12099.5,
        "y_ref": -88700.5,
        "rotation_matrix": null,
        "orientat": null,
        "pixel_scale": 1.5277777777777777e-05,
        "pixel_shape": [
            5000,
            5000
        ],
        "ra_center": 269.6396195128,
        "dec_center": 65.9950258509,
        "ra_corn1": 269.5464353538,
        "dec_corn1": 65.9566065856,
        "ra_corn2": 269.7338706068,
        "dec_corn2": 65.9570445961,
        "ra_corn3": 269.733081791,
        "dec_corn3": 66.0333875137,
        "ra_corn4": 269.545091011,
        "dec_corn4": 66.0329479585
    },
```
It's missing some required attributes.

zacharyburnett:
regtests running with patch names replaced by their closest skycell names:
https://github.com/spacetelescope/RegressionTests/actions/runs/14732583142

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,724 @@
...
+
+    Returns
+    -------
+    wcsobj : WCS
+        The WCS object from the skycell record.
+    """
+    wcsinfo = dict()
+
+    # The scale is given in arcseconds per pixel. Convert to degrees.
+    wcsinfo["pixel_scale"] = float(skycell_record["pixel_scale"]) / 3600.0
```
nden, line 548: This is not covered by tests and currently fails. It should be `skycell_record.pixel_scale`
`

nden: Similar access to the fields below.
It's best to add a test.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,724 @@
...
+    wcsobj = wcsinfo_to_wcs(wcsinfo, bounding_box=bounding_box)
+
+    wcsobj.array_shape = tuple(
+        int(axs[1] - axs[0] + 0.5)
+        for axs in wcsobj.bounding_box.bounding_box(order="C")
+    )
+    return wcsobj
+
+
+def to_skycell_wcs(library: ModelLibrary) -> WCS | None:
```
nden, line 572: Again, it's not necessary for a user to know how to construct a ModelLibrary when only the association is needed.
I suggest this is changed to accept an association (file)

nden: It also needs a test

nden (dismissed):

nden:
I requested some changes to fix a few failures and add tests. In general I think the code is in a good shape and I get reasonable results. Here's a plot of matching skycells for `r0000101001001001001_0001_wfi01_cal.asdf`.

association

zacharyburnett:
regtest run against `roman-pipeline/dev-20250501` on artifactory with renamed truth images: https://github.com/spacetelescope/RegressionTests/actions/runs/14776453611

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,724 @@
...
+
+
+class ProjectionRegion:
+    """projection region in the sky map"""
+
+    _index: int | None
+    _data: np.void
+
+    # area of the smallest projection region in square degrees on the sphere
+    MIN_AREA = 0.002791388883915502
```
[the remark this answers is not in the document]
zacharyburnett: good point about if the file changes; I just ran 
```python
min([ProjectionRegion(index).area for index in range(len(SKYMAP.projection_regions))])
```

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,724 @@
...
+    def from_data(cls, data: np.void) -> "ProjectionRegion":
+        """
+        build an index-less projection region instance from a data array
+
+        Parameters
+        ----------
+        data : numpy.void
+            array with projection region parameters (see schema)
+        """
+        instance = cls(index=None)
```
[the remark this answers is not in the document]
zacharyburnett: good catch!

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,724 @@
...
+        """
+        retrieve a sky cell from WCS info or a target specified in an association
+
+        Parameters
+        ----------
+        asn : asdf.AsdfTree | Path
+            association dictionary or a path to an association file to load
+        """
+        if isinstance(asn, Path):
+            asn = ModelLibrary._load_asn(asn).asn
```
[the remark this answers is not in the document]
zacharyburnett: added a test using some association files from `roman-pipeline/dev-20250501/WFI/image/`. Which file were you testing with?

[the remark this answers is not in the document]
nden: I generated the association file with the code in this PR.

zacharyburnett:
got the following error when generating the data files:
```
ERROR RUNNING STEP 'SourceCatalogStep':
    'nJy2' and 'nJy' (spectral flux density) are not convertible
```
Should the WCS parameters be `Quantities` to solve this issue?

zacharyburnett:
> > got the following error when generating the data files:
> > ```
> > ERROR RUNNING STEP 'SourceCatalogStep':
> >     'nJy2' and 'nJy' (spectral flux density) are not convertible
> > ```
> 
> Is this with your PR or on main too? If it's on main, let's ping Larry to have a look. I don't immediately see how your PR can be relevant here; the suggestion is that something is messed up with the source catalog flux units.

unfortunately not, it runs fine on `main`: https://github.com/spacetelescope/RegressionTests/actions/runs/14821392102/job/41608846319

Brett suggested just commenting the forced photometry section out of the regtest data gen script for now, I did that and am currently working on regtests against that data

zacharyburnett:
I'm do also get this warning in the mosaic pipeline:
```
2025-05-06 14:04:10,423 - stpipe.MosaicPipeline.outlier_detection - INFO - Resampling science and variance data
2025-05-06 14:04:11,885 - stpipe.MosaicPipeline.outlier_detection - INFO - Update S_REGION to POLYGON ICRS  269.725557125 66.030305373 269.724741347 66.106643367 269.913274070 66.106863917 269.913531101 66.030525147
2025-05-06 14:04:11,885 - stpipe.MosaicPipeline.outlier_detection - WARNING - WCS has no clear reference pixel defined by crpix1/crpix2. Assuming reference pixel is center.
2025-05-06 14:04:11,889 - stpipe.MosaicPipeline.outlier_detection - WARNING - WCS has no clear rotation matrix defined by pc_rotation_matrix. Calculating one.
```
so I reverted the model name changes for now

zacharyburnett:
made a ticket https://jira.stsci.edu/browse/RCAL-1072 to track source catalog masking issue

zacharyburnett:
made a ticket https://jira.stsci.edu/browse/AL-925 to track GWCS issue with inverse transform slightly outside of the bounding box (the failures in the unit tests)

zacharyburnett:
the regtests _should_ be using the full sky map reference file with 4058 projection regions, but instead they are using the subsetted one with only 3. Currently the subset is monkey-patched in with a module-scoped pytest fixture, so I'll have to figure out how to make that ONLY apply to the unit tests and not the regtests.
https://github.com/spacetelescope/RegressionTests/actions/runs/14811820471/job/41819200380#step:35:6222
```
IndexError: index 8 is out of bounds for axis 0 with size 3
```

zacharyburnett:
final regtest run against newly-generated data:
https://github.com/spacetelescope/RegressionTests/actions/runs/14910530917

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,633 @@
...
+            "ra_center": self.radec_center[0],
+            "dec_center": self.radec_center[1],
+            "nx": self.pixel_shape[0],
+            "ny": self.pixel_shape[1],
+            "orientat": self.orientation,
+            "orientat_projection_center": self.projection_region.orientation
+            if not self.projection_region.is_polar
+            # rotate polar cap by 180 degrees
+            # TODO: find out why this is necessary...
+            else self.projection_region.orientation + 180,
```
nden, line 281: I don't understand this. What is the issue you are seeing?

romancal/skycell/skymap.py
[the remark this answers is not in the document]
nden: Can you put this in a comment for the time being so it's clear where it comes from.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,633 @@
...
+
+    _index: int | None
+    _data: np.void
+    _skymap: "SkyMap"
+
+    # area of the smallest projection region in square degrees on the sphere
+    MIN_AREA = 0.002791388883915502
+
+    # diagonal length of the longest projection region in degrees on the sphere
+    MAX_LENGTH = 0.08174916691321586
```
nden, line 315: Also a comment on how this was computed.

romancal/skycell/skymap.py
zacharyburnett: if the projection region is a polar cap (the origin at the pole), the WCS transform is 180 degrees rotated in longitude

zacharyburnett: added in 7c39f884678f4d1594224089e0df06c22263b507

zacharyburnett:
regtest running against rebased branch: https://github.com/spacetelescope/RegressionTests/actions/runs/14910530917

nden (commented):
I'm OK with merging this and figuring out any potential accuracy issues, especially around the poles later, as I think it's important for this code to get exercised in SDP as early as possible and to give a chance to other teams to test it.

romancal/regtest/test_match_skycell.py
```diff
@@ -19,12 +19,8 @@
...
 EXPECTED_ASN_FILENAMES = [
-    "r00001_p_v0000101001001001001_r274dp63x31y80_f158_asn.json",
-    "r00001_p_v0000101001001001001_r274dp63x31y81_f158_asn.json",
-    "r00001_p_v0000101001001001001_r274dp63x31y82_f158_asn.json",
-    "r00001_p_v0000101001001001001_r274dp63x32y80_f158_asn.json",
-    "r00001_p_v0000101001001001001_r274dp63x32y81_f158_asn.json",
-    "r00001_p_v0000101001001001001_r274dp63x32y82_f158_asn.json",
+    "r00001_p_v0000101001001001001_270p65x49y70_f158_asn.json",
+    "r00001_p_v0000101001001001001_270p65x51y70_f158_asn.json",
 ]
```
schlafly, line 24: It's peculiar that we used to generate 6 files and now generate 2; the skycells got smaller rather than larger, so I expect to generate 4.  Do these two skycells files completely cover the L2 file?

romancal/regtest/test_skycell.py
```diff
@@ -0,0 +1,82 @@
...
+                        (skycell.pixel_shape[0] + 0.5, 0.5),
+                        (skycell.pixel_shape[0] + 0.5, skycell.pixel_shape[1] + 0.5),
+                        (0.5, skycell.pixel_shape[1] + 0.5),
+                    ]
+                ).T
+            )
+        ).T,
+        skycell.radec_corners,
+        rtol=1e-7,
+    )
```
schlafly, line 51: Do these corners look a little peculiar?  I think I would have guessed (-0.5, 0.5) to (shape - 0.5, shape - 0.5).  Unless this is fits vs. python, in which case we probably want a reference file update?

romancal/regtest/test_skycell_generation.py
```diff
@@ -13,12 +13,21 @@
...
+    "r00001_p_v01001001001_270p65x49y71_f158_asn.json",
+    "r00001_p_v01001001001_270p65x50y69_f158_asn.json",
+    "r00001_p_v01001001001_270p65x50y70_f158_asn.json",
+    "r00001_p_v01001001001_270p65x50y71_f158_asn.json",
+    "r00001_p_v01001001001_270p65x51y69_f158_asn.json",
+    "r00001_p_v01001001001_270p65x51y70_f158_asn.json",
+    "r00001_p_v01001001001_270p65x51y71_f158_asn.json",
+    "r00001_p_v01001001001_270p65x52y69_f158_asn.json",
+    "r00001_p_v01001001001_270p65x52y70_f158_asn.json",
+    "r00001_p_v01001001001_270p65x52y71_f158_asn.json",
```
schlafly, line 30: This is more like what I expect in terms of more new files than old files due to the smaller patch sizes.

romancal/skycell/match.py
```diff
@@ -0,0 +1,232 @@
...
+                elif hasattr(iwcs, "pixel_shape") and iwcs.pixel_shape is not None:
+                    image_shape = (iwcs.pixel_shape[1], iwcs.pixel_shape[0])
+                else:
+                    raise ValueError(
+                        "Cannot infer image footprint from GWCS object because "
+                        "`image_shape` was not specified and "
+                        "the GWCS object does not have `.bounding_box` nor `.pixel_shape`."
+                    )
+
+            iwcs.bounding_box = wcs_bbox_from_shape(image_shape)
```
schlafly, line 67: This roughly subtracts 0.5 from the shape.  Is that consistent with bbintervals adding 0.5?  I worry there is a fits / python convention difference in here.

romancal/skycell/match.py
schlafly: In principle we could define a projection region that way, but we never would.

romancal/skycell/match.py
```diff
@@ -0,0 +1,232 @@
...
+        if self.area > sc.SkyCell.area:
+            return (
+                # the number of times the smallest projection region could fit in the image footprint
+                np.ceil(self.area / sc.ProjectionRegion.MIN_AREA)
+                # plus multiplier for partial intersections on the perimeter
+                * 8
+            )
+        else:
+            # 4 foundational intersections
+            return 4
```
schlafly, line 134: I presume this is so that you can pull the right number of nearest neighbors.  I would suggest that you instead pull all points within some distance, where the distance is something like projection_region_size + image_size + buffer.  I can't think of a case where we need more than 4 points but I'm worried I'm not being creative enough thinking about the geometry.  Likewise for the skycells below.

romancal/skycell/match.py
schlafly: We're only planning on resampling images onto skycells, and not arbitrary mosaics, so I agree that this code is doing more than is needed, though that may be fine for now.

romancal/skycell/match.py
```diff
@@ -0,0 +1,232 @@
...
+        and does not contain a value for either the bounding_box or
+        pixel_shape attributes. Default value is None.
+    skymap: SkyMap
+        sky map instance (defaults to global SKYMAP)
+
+    Returns
+    -------
+    A sequence of the indices of all projection regions that overlap the supplied image
+    (in the referenced projection region table). The indices may be used to obtain all
+    necessary information about the projection regions.
```
schlafly, line 176: Is this description of the return value right?  It looks to me like it's the skycell indices rather than the projection regions, and it's a tuple with two elements, the first of which I think we want, the second of which it's less clear to me is what we want?

romancal/skycell/plot.py
```diff
@@ -0,0 +1,208 @@
...
+        - np.array(tangent_vectorpoint)
+        * np.dot(np.array([0, 0, 1]), np.array(tangent_vectorpoint))
+    )
+    avertices = np.vstack(vertices).T
+    x_coords = np.dot(x_axis, avertices) * RAD_TO_ARCSEC
+    y_coords = np.dot(y_axis, avertices) * RAD_TO_ARCSEC
+    return np.stack([x_coords, y_coords], axis=1)
+
+
+# plt.ion()
```
schlafly, line 70: Probably want to remove.

romancal/skycell/tests/test_skycell.py
```diff
@@ -0,0 +1,280 @@
...
+
+    with pytest.raises(ValueError):
+        sc.SkyCell.from_asn(
+            DATA_DIRECTORY / "L3_regtest_asn.json", skymap=skymap_subset
+        )
+
+    with pytest.raises(ValueError):
+        sc.SkyCell.from_asn(
+            DATA_DIRECTORY / "L3_skycell_mbcat_asn.json", skymap=skymap_subset
+        )
```
schlafly, line 93: These tests are checking to make sure that the skymap_subset is incomplete, basically?  I have more sympathy for the earlier tests showing that "notaname" wasn't in the subset.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,635 @@
...
+        skymap: SkyMap
+            sky map instance (defaults to global SKYMAP)
+
+        References
+        ----------
+        .. [1] `Skymap Tessellation <https://roman-docs.stsci.edu/data-handbook-home/wfi-data-format/skymap-tessellation>`_
+        """
+        return cls.from_name(
+            f"r{round(projregion_center[0]):03}d{'p' if projregion_center[1] >= 0 else 'm'}{round(projregion_center[1]):02}x{'p' if skycell_coordinates[1] >= 0 else 'm'}{skycell_coordinates[1]:02}y{'p' if skycell_coordinates[1] >= 0 else 'm'}{skycell_coordinates[1]:02}",
+            skymap=skymap,
```
schlafly, line 133: This looks like it uses the old style names rather than the new style names?  (i.e., it has names like r270dp66x01y01 rather than 270p66x01y01).  Is that intentional?  Do you also need to do some absolute value-ing on projregion_center[1] and skycell_coordinates[1], so that you get something like m01 rather than m-1 or something?  I think also the 'x' bit of the skycell_coordinates needs to reference skycell_coordinates[0] rather than skycell_coordinates[1].  

But maybe this function just isn't used?  It looks broken to me.

romancal/skycell/skymap.py
```diff
@@ -0,0 +1,635 @@
...
+        }
+
+    @cached_property
+    def wcs(self) -> WCS:
+        """WCS representing this sky cell"""
+        wcsobj = wcsinfo_to_wcs(
+            self.wcs_info,
+            bounding_box=(
+                (-0.5, self.pixel_shape[0] + 0.5),
+                (-0.5, self.pixel_shape[1] + 0.5),
```
schlafly, line 291: I always screw up the conventions here.  Is this right?  I expect this to roughly be the very top of the image to the very bottom of the image.  I think that should contain pixel.shape pixels; i.e., I expect bounding_box[0][1] - bounding_box[0][0] = pixel.shape[0], but here it's pixel.shape[0] + 1.  So a 1x1 pixel image will have a bounding box covering a 2x2 pixel region??

romancal/skycell/match.py
schlafly: Coming back to this later, I think my suggestion is roughly replace query(x, nneighbor) with query(x, 99, distance_upper_bound=size_of_projection_region + size_of_image + buffer).

schlafly (commented):
Bunch of minor comments inline, but basically this looks good to me.  
Follow up items:
- There are some things with the bounding boxes that I don't understand that look like fits vs. pythonisms (off by one errors, is the first pixel at 0 0 or 1 1).  It's worth understandnig these.
- We need to follow up with Dario re the orientation at the celestial poles, though that can happen outside this PR.
- There's at least one from_name function that looks pretty broken to me that I don't understand.

docs/roman/associations/skycell_asn.rst
```diff
@@ -92,16 +69,16 @@ contribute data to. The association files will be json files with names based
...
 
-	r00991_<release product name>_v<visit_id>_<skycell name>_<filter>_coadd_asn.json
+	r00991_<release product name>_v<visit_id>_<skycell name>_<filter>_asn.json
 
 or for the selections above
 
 .. code-block:: text
 
-	r00991_p_v101001001003001_r257dp63x98y83_F158_coadd_asn.json
+	r00991_p_v101001001003001_r257dp63x98y83_F158_asn.json
```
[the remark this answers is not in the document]
ddavis-stsci: This will be fixed in rcal-1076

romancal/associations/mk_skycell_asn_from_skycell_list.py
```diff
@@ -37,57 +36,41 @@ def mk_skycell_asn_from_patchlist(filelist, release_product, optical_element):
...
-            + release_product
-            + "_"
-            + "v"
-            + member_list[0][1:20]
-            + "_"
-            + patch_name
-            + "_"
-            + optical_element
-        )
+        asn_file_name = f"r{program_id}_{release_product}_v{member_list[0][1:20]}_{skycell.name}_{optical_element}"
```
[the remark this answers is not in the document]
ddavis-stsci: The 1:20 is to get the visit_id from the input string.

romancal/skycell/match.py
zacharyburnett: ah, I see now I did not fully understand @nden's point before. This shouldn't even be trying to derive the image shape from the bounding box here, since it's inside `iwcs.bounding_box is None`. I'll take out that `if` block

romancal/regtest/test_match_skycell.py
zacharyburnett: this appears to be working as expected; it's finding the unique skycells between the two matches: https://github.com/spacetelescope/romancal/blob/7592a904df89b99400718705c8746df0dadb9415/romancal/associations/mk_skycell_asn_from_skycell_list.py#L52-L54

Since the skycells are smaller, I would expect the same two footprints would have less overlap.

zacharyburnett:
https://github.com/spacetelescope/RegressionTests/actions/runs/14910530917/attempts/4#summary-42055998379

# regression test results

## differences from previous run

| test case | previous status | current status | previous peakmem | current peakmem |
| --- | --- | --- | --- | --- |
| `romancal.regtest.test_mos_skycell_pipeline::test_log_tracked_resources` |  |  | `4431MB` | `5021MB` |
| `romancal.regtest.test_multiband_catalog::test_multiband_catalog` |  |  | `3261MB` | `6398MB` |
| `romancal.regtest.test_source_catalog::test_log_tracked_resources[L2]` |  |  | `1125MB` | `1119MB` |
| `romancal.regtest.test_source_catalog::test_log_tracked_resources[L3skycell]` |  |  | `1670MB` | `2056MB` |

## test results
889 passed, 8 skipped

romancal/associations/mk_skycell_list.py
```diff
@@ -1,59 +1,55 @@
...
         if not output_dir:
             output_file_name = os.path.join(input_dir, output_file_name)
         else:
             output_file_name = os.path.join(output_dir, output_file_name)
 
         with open(output_file_name + ".match", "w") as outfile:
-            out_string = (
-                file_name + " " + np.array2string(file_patch_list[0], separator=",")
-            )
-            outfile.write(out_string)
```
schlafly, line 52: It looks like mk_patchlist / mk_skycell_asn_from_skycell_list rely on the precise formatting this np.array2string(...) gives.  With the changes this string gets more spaces, which are then parsed incorrectly around line 43 of mk_skycell_asn_from_skycell_list.  This then leads to the smaller number of output skycells.  One  could imagine switching to some less bespoke format, but probably let's just retain the original behavior for this PR?

romancal/regtest/test_match_skycell.py
zacharyburnett: this should be resolved by f4bfc3b2f5f844136f01c1ea336a994cb2360515 now

zacharyburnett:
regtest run: https://github.com/spacetelescope/RegressionTests/actions/runs/14910530917/job/42078459586

docs/roman/associations/skycell_asn.rst
```diff
@@ -92,16 +69,16 @@ contribute data to. The association files will be json files with names based
 
 .. code-block:: text
 
-	r00991_<release product name>_v<visit_id>_<skycell name>_<filter>_coadd_asn.json
+	r00991_<release product name>_v<visit_id>_<skycell name>_<filter>_asn.json
 
 or for the selections above
 
 .. code-block:: text
```
schlafly, line 77: Can we update this file name example now to be one of the new style names?  Just replace r257dp63x98y83 with your favorite skycell name.

schlafly: 270p65x48y69

PR #1698, opened 2025-04-11
Update romancal to accommodate L1 / L2 schema changes for B18

schlafly:
This PR makes romancal changes needed to accommodate the B18 metadata changes in https://github.com/spacetelescope/rad/pull/574.  The biggest code consequences are:
* guide_window.xsize / ysize went away.  SDP will be populating xstop / ystop instead.  So there were changes to move from using start + size to start + stop.
* I fixed one bug where the x & y indices for marking the guide window affected areas were wrong.  This doesn't really matter since we aren't populating real guide star locations.
* I added code to mark the guide window itself as DO_NOT_USE following instructions from Tyler.
* exposure.mid_time went away.  We were using this for astrometric purposes.  I switched this to use start_time instead.  This isn't quite as good, but a few minutes don't make a meaningful difference for astrometry.
* The L3 schema still has a mid_time, so I now compute it for the L3 images as start time + exposure time.
Regression tests of course need significant updates since this depends on the L1 / L2 metadata changes.

Regression tests here show only okifyable errors associated with numerical differences between my machine and the runners.  https://github.com/spacetelescope/RegressionTests/actions/runs/14416370812/job/40433398623


schlafly:
Merging to fix regression tests.

PR #1714, opened 2025-04-23
Update catalog regtest

schlafly:
This PR updates the names of some of the fields used for the regression tests of the catalogs to match their new names after the schema updates.


schlafly:
This fixes the formerly broken regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/14622742698 , merging.

PR #1751, opened 2025-05-09
Include visit in skycell_asn filenames and update default name.

schlafly:
This PR adds the VISIT component to the file names of prompt visit association tables producted by skycell_asn.  It also updates the default root product name to be the usual r<PROGRAM> that begins each file name.

With this PR, skycell_asn r*_f158_cal.asdf for our regression test files creates the following files:
r00001_p_v01001001001001_r274dp63x33y81_f158_asn.json
...
while without this PR the file names are
r00001_p_v01001001001_r274dp63x33y81_f158_asn.json

We want the "001" to fully specify the visit and match the L3 cal file names.

Regtest running here: https://github.com/spacetelescope/RegressionTests/actions/runs/14930782626

## Tasks

### DIFF
--- a/changes/1751.associations.rst
+++ b/changes/1751.associations.rst
@@ -1,0 +1,3 @@
+Adds the visit ID to skycell_asn association file outputs.
+
+
--- a/changes/1751.associations.rst
+++ b/changes/1751.associations.rst
@@ -1,3 +1,1 @@
 Adds the visit ID to skycell_asn association file outputs.
-
-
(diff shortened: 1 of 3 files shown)


ddavis-stsci (approved):
This looks fine. 
Hmmm... I thought that the asn naming was now a function in romancal/associations/lib/utilities.py but I don't see it. Since the code for mk_skycell_asn_from_patchlist.py needs this same update maybe this is a good time to move this to the utilities module?

schlafly:
I agree that that would be a good design.  Do you want to take over this PR and do that, or should we merge this one and spin that off as a separate PR?

ddavis-stsci:
Up to you. I can do it. Do we want this for the upcoming release?

schlafly:
I think it has some value for the upcoming release, at least for skycell_asn, since they're starting to use it to make the first L3 files and I want those files to have the right names (which is how I noticed it).  Please go ahead, thanks!

Issue #1765, opened 2025-05-19
Handle mosaic images with few contributing pixels


schlafly:
Jennifer Melot reports that the source catalog step fails in a handful of cases where we produce a skycell image with no valid pixels.  She's provided example data here:
/grp/roman/ddavis/roman/SDF/bad_boxes
and an example problem invocation would be:
strun roman_mos r00171_p_v01001001001004_270p65x56y67_f062_asn.json

The issue is that the input L2 file only barely overlaps the output skycell.  It looks to me like there "should" be 138 valid pixels in the overlap, but actually resample produces an image that is exactly zero everywhere.  I haven't figured out why those 138 pixels don't contribute to the output skycell---it looks to me like they land on the skycell and have non-zero weight (no issue with good_bits).  Does anyone understand that better than I do?

Regardless, the final "empty" image then goes to source_catalog, where the background subtraction barfs because there are no valid pixels.

I think we want:
- source_catalog should be robust against this; in this case, for example, it would be appropriate for the background subtraction to be skipped, or the background image to be all zero. @larrybradley 
- potentially, we probably do not actually want to produce coadds and catalogs when there are no valid pixels.  We can just log this and exit.  It's awkward to produce an association file and then not do anything with it, but it does seem like we'll have cases where only a corner overlaps, and that corner is masked and unusable for some reason.  i.e., generically I don't think we can avoid this case.
- @zacharyburnett , I note I expected the issue here to be that skycell.match was making some approximation and the image didn't actually overlap the skycell.  I think I understand the code ~here https://github.com/spacetelescope/romancal/blob/main/romancal/skycell/skymap.py#L441 to assume that using only the 4 corners of a skycell is adequate to determine whether a skycell overlaps.  I think we should likely define a bunch of additional points (these only get called on the handful of skycells that come through the kdtree, so this isn't expensive?), and probably also move the pixels out 1 pixel to give a little buffer.  That would cause more issues like this one, but I think it's better to try to resample and have it turn out that there are no pixels, than it is to miss an input pixel of an image.

Do others see this differently than I do?

@stscijgbot-rstdms

larrybradley:
For some reason, @stscijgbot-rstdms bot didn't create a corresponding JIRA issue.

schlafly:
[...]

PR #1767, opened 2025-05-20
Handle invalid input to the SourceCatalogStep

### LINKED ISSUE #1768: Handle invalid data input to SourceCatalogStep
_Issue [RCAL-1082](https://jira.stsci.edu/browse/RCAL-1082) was created on JIRA by [Larry Bradley](https://jira.stsci.edu/secure/ViewProfile.jspa?name=lbradley):_

larrybradley:
Resolves [RCAL-1082](https://jira.stsci.edu/browse/RCAL-1082)

Partially addresses #1765
Closes #1768 

This PR updates the SourceCatalogStep to handle the case where there are no valid pixels in the input image (e.g., all non-finite values or all pixels are masked).  For the issue reported in #1765, the coadd model input to `SourceCatalogStep` has NaN values in every pixel.

As part of this effort, I refactored all the save machinery into a separate function.

The current approach is that the step still saves the same files as usual, but these files are empty or contain all-zeros (segmentation image, and source catalog or input model). If desired, I can update this PR to not save files.

Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/15146463903

CC: @schlafly 

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/source_catalog/source_catalog_step.py
+++ b/romancal/source_catalog/source_catalog_step.py
@@ -117,6 +117,14 @@ class SourceCatalogStep(RomanStep):
         else:
             forced = False
 
+        # return empty segmentation image and catalog if all pixels are
+        # masked
+        if np.all(mask):
+            self.log.warning("Cannot create source catalog. All pixels are masked.")
+            segment_img = np.zeros(model.data.shape, dtype=np.uint32)
+            cat = Table()
+            return self.save_all_results(input_model, segment_img, cat)
+
         bkg = RomanBackground(
             model.data,
             box_size=self.bkg_boxsize,
@@ -128,7 +136,7 @@ class SourceCatalogStep(RomanStep):
             model.data, kernel_fwhm=self.kernel_fwhm, mask=mask
         )
 
-        if not forced:
+        if not self.forced_segmentation:
             segment_img = make_segmentation_image(
                 detection_image,
                 snr_threshold=self.snr_threshold,
@@ -160,18 +168,27 @@ class SourceCatalogStep(RomanStep):
             else:
                 cat_type = "prompt"
 
-            catobj = RomanSourceCatalog(
-                model,
-                segment_img,
-                detection_image,
-                self.kernel_fwhm,
-                fit_psf=self.fit_psf & (not forced),  # skip when forced
-                mask=mask,
-                cat_type=cat_type,
-            )
-            cat = catobj.catalog
+        # return empty segmentation image and catalog if no sources are detected
+        if segment_img is None:
+            self.log.warning("Cannot create source catalog. No sources were detected.")
+            segment_img = np.zeros(model.data.shape, dtype=np.uint32)
+            cat = Table()
+            return self.save_all_results(input_model, segment_img, cat)
 
-        if forced:
+        cat_type = "prompt" if not self.forced_segmentation else "forced_det"
+        fit_psf = self.fit_psf & (not self.forced_segmentation)  # skip when forced
+        catobj = RomanSourceCatalog(
+            model,
+            segment_img,
+            detection_image,
+            self.kernel_fwhm,
+            fit_psf=fit_psf,
+            mask=mask,
+            cat_type=cat_type,
+        )
+        cat = catobj.catalog
+
+        if self.forced_segmentation:
             # TODO: improve this so that the moment-based properties are
             # not recomputed from the forced_detection_image
             forced_detection_image = forced_segmodel.detection_image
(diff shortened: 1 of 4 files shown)


larrybradley:
@schlafly Good point.  The table is completely empty (no columns, no data), so it's not going to pass validation unless there's a condition to skip empty tables.

Would you prefer that the step does write any files in this case (unless `save_results` is explicitly True)?  Is the ops pipeline going to crash if it expects files and they aren't written?

braingram:
> @schlafly Good point. The table is completely empty (no columns, no data), so it's not going to pass validation unless there's a condition to skip empty tables.
> 
> Would you prefer that the step does not write any files in this case (unless `save_results` is explicitly True)? Is the ops pipeline going to crash if it expects files and they aren't written?
> 
> EDIT: we probably need to add that empty-table condition to catalog validator because the step will still return an empty table, and if the user sets `save_results` we don't want the validation to fail.

How about saving a table with columns but no rows? That way validation could still check for the required columns. If we allow the schema to pass with an empty table (no column) than if a bug produces a table with no columns we won't catch that with the schema.

schlafly:
I think I like producing a valid table with no rows, but that seems like it might be a pain and I don't think it's worth working too hard to produce an empty catalog.  Brett, do you think it will be easy to make an empty table from rad/rdm in this case?

braingram:
Can we delay this until the table schemas are added?

Also I don't think this closes https://github.com/spacetelescope/romancal/issues/1765 I believe @mcara has some comments and perhaps updates with respect to that issue and resample.

larrybradley:
> Also I don't think this closes https://github.com/spacetelescope/romancal/issues/1765 I believe @mcara has some comments and perhaps updates with respect to that issue and resample.

Yes, #1765 is really a couple different issues.  I'll remote auto-close for that issue.

braingram:
> Does stpipe raise an error if a step returns None and save_results=True? If not, we could use that approach (coupled with not saving empty data).

I don't believe this is a problem for stpipe (however looking at how [finalize_result is called](https://github.com/spacetelescope/stpipe/blob/3b151b458fe87a9f5f86a05a0c18fcfe994f5a20/src/stpipe/step.py#L606-L615) and is [implemented in romancal](https://github.com/spacetelescope/romancal/blob/0ace3edc283437e5aba7644094baff1cf3060551/romancal/stpipe/core.py#L76) it may fail on `None`).

mcara:
@larrybradley The all NaN result seems reasonable to me, see https://github.com/spacetelescope/romancal/issues/1765#issuecomment-2899617002

mcara:
> @mcara ... why are we still producing a co-add file (with all NaNs)? Seems sensible to not produce any files in this case.

I am not sure what exactly the question refers to: how the pipeline works or `drizzle` package? If it is about the pipeline - I don't know. Seems like it would be extra effort to check overlap between the input image and output image but it would be faster and _possibly_ avoid creating output image at all. It would be duplication since it is done in `drizzle` anyway albeit not as efficient as it could have been done with `spherical_geometry`. On the other hand, evaluating the maximum width (or height) of a polygon in spherical geometry in units of input pixels is also an interesting problem in itself (+ make sure the narrow part falls exactly over one pixel). So, while it may be interesting theoretically, practically it would be more difficult to implement. 

If the question is about `drizzle` then the answer is different. In principle `add_image` method is called per input image and it is not aware of how many input images will be resampled. If first image does not overlap, maybe next one will? In principle, it would be possible to not create output arrays until the first image with a valid overlap is encountered. In practice, it is not very easy as the code is written in C and expects arrays to be supplied by the Python code. So Python essentially creates the arrays and passes them to C code. Also, even if one image has a reasonable overlap, how do we know that not all pixels in the overlap are flagged in DQ arrays and will contribute to the output? We would have to scan and check for this at all time and decrease code performance. Nothing is "impossible" but is it worth the effort?

romancal/source_catalog/source_catalog_step.py
```diff
@@ -95,39 +95,29 @@ def process(self, step_input):
...
-            source_catalog_model.meta["forced_segmentation"] = self.forced_segmentation
-            forced = True
-        else:
-            forced = False
+        # return empty segmentation image and catalog if all pixels are
+        # masked
+        if np.all(mask):
+            self.log.warning("Cannot create source catalog. All pixels are masked.")
+            segment_img = np.zeros(model.data.shape, dtype=np.uint32)
+            cat = Table()
```
braingram, line 103: @larrybradley thanks for your patience with this. There is new API for the various source catalog models that should allow this to be replaced with a call to `create_empty_catalog`:
https://github.com/spacetelescope/romancal/blob/803b2fe7388108cec7f6f9d835f208f55c2b1a77/romancal/source_catalog/source_catalog_step.py#L156
to generate a schema-valid catalog with no rows.

larrybradley: Thanks, @braingram.  I've updated this PR.

schlafly (approved):
Thanks Larry, looks good to me!

PR #1772, opened 2025-05-21
[RCAL-1084] include distortion when matching sky cells to image footprint

zacharyburnett:
Resolves [RCAL-1084](https://jira.stsci.edu/browse/RCAL-1084)

This PR addresses an issue referenced in https://github.com/spacetelescope/romancal/issues/1765, namely that the image footprint is not necessarily a rectangle with perfectly straight sides; rather, it may be distorted (modeled by the WCS).

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/skycell/match.py
+++ b/romancal/skycell/match.py
@@ -43,14 +43,18 @@ class ImageFootprint:
         wcs: WCS,
         extra_vertices_per_edge: int = 0,
     ) -> "ImageFootprint":
-        """
-        create an image footprint from a GWCS object (and image shape, if no bounding box is present)
+        """create an image footprint from a GWCS object (and image shape, if no bounding box is present)
 
         Parameters
         ----------
+        wcs: WCS :
+            WCS object
+        extra_vertices_per_edge: int :
+            extra vertices to create on each edge to capture distortion (Default value = 0)
 
-        wcs: WCS object
-        extra_vertices_per_edge: extra vertices to create on each edge to capture distortion
+        Returns
+        -------
+        image footprint object
         """
 
         # if `.pixel_shape` is not defined, try deriving it from the `.bounding_box`
@@ -195,13 +199,7 @@ class ImageFootprint:
 
     @cached_property
     def possibly_intersecting_projregions(self) -> int:
-        """
-        number of possibly intersecting projection regions
-
-        NOTES
-        -----
-        this assumes a non-degenerate tesselation!
-        """
+        """number of possibly intersecting projection regions"""
         if self.area > sc.SkyCell.area:
             return (
                 # the number of times the smallest projection region could fit in the image footprint
@@ -215,13 +213,7 @@ class ImageFootprint:
 
     @cached_property
     def possibly_intersecting_skycells(self) -> int:
-        """
-        number of possibly intersecting skycells
-
-        NOTES
-        -----
-        this assumes a non-degenerate tesselation!
-        """
+        """number of possibly intersecting skycells"""
         if self.polygon.area() > sc.SkyCell.area:
             return (
                 # number of times a skycell could fit in the image footprint
@@ -254,17 +246,16 @@ def find_skycell_matches(
     image_corners: list[tuple[float, float]] | NDArray[float] | WCS,
     skymap: sc.SkyMap = None,
 ) -> list[int]:
-    """
-    Find sky cells overlapping the provided image footprint
+    """Find sky cells overlapping the provided image footprint
 
     Parameters
     ----------
-    image_corners :
+    image_corners : list | np.ndarray | WCS :
         Either a squence of 4 (ra, dec) pairs, or
         equivalent 2-d numpy array, or a GWCS instance.
         A GWCS instance must have `.bounding_box` or `.pixel_shape` attribute defined.
-    skymap: SkyMap
-        sky map instance (defaults to global SKYMAP)
+    skymap : sc.SkyMap :
+        sky map instance; defaults to global SKYMAP (Default value = None)
 
     Returns
     -------
--- a/romancal/skycell/match.py
+++ b/romancal/skycell/match.py
@@ -13,7 +13,6 @@ import spherical_geometry.polygon as sgp
 import spherical_geometry.vector as sgv
 from gwcs import WCS
 from numpy.typing import NDArray
-from stcal.alignment.util import wcs_bbox_from_shape
 
 import romancal.skycell.skymap as sc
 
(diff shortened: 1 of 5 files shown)


zacharyburnett:
currently, creating a vertex for every edge pixel incurs a significant performance cost when using `spherical_geometry`. I'll investigate @schlafly's suggestion tomorrow, of limiting the amount of vertices per side of the image footprint.

zacharyburnett:
added `step` (pixel interval) to control the amount of vertices on each edge. Right now, `step=100`

romancal/assign_wcs/utils.py
```diff
@@ -29,8 +29,29 @@ def wcs_bbox_from_shape(shape):
...
+
+    Returns
+    -------
+    shape : tuple
+        Pixel shape x, y order.
+    """
+    return (
+        bbox[0][1] + 0.5,
+        bbox[1][1] + 0.5,
+    )
```
schlafly, line 54: This seems right for the normal case where bbox = [[-0.5, shape - 0.5], [-0.5, shape - 0.5]].  As a generic function, though, I wonder if [r[1] - r[0] for r in bbox] is more appropriate?

romancal/skycell/match.py
```diff
@@ -26,88 +26,177 @@ class ImageFootprint:
...
 
-            iwcs.bounding_box = wcs_bbox_from_shape(image_shape)
+        # pixel_shape is in x, y order contrary to numpy convention.
+        image_shape = (wcs.pixel_shape[1], wcs.pixel_shape[0])
+
+        if vertices != (0, 0):
+            if not hasattr(wcs, "bounding_box") or wcs.bounding_box is None:
+                wcs.bounding_box = wcs_bbox_from_shape(image_shape)
+
+            vertex_points = wcs.footprint(center=False)
```
schlafly, line 70: Does this say: if the user specified the vertices, ignore them and replace them with wcs.footprint(...)?  Maybe the intent is "if vertices == (0, 0)"?

romancal/skycell/match.py
```diff
@@ -26,88 +26,177 @@ class ImageFootprint:
...
+                                    endpoint=False,
+                                )
+                            ),
+                            np.repeat(0, repeats=vertices[0] + 1),
+                        ],
+                        axis=1,
+                    ),
+                ],
+                axis=0,
+            )
```
schlafly, line 139: I think I'd suggest making "vertices" a scalar rather than a tuple "how many points do you want per side".  I could imagine someone wanting to specify different numbers for the different sides, but not for Roman, and it makes this code meaningfully more complicated.  Maybe this block would look more like:
np.array([[zeros, xverts], [yverts, xend], [yend, xend - xverts], [yend - yverts, zeros]])
with initial definitions of zeros = np.zeros(vertices), xverts = np.linspace(0, shape[0], vertices, endpoint=False), ...
which might be more legible?

schlafly (commented):
This looks good.  Did you look at any example images to see how many vertices to put as a good default?  From other work it looks to me like the distortion leads to a 2 pixel error at the middle of the detector if there is are only vertices in the corner.  I will bet that that decreases like the square of the length between the vertices, so I would expect that it is 0.5 pix with two segments per side and 0.125 pix with three segments per side.  I did not see any buffer in this code, though; I think my suggestion would be to add one or three extra vertices (corresponding to 2-4 segments per side), and then to add 1 pixel buffer.

romancal/skycell/match.py
zacharyburnett: ah, good catch!

romancal/skycell/match.py
zacharyburnett: done in 34cbca7d0b75cdab95747a8e0cb744d127bfe730

romancal/skycell/match.py
```diff
@@ -26,88 +26,177 @@ class ImageFootprint:
...
+                image_shape[0] - 1 if vertices[0] >= image_shape[0] else vertices[0],
+                image_shape[1] - 1 if vertices[1] >= image_shape[1] else vertices[1],
+            )
+
+            # build a list of pixel indices that represent equally-spaced edge vertices
+            edge_pixel_indices = np.concatenate(
+                [
+                    np.stack(
+                        [
+                            np.repeat(0, repeats=vertices[1] + 1),
```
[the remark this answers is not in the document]
zacharyburnett: done in 1bef5c3dc34e0cfdcd2e1b010343e087d64afcc2

romancal/assign_wcs/utils.py
zacharyburnett: done in 7330f43536bd53ef89746eca3cd8ae0f9176b4e2

romancal/skycell/match.py
```diff
@@ -28,88 +28,165 @@ class ImageFootprint:
...
-                        "`image_shape` was not specified and "
-                        "the GWCS object does not have `.bounding_box` nor `.pixel_shape`."
-                    )
+        """
+        create an image footprint from a GWCS object (and image shape, if no bounding box is present)
+
+        Parameters
+        ----------
+
+        wcs: WCS object
```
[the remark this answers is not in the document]
zacharyburnett: done in 8f3c2e83fe1b8fd268df47c22c66ed799733f493

romancal/skycell/match.py
```diff
@@ -28,88 +29,167 @@ class ImageFootprint:
...
+        """
+
+        # if `.pixel_shape` is not defined, try deriving it from the `.bounding_box`
+        if not hasattr(wcs, "pixel_shape") or wcs.pixel_shape is None:
+            if hasattr(wcs, "bounding_box") and wcs.bounding_box is not None:
+                wcs.pixel_shape = wcs_pixel_shape_from_bbox(wcs.bounding_box)
+            else:
+                raise ValueError(
+                    "Cannot infer image footprint from WCS without `.pixel_shape` nor `.bounding_box`."
+                )
```
nden, line 68: I think this code is too convoluted. Is `ImageFootprint.from_wcs` run on L3 images?
 I would assume if `bounding_box` exists, then `pixel_shape` should exist as well. If it's not set, set it when the WCS is generated.

zacharyburnett: done in d82370341bec1c0d59fdd929eac66e463dbe454c

zacharyburnett:
new regtest at https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml

zacharyburnett:
unexpected behavior with matching:
https://github.com/spacetelescope/romancal/actions/runs/15612410372/job/43976423935?pr=1772#step:10:471
```
 At index 0 diff: '000p86x62y68' != '000p86x62y69'
  Right contains one more item: '045p86x38y69'
  
  Full diff:
    [
  +     '000p86x62y68',
  +     '000p86x61y68',
        '000p86x62y69',
  -     '000p86x62y68',
        '000p86x61y69',
  -     '000p86x61y68',
  ?              ^  ^
  +     '000p86x62y67',
  ?              ^  ^
        '000p86x63y69',
  -     '000p86x61y70',
  -     '000p86x62y67',
        '000p86x60y68',
        '045p86x37y69',
        '045p86x37y68',
        '045p86x38y69',
    ]
```
I'll investigate this

zacharyburnett:
aha! A corner case!
corner_case

romancal/skycell/match.py
schlafly: For these, do we want the edges to be at (-0.5, shape - 1) or (0, shape)?  I think I might guess the former but we're doing the latter?

And here is how I would recommend this block (untested!):

```suggestion
    zeros = np.zeros(extra_vertices_per_edge + 1)
    xend = shape[0] - zeros
    yend = shape[1] - zeros
    xverts = np.linspace(0, shape[0], num = extra_vertices_per_edge, endpoint=False)
    yverts = np.linspace(0, shape[1], num = extra_vertices_per_edge, endpoint=False)
    edge_indices = np.array([
        np.stack([zeros, yverts], axis=1),
        np.stack([xverts, yend], axis=1),
        np.stack([xend, yend - yverts], axis=1),
        np.stack([xend - xverts, zeros], axis=1)])
```

I think that's the same thing and it is more legible, at least to me?  The key thing is to be able to track that one is moving along the different sides in the right order, and that's easier for me when I don't need to reparse the linspaces & zeros.

I've written this following the existing (0, shape) logic, though as I mentioned it's possible this should all be shifted by a half pixel, either here or when calling the wcs(...).

schlafly (approved):
Thanks Zach, looks good to me.  I'll go ahead and merge?

PR #1785, opened 2025-06-04
strip read tags when ramp step constructs the output image model

braingram:
This PR addresses a few failures in the devdeps due to existing rad changes. Here is the latest run showing several failures for ramp fitting due to a tag mismatch:
https://github.com/spacetelescope/RegressionTests/actions/runs/15430372053

This PR updates ramp fitting to copy the input nodes and strip their "read tags" to allow them to be saved with the newest tag versions.

Regtests with this PR: https://github.com/spacetelescope/RegressionTests/actions/runs/15443040145
show 1 failure for the elp run of tvac data:
```
{'dictionary_item_removed': ["root['roman']['meta']['cal_step']['jump']", "root['roman']['meta']['cal_step']['source_detection']"]}
```
showing that the output image model now no longer contains cal_step entries for the non-existent jump and source_detection steps.

I think it makes sense to also update the rad pin in roman_datamodels back to "main" so that rad changes result in nightly instead of devdeps failures. https://github.com/spacetelescope/roman_datamodels/pull/528

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/changes/1785.ramp_fitting.rst
+++ b/changes/1785.ramp_fitting.rst
@@ -1,0 +1,1 @@
+Use ``ImageModel.create_minimal`` to create output model.
--- a/romancal/ramp_fitting/ramp_fit_step.py
+++ b/romancal/ramp_fitting/ramp_fit_step.py
@@ -6,8 +6,8 @@ import logging
 from typing import TYPE_CHECKING
 
 import numpy as np
-from roman_datamodels import datamodels as rdd
-from roman_datamodels import stnode as rds
+from roman_datamodels import datamodels as rdm
+from roman_datamodels import stnode
 from roman_datamodels.dqflags import group, pixel
 from stcal.ramp_fitting import ols_cas22_fit
 from stcal.ramp_fitting.ols_cas22 import Parameter, Variance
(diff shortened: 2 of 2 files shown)


schlafly:
@stscieisenhamer , can you remind me what we decided the "right" thing to do with TVAC/FPS metadata was?  My recollection is something like the following:
- We process the files and copy over the metadata that matches.
- We pass along any extra things without looking at them.

I think if we make the rule that "unrecognized things at the root level of roman.meta gets passed along unchanged", that might be a simple enough rule?

braingram:
@schlafly I updated this to copy the metadata (including non-schema defined attributes) and think this is now ready to go. There is one regtest difference that I think is an improvement. For the TVAC data the test shows:
```
{'dictionary_item_removed': ["root['roman']['meta']['cal_step']['jump']", "root['roman']['meta']['cal_step']['source_detection']"]}
```
Since the output `ImageModel` no longer has a cal_step entry for the removed jump and source_detection steps.

schlafly (approved):
Thank you!  Making sure I understand: deleting the read tags means that ~asdf / stnode no longer knows what tag the various dictionaries are, but when it eventually goes to save the file it adds these back and makes sure that everything matches the top level ImageModel specification?  Thanks!

braingram:
> Thank you! Making sure I understand: deleting the read tags means that ~asdf / stnode no longer knows what tag the various dictionaries are, but when it eventually goes to save the file it adds these back and makes sure that everything matches the top level ImageModel specification? Thanks!

Mostly yes. Stripping the `_read_tag` (which we probably want some public API for at some point) will cause `Foo.tag` to return the latest tag. Since in this case the ImageModel is newly created, it also has the newest tag.

So taking `L2CalStep` as an example, if the input model is read from a file (as is done in the regtests) the `L2CalStep` instance will get a `_read_tag` of whatever version is in the file (in this case `1.1.0`). On main the `L2CalStep` from the input model is referenced in the output model (it's probably a bug that this wasn't a copy but that's fixed in this PR as well). This results in the new `ImageModel` getting a version `1.1.0` `L2CalStep` when it expects a `1.2.0` and failing validation when it goes to save. With this PR the `L2CalStep` from the input model is copied (which also copies the `_read_tag` so still `1.1.0` after the copy) and then the `_read_tag` is stripped which allows the `L2CalStep` instance to pick up the latest (default tag).

PR #1833, opened 2025-07-07
Cache query to catalog service

WilliamJamieson:
This PR attempts to prevent our pounding of the `https://gsss.stsci.edu/webservices` server by having romancal cache the query results.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/tweakreg/astrometric_utils.py
+++ b/romancal/tweakreg/astrometric_utils.py
@@ -1,4 +1,9 @@
+from __future__ import annotations
+
 import os
+from multiprocessing import Manager
+from multiprocessing.managers import DictProxy, SyncManager
+from threading import Lock
 
 import requests
 from astropy import table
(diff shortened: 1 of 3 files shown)


romancal/tweakreg/tests/test_astrometric_utils.py
```diff
@@ -696,26 +701,19 @@ def test_get_catalog_using_epoch(ra, dec, epoch):
...
-    with pytest.raises(Exception) as exec_info:
-        for dt in np.arange(1, 0, -0.01):
-            try:
-                get_catalog(10, 10, sr=0.1, catalog="GAIADR3", timeout=dt)
-            except requests.exceptions.ConnectionError:
-                # ignore if it's a connection error instead of timeout
-                pass
+    monkeypatch.setattr("requests.get", MockTimeoutError)
 
-    assert exec_info.type == requests.exceptions.Timeout
```
WilliamJamieson, line 710: I make this test simplification simply because `get_catalog(10, 10, sr=0.1, catalog="GAIADR3")` can be called by earlier tests meaning the cache will actually have the value and no connection will happen. We could simulate a real timeout if we make sure this test is run before any other tests or choose some set of inputs that are valid query inputs because you get a connection error if one makes a query with inputs that the server doesn't know how to handle because this is essentially just looking at a URL.

romancal/tweakreg/astrometric_utils.py
```diff
@@ -175,6 +180,101 @@ def compute_radius(wcs):
...
+            Set the request timeout (in seconds). Default: 30 s.
+
+        Returns
+        -------
+            A Table object of returned sources with all columns as provided by catalog.
+
+        """
+
+        with self._lock:
+            service_url = f"{SERVICELOCATION}/vo/CatalogSearch.aspx?RA={ra}&DEC={dec}&EPOCH={epoch}&SR={sr}&FORMAT=CSV&CAT={catalog}&MINDET=5"
```
WilliamJamieson, line 242: I stuffed everything into a single f-string, which I am using as the cache key

romancal/tweakreg/astrometric_utils.py
```diff
@@ -175,6 +180,101 @@ def compute_radius(wcs):
...
 
+class _CatalogCacheSingleton:
+    _instance: _CatalogCacheSingleton | None = None
+    _initalized: bool = False
+
+    def __new__(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance = super().__new__(cls, *args, **kwargs)
+
+        return cls._instance
```
WilliamJamieson, line 191: This turns the object into a singleton meaning on any given process/thread one gets the same object. The singletons maybe different actual objects on different threads but this is dealt with below.

WilliamJamieson:
I ran the CI three times in a row with no failures. So it seems robust-ish.

schlafly:
Thanks William.  This looks like a good approach.  Since we're going to delete this with #1835, we should probably close this PR, but I think it would be helpful in stcal.

PR #1843, opened 2025-07-14
Improve wcsinfo accuracy.

schlafly:
This PR improves the accuracy with which the wcsinfo round trips in romancal.

romancal computes a bunch of information for the wcs.  One of these terms is the orientation, and another is the projection matrix.  The former is used directly in the WCS, while the latter contains the same information but is a slightly different computation (e.g., it can be computed as `np.degrees(np.arcsin(w.forward_transform.matrix_2.value[0, 1]))`.  In the new L3 metadata, Tyler proposes removing the rotation matrix and keeping only the orientation, so that the wcsinfo does not contain redundant information.  When making that change, the WCS began to fail a round-trip test.  This PR makes a few changes so that that round-trip test continues to pass.

As Brett & Nadia reported, this was only a roundoff error; there were no real code errors.  The regtest test documentation was incorrect, and the errors were reported in mas; the test required that the WCS agree at the 10^-5 mas level.  That's fairly stringent!  This PR does a few things:
- it changes the rotation computation to have a longer level arm to reduce the error in the orientation computation
- it scales the error in the ra direction by the cosine of the declination, so that locations near the pole do not artificially get large ra errors.
- it _does not_ fix a bug Brett noticed in the use of vparity.
- it updates the docs

The parity is a challenge.  We don't record the parity in the L3 metadata; in romancal, we assume +1.  If that assumption is wrong and the original WCS had reversed parity, we don't record that anywhere except in the rotation matrix.  However, stcal uses -1 by default, while our skycells use +1.  I've left in the +1 until the skycells reference file is updated.  Then we will probably need to update this to use -1 and also update some unit tests.

With this PR, the test now passes.  Formerly the standard deviation was ~8e-5 mas; now it is ~2e-7 mas.  Roman WCS accuracy requirements are ~1 mas, so this is in any case incredibly small.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/regtest/util.py
+++ b/romancal/regtest/util.py
@@ -28,8 +28,12 @@ def comp_wcs_grids_arcs(wcs_a, wcs_b, npix=4088, interval=10):
     ra_b, dec_b = wcs_b(xx, yy, with_bounding_box=False)
     dec_med = np.nanmedian(dec_b)
 
-    ra_mad = mad_std((ra_a - ra_b) * np.cos(np.radians(dec_med)), ignore_nan=True
-                     ) * 60.0 * 60.0 * 1000.0
+    ra_mad = (
+        mad_std((ra_a - ra_b) * np.cos(np.radians(dec_med)), ignore_nan=True)
+        * 60.0
+        * 60.0
+        * 1000.0
+    )
     dec_mad = mad_std(dec_a - dec_b, ignore_nan=True) * 60.0 * 60.0 * 1000.0
 
     return ra_mad, dec_mad
--- a/romancal/regtest/util.py
+++ b/romancal/regtest/util.py
@@ -21,13 +21,15 @@ def comp_wcs_grids_arcs(wcs_a, wcs_b, npix=4088, interval=10):
     Returns
     -------
     mad_std : float
-        The numpy MAD_STD in arcseconds
+        The numpy mad_std in mas
     """
     xx, yy = np.meshgrid(np.linspace(0, npix, interval), np.linspace(0, npix, interval))
     ra_a, dec_a = wcs_a(xx, yy, with_bounding_box=False)
     ra_b, dec_b = wcs_b(xx, yy, with_bounding_box=False)
+    dec_med = np.nanmedian(dec_b)
 
-    ra_mad = mad_std(ra_a - ra_b, ignore_nan=True) * 60.0 * 60.0 * 1000.0
+    ra_mad = mad_std((ra_a - ra_b) * np.cos(np.radians(dec_med)), ignore_nan=True
+                     ) * 60.0 * 60.0 * 1000.0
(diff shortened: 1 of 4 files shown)


romancal/regtest/util.py
```diff
@@ -21,13 +21,19 @@ def comp_wcs_grids_arcs(wcs_a, wcs_b, npix=4088, interval=10):
     Returns
     -------
     mad_std : float
-        The numpy MAD_STD in arcseconds
+        The numpy mad_std in mas
     """
     xx, yy = np.meshgrid(np.linspace(0, npix, interval), np.linspace(0, npix, interval))
     ra_a, dec_a = wcs_a(xx, yy, with_bounding_box=False)
     ra_b, dec_b = wcs_b(xx, yy, with_bounding_box=False)
```
braingram, line 28: For the default arguments (which are used in the [regtests](https://github.com/spacetelescope/romancal/blob/a85b200ff7d726a7f8beddd6298511f00b73f314/romancal/regtest/test_mos_skycell_pipeline.py#L72)) this generates a grid spanning 0 to 4088. For skycell wcs (5000 x 5000) this doesn't cover the whole wcs and for the [other regtest use](https://github.com/spacetelescope/romancal/blob/a85b200ff7d726a7f8beddd6298511f00b73f314/romancal/regtest/test_mos_pipeline.py#L132) the wcs bounding box is 4138 x 4389 (so this funtion also under samples that wcs).

Perhaps it makes sense to remove this function and instead use the other [wcs comparison utility](https://github.com/spacetelescope/romancal/blob/a85b200ff7d726a7f8beddd6298511f00b73f314/romancal/regtest/regtestdata.py#L650) that's used for compare_asdf.

braingram: More generally for the wcs -> wcsinfo -> wcs comparisons. Would it make sense to compare these using difference in pixel coordinates instead of world coordinates?

schlafly: I don't feel strongly about on-sky vs. pixel coordinates.  That said, ultimately our requirements are in on-sky coordinates and I don't want the requirements to be tighter for heavily oversampled mosaics, so on-sky makes sense.

I agree with you that this should be using the bounding boxes to get the right number of pixels and isn't doing this.

I agree that this is duplicated wrt the compare_asdf version (which I forgot about while fixing this one!).  In both of those cases we should probably be correcting by cos(dec), the part added here, and really I should be looking at differences of positions on the sky.  I would spin this off into a separate issue though.

braingram: Thanks for the detailed response. That helped me understand the problem. If it works for you we could go ahead with this PR then address consolidating and improving the comparisons in a follow-up.

braingram (approved):
Thanks!

PR #1978, opened 2025-09-27
RCAL-1125: Generate SI Cosmos and Point Source Catalog

### LINKED ISSUE #1876: Generate SI Catalog
_Issue [RCAL-1125](https://jira.stsci.edu/browse/RCAL-1125) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Add source injection routine that generates an artificial catalog of COSMOS objects via RSIM

PaulHuwe:
Resolves [RCAL-1125](https://jira.stsci.edu/browse/RCAL-1125)

Closes #1876 

This PR adds a method to generate cosmos galaxies and point sources for source injection.

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/source_catalog/injection.py
+++ b/romancal/source_catalog/injection.py
@@ -85,7 +85,7 @@ def make_cosmoslike_catalog(cen, xpos, ypos, exptimes, filters=None, seed=50, **
 
     Parameters
     ----------
-    coord : astropy.coordinates.SkyCoord
+    cen : astropy.coordinates.SkyCoord
         Location around which to generate sources.
     xpos, ypos : array_like (float)
         x, y positions of each source in objlist
@@ -203,7 +203,6 @@ def make_cosmoslike_catalog(cen, xpos, ypos, exptimes, filters=None, seed=50, **
                 HRPOINTMAGLIMIT[bp]
                 + (1.25 * np.log10((exptimes[bp] * u.s).to(u.hour).value))
             )
-
     point_mag_limit = max(point_band_mag_limit)
 
     # Set bandpass fluxes


romancal/source_catalog/injection.py
```diff
@@ -40,3 +46,93 @@ def inject_sources(model, si_cat):
...
+    # a full sample.
+    gal_cat = catalog.make_cosmos_galaxies(cen, radius=1.0, cat_area=(np.pi), **kwargs)
+
+    # Trim to the required number of objects
+    gal_cat = gal_cat[:num_gals]
+
+    # Set the sizes
+    gal_cat["half_light_radius"] = -2.5 * np.log10(
+        gal_cat[filter] + rng_numpy.normal(num_gals)
+    )
```
schlafly, line 104: I updated the doc to contain more information on this equation; this isn't quite what we want.  There's a new figure intending to illustrate this, but the upshot is:

> Accordingly, we recommend: sigma = 0.15, mu = -0.1555 * (J_AB - 17) - 3.55, and drawing points x normally distributed at each magnitude with that standard deviation.  Then converting to a size in arcseconds of (10^x) / 3600.

We could use the Roman J band magnitude, though I used the UVISTA one when developing that relation.  I'm still thinking about how exactly the COSMOS fields connect to the half light radius, so I might adjust the relationship in the future.  But this one has the right form---the logarithm is normal, so you want to be taking something normally distributed and taking 10^x of it, not the other way around.

romancal/source_catalog/injection.py
```diff
@@ -40,3 +46,93 @@ def inject_sources(model, si_cat):
...
+    star_cat["dec"] = num_stars * [0]
+    star_cat["type"] = num_stars * ["PSF"]
+    star_cat["n"] = num_stars * [-1]
+    star_cat["half_light_radius"] = num_stars * [0]
+    star_cat["pa"] = num_stars * [0]
+    star_cat["ba"] = num_stars * [1]
+
+    # Set the point magnitude limit
+    point_mag_limit = HRPOINTMAGLIMIT + (
+        1.25 * np.log10((exptime * u.s).to(u.year).value)
```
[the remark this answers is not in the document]
PaulHuwe: Fixed

romancal/source_catalog/injection.py
```diff
@@ -4,11 +4,17 @@
...
+import numpy as np
+from astropy import table
+from astropy import units as u
 from roman_datamodels.datamodels import ImageModel, MosaicModel
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+# One hour point source magnitude limit for F213
+HRPOINTMAGLIMIT = 25.64
```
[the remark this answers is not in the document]
PaulHuwe: Implemented

romancal/source_catalog/injection.py
PaulHuwe: Done, with the * 3600 correction.

romancal/source_catalog/injection.py
```diff
@@ -40,3 +69,120 @@ def inject_sources(model, si_cat):
...
+
+    # Set magnitude spread
+    mags = rng_numpy.uniform(low=-4.0, high=1.0, size=num_stars)
+
+    # Set bandpass fluxes
+    for bp in BANDPASSES:
+        point_mag_limit = HRPOINTMAGLIMIT[bp] + (
+            1.25 * np.log10((exptime * u.s).to(u.hour).value)
+        )
+        star_cat[bp] = (10.0 ** (-(mags + point_mag_limit) / 2.5)).astype("f4")
```
schlafly, line 179: I think we want to have the stars all have color = 0.  This makes all of the stars have colors point_mag_limit[x] - point_mag_limit[y]?

romancal/source_catalog/injection.py
```diff
@@ -40,3 +69,120 @@ def inject_sources(model, si_cat):
...
+        # Normalize the mag limit to exptime
+        gal_mag_limit = HRGALMAGLIMIT[bp] + (
+            1.25 * np.log10((exptime * u.s).to(u.hour).value)
+        )
+        mag_tot = mags + gal_mag_limit
+
+        # Set scaled fluxes
+        gal_cat[bp] = (gal_cat[bp] * (10.0 ** (-(mag_tot) / 2.5)) / deepflux).astype(
+            "f4"
+        )
```
schlafly, line 148: We want the galaxies to retain their COSMOS colors and scale only their overall brightness so that they land in [-4, +1] around the faint limit.  Choosing the galaxy faint limit rather than the point source faint limit is reasonable.  But since the gal_mag_limits are different in the different bands, this will change the colors of all of the galaxies by some amounts depending on Roman's sensitivity and the exposure times in the different bands?

My suggestion was to take the "deepest band", i.e., max(gal_mag_limit) over the bands available.  I think you want to take bandpasses and exptimes and arguments, and have them be the list of exposure times and bands that are present in the multiband association you're getting (meta.coadd_info.exposure_time).  Then cosmos_flux = gal_cat_data[deepest_band], and gal_cat[bp] = gal_cat[bp] * 10 ** (-(mags + max(gal_mag_limit)) / 2.5) / cosmos_flux.

I think I have a vague preference for using gal_cat_data[deepest_band] rather than deepflux.  The case I'm imagining is the one where a galaxy drops out of the blue bands due to LyA forest absorption, but is comparably very bright in the red bands.  I think your approach would make the drop out bands always near the detection limit, so that the red bands were always very bright.  That said, if you took the opposite approach and said that the brightest band was the one that things got rescaled to, I think that could work.

In the context of the related romanisim PR, I worry about sources with 1e-30 flux---they're positive, so they don't get removed, but they're tiny and we don't trust them.  Then maybe they have secure detections in some other band, so if you rescale the tiny flux so that it's near the Roman detection limit, you might take the secure flux and turn it into some hurt-your-eyes bright value.  I'd make sure that the COSMOS catalog doesn't have a long tail to faint magnitudes and / or you cut these out, possibly via an SNR cut (rather than just a "flux > 0" cut).

romancal/source_catalog/tests/test_injection.py
```diff
@@ -200,3 +201,81 @@ def test_inject_sources(image_model, mosaic_model):
...
+    # Set wcs metadata
+    meta = {
+        "wcsinfo": {
+            "ra_ref": RA,
+            "dec_ref": DEC,
+            "roll_ref": ROLL,
+        }
+    }
+
+    mcat = make_catalog(meta)
```
schlafly, line 237: Let's call this with the explicit seed as well.

romancal/source_catalog/tests/test_injection.py
```diff
@@ -200,3 +201,81 @@ def test_inject_sources(image_model, mosaic_model):
...
+
+    # Ensure that locations are as expected
+    assert np.allclose(np.sort(cat["ra"]), np.sort(mcat["ra"]), rtol=1.0e-6)
+    assert np.allclose(np.sort(cat["dec"]), np.sort(mcat["dec"]), rtol=1.0e-6)
+
+    # Ensure correct number of point sources
+    assert np.sum(cat["type"] == "PSF") == 1
+    assert np.sum(cat["n"] == -1) == 1
```
schlafly, line 245: Suggest len(XPOS_IDX) / 4 instead of 1.

romancal/source_catalog/injection.py
PaulHuwe: It does, which is what I thought you wanted it changed to in your previous comment (when switching to per band point mag limit). I can add an offset to preserve color = 0.

PR #2018, opened 2025-10-27
Add reference read to science data

schlafly:
This PR adds the romancal code for using the new reference_read and reference_amp33 information from raw files if present.  The corresponding rad ticket is here:
https://github.com/spacetelescope/rad/issues/726

The concept here is that the raw science data are what is actually downlinked from the telescope, which is the raw pixels minus an initial reference read, which is optionally recorded in roman.reference_read (for the WFI pixels) and roman.reference_amp33 (for the virtual amp33 pixels).  For scientific purposes this distinction between reference and science pixels isn't valuable and it is better to combine the two during the dq_init step by adding the reference pixels back to the science pixels and deleting the reference pixels.

The procedure is skipped for TVAC & FPS data since, while those files contain 'reference_read' attributes, it means something else and Tyler directed me to not include the reference read for those data sets.  

I couldn't decide whether this code more logically belonged in roman_datamodels.datamodels.from_science_raw or in the dq_init step.  I ended up putting it in dq_init since the manipulation of the science pixels feels like a romancal thing, while roman_datamodels is more about reformatting.  OTOH removing the reference pixels from the ramp might better fit as a roman_datamodels thing, but I feel like I'd go to romancal first to look at how the ramp gets constructed from the raw data, and then to roman_datamodels afterward.

I took the opportunity to factor some raw image construction code out of the different dq_init tests and into a common function, so that I didn't need to copy this code again in my new test.  I also updated the documentation.

Regression tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/18847913391

## Tasks
  - [X] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job (https://github.com/spacetelescope/RegressionTests/actions/runs/18847913391)

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,11 +69,9 @@ jobs:
         - linux: py311-oldestdeps-cov
         - linux: py311-nolegacypath
         - linux: py312-ddtrace
-        - macos: py312-ddtrace
-        # `tox` does not currently respect `requires-python` versions when creating testing environments;
-        # if this breaks, add an upper pin to `requires-python` and revert this py3 to the latest working version
-        - linux: py3-cov
+        - linux: py313-cov
           coverage: codecov
           pytest-results-summary: true
-        - macos: py3-pyargs
+        - macos: py313-ddtrace
+        - macos: py313-pyargs
           pytest-results-summary: true
--- a/.github/workflows/tests_extra.yml
+++ b/.github/workflows/tests_extra.yml
@@ -63,12 +63,13 @@ jobs:
       envs: |
         - macos: py311
         - macos: py312
+        - macos: py313
         # TODO: merge stdevdeps into devdeps
-        - linux: py3-stdevdeps
+        - linux: py313-stdevdeps
           pytest-results-summary: true
-        - macos: py3-stdevdeps
+        - macos: py313-stdevdeps
           pytest-results-summary: true
-        - linux: py3-devdeps
+        - linux: py313-devdeps
           pytest-results-summary: true
-        - macos: py3-devdeps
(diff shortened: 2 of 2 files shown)


docs/roman/dq_init/description.rst
```diff
@@ -35,10 +35,10 @@ WFI_WIM_ACQ, WFI_WIM_TRACK, WFI_WSM_ACQ1, WFI_WSM_ACQ2, WFI_WSM_TRACK.
...
 -----------------------------------
 
-The output of the initialization step is the RampModel. This is the form of the
-data used throughout most of the exposure pipeline steps. For the most part, the
-meta information between the input raw, or "uncal", model and the ramp model is
-complete.
+The DQ init takes an input raw "uncal" file and outputs a intermediate
+"ramp" model (RampModel) that is the input for all following steps
+until ramp fitting.  For the most part, the meta information between
+the input raw, or "uncal", model and the ramp model is identical.
```
ddavis-stsci: I'm not really sure what this paragraph is trying to say. For instance what is an "intermediate" ramp model?
I think you're trying to say the the initial ramp model from SDF has a data array that are the differences described below? I'd just say that instead of implying we have different ramp models. 

Should we have an attribute that indicates if the data values are differences or absolute values?

schlafly: I'll delete "intermediate".  This model is intermediate for me in the sense that we never write this to disk in operations.  The only 3D object most users are likely to see is the uncal model.

PR #2038, opened 2025-11-05
PSF kernel matching routines

schlafly:
This PR adds new PSF matching functionality to the source_catalog.psf module, as well as associated tests.  The goal is to facilitate getting the right PSF matching kernel.  This ended up being harder than I expected because the band-limited Roman PSFs have FFTs that sharply decline to zero over a small area, so it's challenging to get the right window function to use with the photutils approach without setting everything by hand.

As part of trying to get this right, I replaced the azimuthal averaging function with one that operates similarly, but via the FFT, to avoid introducing power at high frequencies in the Roman PSFs.  This function is much better at delivering zero power in the wings but is worse at the not-band-limited Gaussian test that we already had, so I was forced to reduce the precision of that test.

I also added a downsampling function that is designed to take a PSF or a kernel and downsample while preserving its center.  The matching code starts by creating an oversampled matching kernel from the oversampled PSFs, but we will ultimately need an ordinarily sampled kernel; the downsampling function handles converting that to a downsampled image while preserving the centering.

The new create_convolution_kernel routine does the actual kernel creation.  It regularizes the FFT by adding a small amount into the denominator of the FFT of the matching kernel.  There's a new argument that controls the strength of that regularization, which is scaled to the peak power in the FFT.

I add a utility routine central_stamp which pulls out the central region of an image.  I use this, for example, to pull out the center of the matching kernel, which is where most of the action is.

I then add a bunch of unit tests testing the new routines.

I've added decent docstrings but the real documentation should wait for actual use in romancal steps.

@ddavis-stsci , I think you should be able to use these to get the right convolution kernels.

## Tasks
  - [X] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job (https://github.com/spacetelescope/RegressionTests/actions/runs/19137594170)

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>

### DIFF
--- a/romancal/source_catalog/psf.py
+++ b/romancal/source_catalog/psf.py
@@ -286,7 +286,7 @@ def get_gridded_psf_model(psf_ref_model, focus=0, spectral_type=1):
     oversample = psf_ref_model.meta.oversample
     pixel_response_kernel = Box2DKernel(width=oversample)
     for i in range(psf_images.shape[0]):
-        psf = psf_images[i, :, :].copy()
+        psf = psf_images[i, :, :]
         im = convolve(psf, pixel_response_kernel) * oversample**2
         psf_images[i, :, :] = im
 


schlafly:
In case anyone's interested, I've placed a PDF showing the PSF kernels and the corresponding matching kernels for all of the filter combinations at:
/grp/roman/eschlafly/misc/psf-matching-kernels.pdf

Here's an example:
image
The two leftmost panels show the two PSFs; the reference PSF and the target PSF.  The third panel shows the reference PSF after convolution with the matching kernel; it should look more like the target PSF than the reference PSF did.  The 4th panel shows the straight difference between the two PSFs, and the 5th panel shows the difference between the convolved reference and the target PSFs, which should be zero if we did everything right.  The final panel shows the convolution kernel.

I didn't originally expect the convolution kernels to be ~airy-function like.  But I think I should have expected this in retrospect.  The Roman PSFs are vaguely Airy-like in that they are very band-limited.  The matching kernel to go from one Airy function to another Airy function is the Airy function of the broader PSF.  So I should have expected this to work out.

ddavis-stsci (approved):

PR #2128, opened 2026-01-07
RCAL-1272: Consolidate Step input handling to handle data migration

### LINKED ISSUE #1313: Refactor input handling to avoid shallow copies
_Issue [RCAL-875](https://jira.stsci.edu/browse/RCAL-875) was created on JIRA by [Brett Graham](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bgraham):_

Related to the resample step input handling described in #1309 a more general reorganization of input handling could yield performance improvements and a more consistent step behavior for romancal.

Most steps being with a call to `datamodels.open`. For an input filename this will open the file (and return a datamodel). However for an already open datamodel this will create a shallow copy. However, the general convention in romancal is that steps modify their inputs.

The following changes could be considered:
 * change the `datamodels.open` call at the start of each step to `self.open_model`
 * update `Step._datamodels_open` to return the same model when called with an open model (this has the added benefit of allowing stpipe to open the input file fewer times and create fewer shallow copies)
 * document and test the established convention that steps modify inputs (with a few exceptions where the input model type is changed)

### LINKED ISSUE #2075: romancal backwards compatibility; processing out-of-date files
B22 will be our "commissioning" build.  We should begin targeting backwards compatibility ~now.  I think the first step there is to add regression tests that use B20 versions of files, and we should aim to let romancal continue processing those files (and at least be loudly notified when they no longer process so that we don't break backward compatibility if there's an easy workaround).

We should try to maintain backward compatibility for three data products:
- L1 files: input to roman_elp
- L2 files: input to roman_mos and source catalog
- L3 files: input to source catalog & multiband catalog

We should add regression tests that show that those steps run on old versions of the L1, L2, and L3 files.  It will be especially important for us to be able to run on old L1 files.  Running on L2 and L3 files is also valuable.

Harry is preparing another round of L2 changes that may mean this immediately goes out the window for the L2 and maybe L3 files, but let's set up the tests so that we can see that and decide how ugly the workarounds to handle it would be.

@stscijgbot-rstdms

braingram:
Consolidate/standardize `Step` input processing by adding a new `open_dataset` function and using it to:
- simplify steps (and step testing)
- provide a more consistent interface for steps
- ease implementation for migrating old files to new versions (tags)

A new parameter is added to the spec of the base romancal `Step`, `update_version` which when enabled will trigger `open_dataset` to convert old models (with old tags) to new models (this needs more testing).

The new parameter allows the addition of a few new regtests (in this PR) that test:
- L1 file through exposure level pipeline
- L2 file through mosaic pipeline
- L2 file through source catalog step
- L3 file through source catalog step
- L3 file through multiband catalog step

Without enabling the new `update_version` setting the catalog steps all fail when run with build 20 data.

Link to docs section describing this new option: https://roman-pipeline--2128.org.readthedocs.build/en/2128/roman/pipeline_parameters.html#update-version

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20729582479 show only the `wfi18_transient` failures seen on main

Closes: #1313
Closes: #2075
Closes: https://jira.stsci.edu/browse/RCAL-1272
Closes: https://jira.stsci.edu/browse/RCAL-875

## Tasks

  - ``changes/<PR#>.docs.rst``

  ## steps
  - ``changes/<PR#>.flux.rst``
</details>


romancal/outlier_detection/outlier_detection_step.py
```diff
@@ -44,16 +44,18 @@ class OutlierDetectionStep(RomanStep):
         resample_data = boolean(default=True) # Specifies whether or not to resample the input images when performing outlier detection
         resample_on_skycell = boolean(default=True) # if association contains skycell information use the skycell wcs for resampling
         good_bits = string(default="~DO_NOT_USE+NON_SCIENCE")  # DQ bit value to be considered 'good'
-        in_memory = boolean(default=False) # Specifies whether or not to keep all intermediate products and datamodels in memory
+        in_memory = boolean(default=True) # Specifies whether or not to keep all intermediate products and datamodels in memory, ignored if run as part of a pipeline
```
braingram: This parameter was previously unused. It is now used where:
- when run as a stand-alone step, enabling `in_memory=False` will cause the `ModelLibrary` and the median calculation to use temporary files
- when run as part of the mosaic pipeline (where the step is passed a `ModelLibrary` instance) the median calculation will use temporary files only if the `ModelLibrary` is using temporary files (which is not the default)

This matches the current behavior in jwst.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -159,7 +153,7 @@ def process(self, input):
         # or a DataModel (for non-asn non-lib inputs)
         with lib:
             model = lib.borrow(0)
-            lib.shelve(model)
+            lib.shelve(model, modify=False)
```
braingram: Avoids a temporary file write if the pipeline is run with an `on_disk` `ModelLibrary`.

romancal/skymatch/skymatch_step.py
```diff
@@ -53,13 +53,10 @@ class SkyMatchStep(RomanStep):
 
     reference_file_types: ClassVar = []
 
-    def process(self, input):
-        log.setLevel(logging.DEBUG)
```
braingram: Removes a stray change to the logging level.

romancal/regtest/compatibility/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture(scope="module")
+def old_rtdata_module(rtdata_module):
+    rtdata_module._env = "build/26Q1_B20"
```
braingram: Note that this ignores the artifactory environment passed to the test run (the default is `dev`). Some discussion is worthwhile about how failures of these compatibility tests will be handled but perhaps that is best handled when we have a failure. We could wait until the first compatibility issue and then xfail the test until a plan for addressing the failure is decided on.

schlafly: I like this design.  Do I follow correctly that you followed my suggestion and just have these tests make sure that processing proceeds without worrying about whether the results exactly duplicate some past results?  Certainly in terms of test maintenance it looks to me like your approach will be good there!  I'm okay with losing information about every time file content changes in out-of-date versions, but I want to make sure that doesn't sound crazy to you.

schlafly (approved):
This looks good to me, thank you!  Anything you want to do before I merge?

PR #2131, opened 2026-01-13
Update L2 schema support for B21 changes.

schlafly:
This PR has the romancal changes needed to support the new L2 schema introduced in
https://github.com/spacetelescope/rad/pull/795
The big things there are:
- new dumo / chisq fields.  These are currently just added as all zeros in an ad-hoc way to the support the schema changes; we'll be doing the actual population of these fields in a later PR.
- removing the systematic uncertainty in the dark from the dark ramp fit uncertainty; this is a largely unrelated change requested by RTB that I bundled since we need it and I was touching the variances anyway.
- changing the variances to be float16 and removing var_rnoise.  With the removal of var_dark, var_rnoise is always sqrt(err ** 2 - var_poisson), so we can reproduce var_rnoise.  To save space, this has now been removed.
- Added a function to compute var_rnoise.  Optionally, one can turn on var_flat, which would add another contribution to the total noise, so this function subtracts whatever variances are present.  Former users of var_rnoise have been migrated to use this function.
- We allow users to specify that they want to keep var_rnoise in the files, in which case it is tabulated and directly returned by compute_var_rnoise.
- Tests and the ramp fitting step were updated to accommodate the float16 dtype changes.

## Tasks
  - [X] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job (https://github.com/spacetelescope/RegressionTests/actions/runs/21005430023)


schlafly:
Regression tests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/21005430023
These show the expected differences---the new files include the new fields; the old files do not.  Pleasantly, no tests fail due to mismatched models / validation issues; romancal works with either the new or old files.

@braingram , I haven't added any migration code, but I could imagine doing that to change the types of fields or to add empty dumo / chisq fields.  But because these aren't causing problems my tendency would be to not do that; would you agree?

schlafly:
With some updates to the mosaic pipeline (thanks Brett) I can now generate a new set of reference files.
https://github.com/spacetelescope/RegressionTests/actions/runs/21035439463/job/60504474354
I guess I'll run the regression tests against them and I think that will be adequate testing on top of the usual test suite?

romancal/dark_current/dark_current_step.py
```diff
@@ -46,13 +45,11 @@ def process(self, dataset):
...
         # Open dark model
         with rdm.open(self.dark_name) as dark_model:
-            # get the dark slope and dark slope error from the reference file & trim ref pixels
+            # get the dark slope from the reference file & trim ref pixels
             dark_slope = dark_model.dark_slope[4:-4, 4:-4]
-            dark_slope_err = dark_model.dark_slope_error[4:-4, 4:-4]
 
             # Do the dark correction
             input_model.data -= dark_slope
-            input_model.err = np.sqrt(input_model.err**2 + (dark_slope_err) ** 2)
```
braingram: I don't see mention of a change to the dark step in the description or a changelog fragment. Is there a ticket that describes this change?

braingram (commented):
Overall LGTM. I left minor requests mostly about docs.

How would you like to handle the missing wfi18_transient truth file?

romancal/ramp_fitting/ramp_fit_step.py
```diff
@@ -37,11 +37,11 @@ class RampFitStep(RomanStep):
 
     spec = """
         algorithm = option('ols_cas22', 'likely', default='ols_cas22')  # Algorithm to use to fit.
-        save_opt = boolean(default=False) # Save optional output
         suffix = string(default='rampfit')  # Default suffix of results
         use_ramp_jump_detection = boolean(default=True) # Use jump detection during ramp fitting
         threshold_intercept = float(default=None) # Override the intercept parameter for the threshold function in the jump detection algorithm.
         threshold_constant = float(default=None) # Override the constant parameter for the threshold function in the jump detection algorithm.
+        include_var_rnoise = boolean(default=False) # include var_rnoise in output (can be reconstructed from err and other variances)
```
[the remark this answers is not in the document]
schlafly: Done, thanks.

romancal/dark_current/dark_current_step.py
schlafly: I've added the change log fragment and removed the problematic section from the dark docs.

schlafly:
Okay, I think these 2 PRs pull it together finally, sorry for the delay:
- against newly generated regression files, everything passes: https://github.com/spacetelescope/RegressionTests/actions/runs/21079574837
- against the current regression files, some things need to be okified: https://github.com/spacetelescope/RegressionTests/actions/runs/21079585614
I added a new wfi18_transient file to artifactory so that the linearity files could be run against the appropriate step.

braingram (approved):

PR #2133, opened 2026-01-14
Update source injection tests to reflect romanisim abandoning astropy units

schlafly:
The source injection test relied on romanisim's gain parameter having units.  This was removed in romanisim main, so that now the gain parameter has no units.  This PR removes the '.value' attribute access that relies on gain being a unit.

This only updates a test and doesn't have implications on regtests or need a changelog entry.

This makes regtests pass.  https://github.com/spacetelescope/RegressionTests/actions/runs/20997706259


mairanteodoro (approved):
Looks good to me!
Thanks, Eddie!

stscieisenhamer (approved):

PR #2136, opened 2026-01-15
RCAL-1311 Add parameters stepsize & order to outlier_detection and resample steps for updated gwcs_blot api

stscieisenhamer:
Resolves [RCAL-1311](https://jira.stsci.edu/browse/RCAL-1311)

Depends on [PR stcal#488](https://github.com/spacetelescope/stcal/pull/488)
Triggers [PR rad#806](https://github.com/spacetelescope/rad/pull/806)

This PR adds the new parameters `stepsize` and `order` to the outlier detection and resample steps. These are ultimately passed to the function `gwcs_blot`. These changes represent an efficiency gain in resampling and blotting, allowing the WCS matrix over an image to be calculated in a coarser fashion. See the `stcal` PR for more details.

## Tasks
- [ ] reset stcal dependency to stcal#main

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,8 +33,7 @@ dependencies = [
     "crds >=13.0.2",
     "drizzle >=2.2.0,<2.3.0",
     "gwcs >=1.0.1,<2",
-    #    "stcal @ git+https://github.com/spacetelescope/stcal.git",
-    "stcal @ git+https://github.com/t-brandt/stcal.git@resample_speedup",
+    "stcal @ git+https://github.com/spacetelescope/stcal.git",
     "stpipe >=0.11.0,<0.12.0",
     "spherical-geometry >=1.3.3,<1.4",
 ]


romancal/resample/resample.py
```diff
@@ -248,6 +268,10 @@ def finalize(self):
         output_model.data = self.output_model["data"]
         output_model.weight = self.output_model["wht"]
 
+        # WCS interpolation
+        output_model.meta.resample.stepsize = self.stepsize
+        output_model.meta.resample.order = self.order
```
braingram: Is there a RAD ticket to add these to the schema? If not would you open one?

schlafly: This is just so far in the weeds that saying a file that doesn't have these isn't valid is crazy to me.  I think it's reasonable for us to populate them (it's easy and someone could use them), but making it a contract that files must have this and that the archive database needs to know about them, etc., is a bridge too far for me.

braingram: WRT archive, I agree that I don't see any reason to add these to the archive database.

I also don't feel strongly that they should be required. The other resample parameters were added and required in https://github.com/spacetelescope/rad/pull/241 Following the breadcrumbs the initial discussion occurred on https://jira.stsci.edu/browse/RCAL-526 but I see no mention of making the resample metadata items required.

I do think adding descriptions of this new metadata to the schemas would be helpful. As the other items are required (like `resample.good_bits`) it seems reasonable to me at this point to make the new items required as well. This would catch bugs where `good_bits` is defined but `order` is not.

I don't have any objection to making them optional but I think we should add unit tests to check that the metadata is recorded since that's the expectation and the schemas aren't being used to check this.

PR #2145, opened 2026-01-20
[SCSB-234] use Python 3.13 to build docs with Sphinx 9

zacharyburnett:
Resolves [SCSB-234](https://jira.stsci.edu/browse/SCSB-234)


schlafly (approved):
Unit tests are failing due to VO flakiness; approving.

PR #2146, opened 2026-01-21
[SCSB-234] remove sphinx from RTD Conda environment

zacharyburnett:
so that `pip install .[docs]` can pick up Sphinx 9 from PyPI


braingram:
I didn't see this draft until now. https://github.com/spacetelescope/romancal/pull/2146 includes this change (and other fixes to get the docs to work on sphinx 9).

PR #2157, opened 2026-01-29
Integral non-linearity support

schlafly:
This PR adds an implementation of integral nonlinearity to romancal and associated tests and documentation.  The key elements are loading the integral nonlinearity file and creating a callable that implements the correction, which gets sent to the new nonlinearity implementation in stcal.  An associated rad PR needs to update ref_file and cal_step so that these new files are referenced in the Roman metadata.

I added a second linearity unit test.  As part of that work I factored out the common set-up code into a conftest.py in linearity/tests.

I updated the documentation to include references to the new integral nonlinearity and inverse linearity reference files and their use in the linearity correction.

## Tasks


rgcosentino (approved):
@t-brandt asked me to take a look at this PR.

One thing I noticed that might be missing is the treatment of data types within applying the correction. If the lookup values are UNIT16 and the correction arrays float32, so what are the input and output data types that are most appropriate here? At this point in the pipeline I think the data in DN is already cast to float32 so does upcasting the INL correction from a UNIT16 value to a float matter would be my question.

Otherwise everything else looks pretty good.

schlafly:
Thanks Rick.  I didn't think about this.  I've now checked that:
np.interp(float, uint16, float) returns a float
which I think is the only thing that matters here?  i.e., you can interpolate from a function tabulated over integers and get floats out.  numpy will also do the casting from ints to floats if you had tabulated the corrections as integers, for what it's worth, though of course you didn't do that.

docs/roman/linearity/description.rst
```diff
@@ -29,6 +43,18 @@ coefficients, and :math:`F_\text{c}` is the corrected counts. There is no
...
+has its own correction table that maps observed DN values to a correction
+offset. See `Brandt & Perera (2025)
+<https://www.stsci.edu/files/live/sites/www/files/home/roman/documentation/technical-documentation/_documents/Roman-STScI-000866.pdf>`_
+for details.
+
+The INL correction is optional; if no integral nonlinearity reference file is
+available, only the classical polynomial correction is applied.
+
 Upon successful completion of the linearity correction, "cal_step" in the
 metadata is set to "COMPLETE".
```
ddavis-stsci: Is this set to "COMPLETE" if both CNL and INL runs?

docs/roman/linearity/description.rst
```diff
@@ -29,6 +43,18 @@ coefficients, and :math:`F_\text{c}` is the corrected counts. There is no
...
 the reference file will be applied.
 
+Integral Nonlinearity Correction
+++++++++++++++++++++++++++++++++
+The integral nonlinearity (INL) correction addresses nonlinearity in the
+analog-to-digital converter. Each of the 32 readout channels (128 columns each)
+has its own correction table that maps observed DN values to a correction
+offset. See `Brandt & Perera (2025)
+<https://www.stsci.edu/files/live/sites/www/files/home/roman/documentation/technical-documentation/_documents/Roman-STScI-000866.pdf>`_
+for details.
```
ddavis-stsci: When I try this page, https://www.stsci.edu/files/live/sites/www/files/home/roman/documentation/technical-documentation/_documents/Roman-STScI-000866.pdf, I'm getting
`https://www.stsci.edu/page-not-found`
Is this not in place yet?

ddavis-stsci (approved):
I think the code looks fine. A few minor comments but nothing to hold up the process.

docs/roman/linearity/description.rst
schlafly: I've moved this bit higher to make clear that it refers to the whole step and not just the INL part.

PR #2158, opened 2026-01-31
Include chi squared, diff with uniform weight ramps

t-brandt:
Resolves [RCAL-1315](https://jira.stsci.edu/browse/RCAL-1315)

Closes #

This PR returns the chi squared computed by likelihood-based ramp fitting (if available) and writes it to the output data model.  It also computes the uniform-weighted slope and saves the difference between this and the optimal-weighted slope.  Finally, the order of the gain scaling in ols_cas22 is adjusted so that the input model is never modified and the arrays as they are passed to the output constructor are in the same units as for likelihood-based ramp fitting.

Requires https://github.com/spacetelescope/stcal/pull/508

## Tasks


schlafly:
Regression tests show some tiny changes in the var_poisson due to the slightly different way in which the gain is handled.
https://github.com/spacetelescope/RegressionTests/actions/runs/21600919190
I will merge & okify after the stcal PR is merged.

schlafly (approved):
This looks good.  The slightly different handling of the gain means that all of the values change very slightly and the regtests will need updates; I'll handle that after merging.

PR #2160, opened 2026-02-02
RCAL-1292 Initialize snowball detection parameters to use in the maximum likelyhood ramp fitting algorithm

stscieisenhamer:
Resolves [RCAL-1292](https://jira.stsci.edu/browse/RCAL-1292)

This PR initializes the relevant snowball detection parameters used by the ramp fitting maximum likelihood algorithm.

## Tasks
      - [x] **after the reviewer has approved these changes**, run `okify_regtests` to update the truth files

### DIFF
--- a/romancal/ramp_fitting/tests/common.py
+++ b/romancal/ramp_fitting/tests/common.py
@@ -36,6 +36,7 @@ SIMPLE_RESULTANTS = np.array(
 # Utilities
 # #########
 
+
 def create_linear_ramp(n_resultants=4, nrows=2, ncols=2, rate=2, pedistal=1):
     """Create linear resultants"""
     r1 = np.array(list(range(n_resultants)), dtype=np.float32) * rate + pedistal
--- a/romancal/ramp_fitting/tests/common.py
+++ b/romancal/ramp_fitting/tests/common.py
@@ -35,6 +35,18 @@ SIMPLE_RESULTANTS = np.array(
 # #########
 # Utilities
 # #########
+
+def create_linear_ramp(n_resultants=4, nrows=2, ncols=2, rate=2, pedistal=1):
+    """Create linear resultants"""
+    r1 = np.array(list(range(n_resultants)), dtype=np.float32) * rate + pedistal
+    r2 = r1[:, np.newaxis]
+    r2 = np.repeat(r2, repeats=nrows, axis=1)
+    ramps = r2[:, np.newaxis]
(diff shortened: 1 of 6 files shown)


stscieisenhamer:
Notifying @t-brandt because he cannot be added as reviewer.

stscieisenhamer:
Notes: 
- The JumpData initialization has been limited solely to the snowball detection parameters as originally indicated by @t-brandt. There are a significant number of other parameters available but have not been included. Review of the parameters should be done to check for relevancy.
- Though there are `stcal` tests for snowball detection, having a `romancal` test would probably be good. Suggestions on how to best implement such a test will be needed. Otherwise, I will examine `jwst` tests to see what can be used.

schlafly (commented):
Following up on our discussion during the tag up:
- Let's let Tim review the options later.  These are fine for now.
- Fine to leave expand_large_events default off for now.
- Let's add at least one test that makes the expand_large_events thing do something.  That probably means making a test image with some saturated pixels (a large contiguous region of saturated pixels ones depending on the options) and running the step with & without expand_large_events, and showing that with expand_large_events on we flag more pixels.

stscieisenhamer:
WRT Regressions #3:

The new snowball-specific test `romancal/ramp_fitting/tests/test_ramp_fit_likelihood.py::test_flag_large_events_withsnowball` succeeds when run locally on an M4 architecture. However, the GA version, rather significantly, fails:
```
        # Check that the uncertainties are the same for all pixels with jumps
        # and save this average uncertainty
        meanerr_jumppixels_new = np.mean(m_image.err[m_image.dq == pixel.JUMP_DET])
>       assert np.std(m_image.err[m_image.dq == pixel.JUMP_DET]) < 1e-6
E       assert np.float16(0.000546) < 1e-06
```
Unsure of what to make of that, except that this maybe what has been seen in the past where minor precision changes upstream cause dramatic changes in the ramp fitting.

romancal/ramp_fitting/tests/common.py
```diff
@@ -35,6 +35,19 @@
 # #########
 # Utilities
 # #########
+
+
+def create_linear_ramp(n_resultants=4, nrows=2, ncols=2, rate=2, pedistal=1):
```
schlafly: pedistal -> pedestal

t-brandt:
If I force the stcal install to point to my PR I get a passing test where I get failure on stcal main.

schlafly (approved):

PR #2184, opened 2026-02-12
Add romanisim package to devdeps

schlafly:
We don't currently test against romanisim main in our devdeps, but we probably should?  This PR adds romanisim to the list of devdeps we test nightly against.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,7 @@
 # Roman upstream packages
 git+https://github.com/spacetelescope/roman_datamodels.git
 git+https://github.com/spacetelescope/rad.git
+git+https://github.com/spacetelescope/romanisim.git
 
 # shared upstream packages
 git+https://github.com/spacetelescope/stcal.git


braingram (approved):
LGTM. This will require updating https://github.com/spacetelescope/romancal/pull/1914 Any opinions on that PR?

zacharyburnett (approved):

schlafly:
I was trying to come back and clean up some of these old PRs and took another look at this one.  @zacharyburnett , if you get a chance, could you see if the devdeps errors make sense to you?

zacharyburnett:
So the problem here is we're requesting `rad@main`, however `roman_datamodels@main` currently constrains `rad>=1.1.0,<1.2.0`, which are released versions on PyPI:
```
  × No solution found when resolving dependencies:
  ╰─▶ Because only roman-datamodels==1.2.0.dev5+g224241f3a is available and
      roman-datamodels==1.2.0.dev5+g224241f3a depends on rad>=1.1.0,<1.2.0,
      we can conclude that all versions of roman-datamodels depend on
      rad>=1.1.0,<1.2.0.
```

I'm unsure how to make this work with the current version of `tox`; my go to solution would have been just doing a `pip install -r requirements-dev.txt` in `commands_pre`, which occurs after the install step so would override everything without complaint; HOWEVER, testing has revealed that even running `pip` in `commands_pre` doesn't work anymore:
```
py3-devdeps-xdist: commands_pre[1]> pip install -r requirements-dev.txt
py3-devdeps-xdist: Exception running subprocess [Errno 2] No such file or directory: 'pip'
```

This also means that now our existing `oldestdeps` toxenv config won't work anymore either:
```
py3-oldestdeps-xdist: commands_pre[1]> pip install -r requirements-min.txt
py3-oldestdeps-xdist: Exception running subprocess [Errno 2] No such file or directory: 'pip'
```

zacharyburnett:
I'll see if an older version of `tox` restores this functionality. In the meantime, I'll also explore how much work it might be to convert our `tox.ini` to a `noxfile.py` and use Nox instead (it's reportedly designed to be easy to migrate from Tox)

braingram (changes requested):
Looks like it's not pulling in main from the version (0.14.0):
https://github.com/spacetelescope/romancal/actions/runs/35661144536/job/106536484534?pr=2184#step:11:116
@schlafly the pin in pyproject.toml doesn't allow 0.15.0:
https://github.com/spacetelescope/romancal/blob/15cb2e3d11558ac415111863aca63000d7a92118/pyproject.toml#L30
which appears to be the latest release:
https://github.com/spacetelescope/romanisim/releases/tag/v0.15.0
Does something need updating?

schlafly:
Thanks, I'm trying to loosen the pin here: https://github.com/spacetelescope/romancal/pull/2463

schlafly:
This finally looks good.  It was useful; I had broken romancal in https://github.com/spacetelescope/romanisim/pull/379, so I fixed romanisim in https://github.com/spacetelescope/romanisim/pull/396, fixing the issue.  That had the side effect of verifying that this PR is now functional.  I'm ready to merge this.

braingram (approved):
Success!
https://github.com/spacetelescope/romancal/actions/runs/35886019020/job/107351904787?pr=2184#step:11:116

PR #2224, opened 2026-03-10
Reduce ExposurePipeline peak memory by ~46% for high-read-count data

adamksmith:
## Summary

- **Memory-efficient linearity correction**: Processes reads one at a time instead of materializing all reads per resultant simultaneously, reducing linearity peak from ~6GB temporaries to ~200MB for 32-read resultants
- **Chunked ramp fitting**: Adds `maximum_memory` parameter to `RampFitStep` for optional spatial chunking via `_ChunkedRampData`/`_RefDataSlice` proxy classes
- **In-place flat field operations**: Uses `np.square(out=)` and in-place division in `flat_field.py` to eliminate temporary arrays
- **Pipeline plumbing**: Adds `on_disk` parameter to `ExposurePipeline` spec (matching `MosaicPipeline` pattern), fixes `fileio.py` to strip `on_disk` kwarg for single ASDF files, adds `gc.collect()` before memory-intensive steps

## Test plan

- [ ] Verify linearity correction produces identical results to stcal's `linearity_correction` for multi-resultant data
- [ ] Verify single-resultant edge case falls back to simple polynomial linearization
- [ ] Run ExposurePipeline on high-read-count data (e.g., MA table SCI0005 with 63 reads) and confirm reduced peak memory
- [ ] Test chunked ramp fitting with `maximum_memory` parameter and verify stitched output matches full-frame processing
- [ ] Confirm flat field in-place operations produce identical results to previous implementation
- [ ] Run existing regression tests to ensure no numerical changes

### DIFF
--- a/romancal/datamodels/fileio.py
+++ b/romancal/datamodels/fileio.py
@@ -78,7 +78,9 @@ def open_dataset(
             result = ModelLibrary(dataset, update_version=update_version, **open_kwargs)
 
         case "asdf":
-            model = rdm.open(dataset, **open_kwargs)
+            # on_disk is a ModelLibrary concept; strip it before opening a single file
+            asdf_kwargs = {k: v for k, v in open_kwargs.items() if k != "on_disk"}
+            model = rdm.open(dataset, **asdf_kwargs)
             if update_version:
                 result = update_model_version(model, close_on_update=True)
             else:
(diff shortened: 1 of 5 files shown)


schlafly:
This PR has some different things.  The linearity updates made here inspired updates to stcal (https://github.com/spacetelescope/stcal/pull/531), which was the tall pole.  We've also made a few related memory updates in the interim.  We didn't adopt some of the in-place operations or the on-disk mode in favor of readability and not changing I/O patterns (i.e., on_disk).  Closing this one.

Issue #1994, opened 2025-10-07
Add infrastructure for parquet testing to regtests and use for relevant files


braingram:
Currently parquet files are generated for catalogs. However the contents of these files are not generally read back in and compared against a "truth" as is common practice for the ASDF products (and regtests).

Comparing the table contents might be noisy as small input differences could result in a different number of rows that would be difficult and largely not informative to compare. However the metadata and general table structure (number of columns, column types, etc) could be more easily and more informatively compared.

This ticket is to discuss what portions of the output parquet files to target for comparison and how to perform those comparisons.

stscijgbot-rstdms:
_This issue is tracked on JIRA as [RCAL-1241](https://jira.stsci.edu/browse/RCAL-1241)._

PR #2237, opened 2026-03-19
Add `compare_parquet` to regtestdata and use it to compare output to truth for parquet files

braingram:
Resolves https://jira.stsci.edu/browse/RCAL-1241
See: https://github.com/spacetelescope/romancal/issues/1994

Add `compare_parquet` (similar to `compare_asdf`) for comparing parquet files generated by regression tests with "truth" parquet files.

At the moment this compares both the table and metadata (but as discussed in the ticket we might want to disable the table comparison). I left it in since:
- it's essentially "free" since we already have [table comparison code](https://github.com/spacetelescope/romancal/blob/0bec7d524678a60641e24872e68d5fb8e987d5ae/romancal/regtest/regtestdata.py#L427)
- it helps to illustrate some differences (which we currently could okify)

`compare_asdf` works by:
- reading the parquet files with pyarrow
- converting the metadata back into a tree structure (to allow us to use the standard "ignores") but with all values as strings (a limitation of the parquet format)
- generating trees to compare containing the metadata and the table stored under a "table" key
- reuse the existing tree comparison code used by `compare_asdf` (refactored a bit to make reuse easier)

EDIT: the previous run is old enough the logs are now missing.
New regtest run here: https://github.com/spacetelescope/RegressionTests/actions/runs/29272068407
shows several errors where the new output contains dust_ebv and the truth files do not. This is something that should be okified.

## Tasks

### DIFF
--- a/romancal/regtest/conftest.py
+++ b/romancal/regtest/conftest.py
@@ -209,3 +209,8 @@ def rtdata_module(artifactory_repos, envopt, request, module_jail):
 @pytest.fixture
 def ignore_asdf_paths(ignore_metadata_paths):
     return {"ignore": ignore_metadata_paths}
+
+
+@pytest.fixture
+def ignore_parquet_paths(ignore_parquet_metadata_paths):
+    return {"ignore": ignore_parquet_metadata_paths}
--- a/romancal/regtest/regtestdata.py
+++ b/romancal/regtest/regtestdata.py
@@ -533,10 +533,12 @@ class DiffResult:
         return f"{title}\n{report}"
 
 
-def _compare_trees(
-    result, truth, exclude_paths=None, rtol=1e-05, atol=1e-08, equal_nan=True
-):
-    exclude_paths = exclude_paths or []
+def _compare_trees(result, truth, ignore=None, rtol=1e-05, atol=1e-08, equal_nan=True):
+    exclude_paths = []
+    if ignore:
+        for path in ignore:
+            key_path = "".join([f"['{k}']" for k in path.split(".")])
+            exclude_paths.append(f"root{key_path}")
     operators = [
         NDArrayTypeOperator(
             rtol,
@@ -595,11 +597,6 @@ def compare_asdf(result, truth, ignore=None, rtol=1e-05, atol=1e-08, equal_nan=T
     diff_result : DiffResult
         result of the comparison
     """
-    exclude_paths = []
-    if ignore:
-        for path in ignore:
-            key_path = "".join([f"['{k}']" for k in path.split(".")])
-            exclude_paths.append(f"root{key_path}")
     with asdf.open(result) as af0:
         with asdf.config_context() as cfg:
             # Disable validation of the truth file to allow for more
@@ -611,7 +608,7 @@ def compare_asdf(result, truth, ignore=None, rtol=1e-05, atol=1e-08, equal_nan=T
                 diff = _compare_trees(
                     af0.tree,
                     af1.tree,
-                    exclude_paths=exclude_paths,
+                    ignore=ignore,
                     rtol=rtol,
                     atol=atol,
                     equal_nan=equal_nan,
@@ -621,27 +618,25 @@ def compare_asdf(result, truth, ignore=None, rtol=1e-05, atol=1e-08, equal_nan=T
 
 def _parquet_to_tree(parquet_file):
     table = pyarrow.parquet.read_table(parquet_file)
-    return {
-        "table": astropy.table.Table.from_pandas(table.to_pandas()),
-        "meta": {
-            k.decode("ascii"): v.decode("ascii")
-            for k, v in table.schema.metadata.items()
-        },
-    }
+    # convert meta back into a tree
+    result = {}
+    for k, v in table.schema.metadata.items():
+        path = k.decode("ascii")
+        value = v.decode("ascii")
+        *subpaths, key = path.split(".")
+        obj = result
+        for subpath in subpaths:
+            obj = obj.setdefault(subpath, {})
+        obj[key] = value
+    result["table"] = astropy.table.Table.from_pandas(table.to_pandas())
+    return result
 
 
 def compare_parquet(result, truth, ignore=None, rtol=1e-05, atol=1e-08, equal_nan=True):
-    exclude_paths = []
-    if ignore:
-        # prepend meta since _parquet_to_tree puts all metadata under the "meta" key
-        for path in ignore:
-            exclude_paths.append(f"root['meta']['{path}']")
-    if ignore:
-        ignore = [f"meta.{key}" for key in ignore]
     diff = _compare_trees(
         _parquet_to_tree(result),
         _parquet_to_tree(truth),
-        exclude_paths=exclude_paths,
+        ignore=ignore,
         rtol=rtol,
         atol=atol,
         equal_nan=equal_nan,
--- a/romancal/regtest/regtestdata.py
+++ b/romancal/regtest/regtestdata.py
@@ -13,7 +13,7 @@ import astropy.time
 import deepdiff
 import gwcs
 import numpy as np
-import pyarrow
+import pyarrow.parquet
 import requests
 import roman_datamodels as rdm
 from astropy.units import Quantity
--- a/romancal/regtest/regtestdata.py
+++ b/romancal/regtest/regtestdata.py
@@ -13,6 +13,7 @@ import astropy.time
 import deepdiff
 import gwcs
 import numpy as np
+import pyarrow
 import requests
 import roman_datamodels as rdm
 from astropy.units import Quantity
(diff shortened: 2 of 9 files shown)


schlafly:
I recall that I stalled on this one because I remember your uncovering that the source catalog when run on the final L2 file and when created as part of the ELP and then updated were not identical, because of very tiny differences in the shapes and position angles of sources.  We correct the positions at the end of the ELP but we don't correct the galaxy angles or sizes for the tiny rotation and scaling that comes from tweakreg.  Those differences seemed 'fine' to me (i.e., close enough that one would want to address it through tolerances rather than by rearchitecting the regtests).

I don't see that issue now but maybe it's hidden by the different field list?  I can't find evidence that I communicated this concern to you before, sorry!

braingram:
Thanks for taking a look. Maybe it would be helpful for me to:
- run the regtests on a copy of the input/truth files
- okified the field differences
- run again to confirm that no differences are found

I'll get those started: https://github.com/spacetelescope/RegressionTests/actions/runs/29848076516

braingram:
I'm converting this back to draft. With the tests above after okifying there are 3 failures. I think there may be multiple tests using the same truth files but generating different results. More investigation is needed to untangle this.

braingram:
The rerun showed 3 errors:
<details>
<summary>Click to expand...</summary>

```
FAILED romancal/regtest/test_mos_pipeline.py::test_catalog_matches_truth - AssertionError: 
  Diff report for:
      result file: /home/runner/_work/_temp/pytest_basetemp/popen-gw0/test_mos_pipeline_rtdata_module0/r0000101001001001001_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
      truth file: /home/runner/_work/_temp/pytest_basetemp/popen-gw0/test_mos_pipeline_rtdata_module0/truth/r0000101001001001001_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
  
  {'tables_differ': {"root['table']": {'column_values': {'psf_flux_err': {'abs_diff': np.float32(17.095085),
                                                                          'n_diffs': np.int64(1),
                                                                          'worst_abs_diff': {'index': (np.int64(70),),
                                                                                             'value': np.float32(17.094727)},
                                                                          'worst_fractional_diff': {'index': (np.int64(16),),
                                                                                                    'value': np.float32(1.0000001)}},
                                                         'x_psf_err': {'abs_diff': np.float32(0.16383362),
                                                                       'n_diffs': np.int64(1),
                                                                       'worst_abs_diff': {'index': (np.int64(70),),
                                                                                          'value': np.float32(0.16383362)},
                                                                       'worst_fractional_diff': {'index': (np.int64(0),),
                                                                                                 'value': np.float32(1.0000001)}},
                                                         'y_psf_err': {'abs_diff': np.float32(0.09316635),
                                                                       'n_diffs': np.int64(1),
                                                                       'worst_abs_diff': {'index': (np.int64(70),),
                                                                                          'value': np.float32(0.09316635)},
                                                                       'worst_fractional_diff': {'index': (np.int64(0),),
                                                                                                 'value': np.float32(1.0000001)}}}}},
   'values_changed': {"root['roman']['meta']['resample']['good_bits']": {'new_value': '~DO_NOT_USE',
                                                                         'old_value': '~DO_NOT_USE+NON_SCIENCE'}}}
assert False
 +  where False = <romancal.regtest.regtestdata.DiffResult object at 0x72330927f800>.identical
FAILED romancal/regtest/test_wfi_image_pipeline.py::test_catalog_matches_truth - AssertionError: 
  Diff report for:
      result file: /home/runner/_work/_temp/pytest_basetemp/popen-gw3/test_wfi_image_pipeline_rtdata_module0/r0000101001001001001_0001_wfi01_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
      truth file: /home/runner/_work/_temp/pytest_basetemp/popen-gw3/test_wfi_image_pipeline_rtdata_module0/truth/r0000101001001001001_0001_wfi01_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
  
  {'tables_differ': {"root['table']": {'shapes': [(127,), (120,)]}},
   'values_changed': {"root['roman']['meta']['wcsinfo']['s_region']": {'new_value': 'POLYGON '
                                                                                    'ICRS  '
                                                                                    '269.986945913 '
                                                                                    '65.974265020 '
                                                                                    '269.986740071 '
                                                                                    '66.097228517 '
                                                                                    '269.676782731 '
                                                                                    '66.097195577 '
                                                                                    '269.680114248 '
                                                                                    '65.974342004',
                                                                       'old_value': 'POLYGON '
                                                                                    'ICRS  '
                                                                                    '269.986946008 '
                                                                                    '65.974264996 '
                                                                                    '269.986740349 '
                                                                                    '66.097228493 '
                                                                                    '269.676783009 '
                                                                                    '66.097195629 '
                                                                                    '269.680114343 '
                                                                                    '65.974342056'}}}
assert False
 +  where False = <romancal.regtest.regtestdata.DiffResult object at 0x7b03f326f4d0>.identical
FAILED romancal/regtest/test_mos_skycell_pipeline.py::test_catalog_matches_truth - AssertionError: 
  Diff report for:
      result file: /home/runner/_work/_temp/pytest_basetemp/popen-gw2/test_mos_skycell_pipeline_rtdata_module0/r00001_p_v01001001001001_270p65x70y49_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
      truth file: /home/runner/_work/_temp/pytest_basetemp/popen-gw2/test_mos_skycell_pipeline_rtdata_module0/truth/r00001_p_v01001001001001_270p65x70y49_f158_cat.parquet
          model type: Not a model
          crds_context: Unknown context
  
  {'values_changed': {"root['roman']['meta']['resample']['good_bits']": {'new_value': '~DO_NOT_USE',
                                                                         'old_value': '~DO_NOT_USE+NON_SCIENCE'}}}
assert False
 +  where False = <romancal.regtest.regtestdata.DiffResult object at 0x701fda6d3100>.identical
```
</details>

- romancal/regtest/test_mos_pipeline.py::test_catalog_matches_truth: good_bits mismatch, psf columns data mismatch
- romancal/regtest/test_wfi_image_pipeline.py::test_catalog_matches_truth: s_region mismatch, different number of rows
- romancal/regtest/test_mos_skycell_pipeline.py::test_catalog_matches_truth: good bits mismatch

Taking the last one (test_mos_skycell_pipeline). The truth file is `r00001_p_v01001001001001_270p65x70y49_f158_cat.parquet` which is used in 2 places:
- test_mos_skycell_pipeline
- test_source_catalog

Since the latter runs just the source catalog on a previously resampled file (which used code prior to https://github.com/spacetelescope/romancal/pull/2315) the output file will have `NON_SCIENCE` whereas the former includes resampling so the output file will not have `NON_SCIENCE`. Options here are to either:
- regenerate the input file so it does not have `NON_SCIENCE`
- use 2 different truth files for these 2 tests

test_mos_pipeline also has the good_bits issue. However I do not understand the psf columns data mismatch.

test_wfi_image_pipeline has a related issue where the file is used for 2 tests:
- test_wfi_image_pipeline
- test_source_catalog

I think what's happening here is that the test_source_catalog input file has already had tweakreg run (as part of [make_regtestdata.sh](https://github.com/spacetelescope/romancal/blob/83971dd38bef8dc9036890dc376a50e35280b7f8/romancal/scripts/make_regtestdata.sh#L45)) whereas the S_REGION for the file generated in test_wfi_image_pipeline does not reflect tweakreg changes. I think the right option here is to use 2 different truth files for the 2 tests.

To summarize I think we need separate truth files for the test_source_catalog tests and more investigation is needed to determine why the psf columns differ.

schlafly:
I don't understand the PSF differences either for the mosaic pipeline but I would note that it's a single entry in the catalog with a relative difference of 1e-7; i.e., I'd be okay with addressing this with tolerances.

This one is the worst:
```
romancal/regtest/test_wfi_image_pipeline.py::test_catalog_matches_truth
 'tables_differ': {"root['table']": {'shapes': [(127,), (120,)]}
```
which I haven't understood.  But ultimately I agree that we'd be better off not reusing test files.

braingram:
> I don't understand the PSF differences either for the mosaic pipeline but I would note that it's a single entry in the catalog with a relative difference of 1e-7; i.e., I'd be okay with addressing this with tolerances.
> 
> This one is the worst:
> 
> ```
> romancal/regtest/test_wfi_image_pipeline.py::test_catalog_matches_truth
>  'tables_differ': {"root['table']": {'shapes': [(127,), (120,)]}
> ```

Thanks for pointing that out. I overlooked that difference in the above comment (updated to include this now). I'm still stuck on the first test. Running the mos and source catalog tests locally I see even different results where the relative difference is 1e-4. What's the expected tolerance for `psf_flux_err`?

schlafly:
Scientifically if the uncertainties are different by <1% it is trouble to imagine people caring much.  For the fluxes themselves, rather than the errors, people get worried if difference / uncertainty is larger than something like 1%, with some floor on the uncertainty at the 0.1% level.

I think in general that these nonlinear iterative fits are more architecture dependent than most of the pixel-level processing is expected.

braingram:
I was able to get a successful run:
https://github.com/spacetelescope/RegressionTests/actions/runs/30017833007/job/89307738929
by:
- copying input/truth files
- creating new truth files for the source catalog tests https://github.com/spacetelescope/romancal/pull/2237/commits/2923b67b72e775f3048d1eae29d3ac7eb518abae
- playing some "tricks" to get rtdata to behave for the multiple truth/result comparisons in a single file
- okifying the differences

This PR is not ready for review for a few reasons:
- there are unexpected difference that will require inspecting files and likely separate tickets. EDIT: @schlafly pointed out that given the age of the input cal file for the source_catalog test the difference in the number of sources could be explained by the sqrt(2) fixes in https://github.com/spacetelescope/romancal/pull/2360
- the "tricks" to get rtdata to work are fragile (depend on test order) it would be good to find a better solution
- this starts on https://github.com/spacetelescope/romancal/issues/2398 (by introducing `truth/test_source_catalog/`) which hasn't been discussed

PR #2261, opened 2026-04-07
RCAL-1371: HGA Migration Support

PaulHuwe:
Resolves [RCAL-1371](https://jira.stsci.edu/browse/RCAL-1371)

Closes #2259

This PR adds migration support for HGA metadata.

## Tasks

### DIFF
--- a/romancal/datamodels/migration.py
+++ b/romancal/datamodels/migration.py
@@ -1,4 +1,4 @@
-import warnings
+import logging
 
 from astropy.time import Time
 from roman_datamodels.datamodels import ImageModel
@@ -45,10 +45,9 @@ def update_model_version(model, *, close_on_update=False):
     if isinstance(model, ImageModel) and "hga_move" not in model.meta.get(
         "exposure", "hga_move"
     ):
-        warnings.warn(
+        logging.warning(
             "Migration is adding keyword hga_move to the exposure "
-            "block and arbitrarily setting it to False.",
-            stacklevel=2,
+            "block and arbitrarily setting it to False."
         )
         updated_model.meta.exposure.hga_move = False
 
--- a/romancal/datamodels/migration.py
+++ b/romancal/datamodels/migration.py
@@ -1,4 +1,6 @@
+import logging
 from astropy.time import Time
+from roman_datamodels.datamodels import ImageModel
 
 __all__ = ["update_model_version"]
 
@@ -39,9 +41,11 @@ def update_model_version(model, *, close_on_update=False):
         updated_model.meta.observation.wfi_parallel = visit_file_sequence > 1
 
     # old files (<B21) lack HGA_move
-    if "hga_move" not in model.meta.get("exposure", "hga_move") and (
-        "effective_exposure_time" in model.meta.get("exposure", {})
+    if (isinstance(model, ImageModel) and
+        "hga_move" not in model.meta.get("exposure", "hga_move")
     ):
+        logging.warning(f"Migration is adding keyword hga_move to the exposure "
+                         "block and arbitrarily setting it to False.")
         updated_model.meta.exposure.hga_move = False
 
     # old files (<B21) may have lower case psf_match_reference_filter
--- a/romancal/datamodels/migration.py
+++ b/romancal/datamodels/migration.py
@@ -46,8 +46,9 @@ def update_model_version(model, *, close_on_update=False):
 
     # old files (<B21) may have lower case psf_match_reference_filter
     if "psf_match_reference_filter" in model.get("meta", {}):
-        updated_model.meta.psf_match_reference_filter = \
+        updated_model.meta.psf_match_reference_filter = (
             updated_model.meta.psf_match_reference_filter.upper()
+        )
 
     new_ref_files = ["darkdecaysignal", "integralnonlinearity", "inverselinearity"]
     ref_file = model.meta.get("ref_file", None)
--- a/romancal/datamodels/migration.py
+++ b/romancal/datamodels/migration.py
@@ -44,6 +44,11 @@ def update_model_version(model, *, close_on_update=False):
     ):
         updated_model.meta.exposure.hga_move = False
 
+    # old files (<B21) may have lower case psf_match_reference_filter
+    if "psf_match_reference_filter" in model.get("meta", {}):
+        updated_model.meta.psf_match_reference_filter = \
+            updated_model.meta.psf_match_reference_filter.upper()
+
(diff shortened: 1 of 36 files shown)


PaulHuwe:
Regtests with exptected validation errors: https://github.com/spacetelescope/RegressionTests/actions/runs/24089488254/

Will require an OKify run

schlafly:
The test_source_catalog, ramp_fitting, and tweakreg bits are validation errors; I don't think I expect an okify run to be able to fix them?  They'll need a new set of files.

I'm feeling dense, but I am not sure why some of these fail validation and others fail due to the new keyword.  e.g., I would have expected
test_pipeline_suffix: romancal.regtest.test_wfi_image_pipeline
to fail to produce the new file due to a validation failure, but it makes the new file without the keyword.  Meanwhile making a source catalog from a file without hga_move fails to produce the file.  Is the difference between these two cases clear to you?

PaulHuwe:
This requires recreation of regtest files. With that, it passes regression testing. URLs below:

roman_0046.pmap:
Reg files: https://github.com/spacetelescope/RegressionTests/actions/runs/24348019222
Reg tests: https://github.com/spacetelescope/RegressionTests/actions/runs/24417602382

roman-edit:
Reg files: https://github.com/spacetelescope/RegressionTests/actions/runs/24414206795
Reg tests: https://github.com/spacetelescope/RegressionTests/actions/runs/24422091104

romancal/datamodels/migration.py
```diff
@@ -30,14 +30,26 @@ def update_model_version(model, *, close_on_update=False):
...
     if ("wfi_parallel" not in model.meta.get("observation", {})) and (
         visit_file_sequence := model.meta.get("observation", {}).get(
             "visit_file_sequence", None
         )
     ):
         updated_model.meta.observation.wfi_parallel = visit_file_sequence > 1
 
+    # old files (<B21) lack HGA_move
+    if "hga_move" not in model.meta.get("exposure", "hga_move") and (
+        "effective_exposure_time" in model.meta.get("exposure", {})
```
schlafly: What's the motivation for this line?  I'm not sure what the connection is between hga_move and the effective exposure time.

PaulHuwe: Is a way to check that it is using the appropriate `exposure` meta object, as `exposure` is not unique across all the RAD schemas.

schlafly: Would it make more sense, for example, to check that the file is an L1 or L2 file?  What datamodel has meta.exposure that we wouldn't want to add hga_move to?

It just feels weird that changing the name of effective_exposure_time would change the hga_move migration behavior.

braingram: Should we be adding a migration for this? If we assume this is "real" data we wouldn't know if the high gain antenna moved or not. Setting this to `False` means we're asserting that it did not move.

schlafly: Do you have an alternative that lets these files process cleanly?  We shouldn't break backward compatibility to propagate this informational flag when present.  I'd be conceptually okay with setting it to None, but that would likely break validation / archive.

braingram: Is there any additional information on this new metadata? The linked rcal ticket:
https://jira.stsci.edu/browse/RCAL-1371
doesn't link to any information and doesn't describe the keyword.

PaulHuwe: The keyword request is here:

https://github.com/spacetelescope/rad/issues/851

braingram: Thanks. Am I reading the RAD ticket correct in assuming that the `hga_move` value is set in the L1 file and the L2 file just passes on what is in the L1?

Is there any method by which the `hga_move` value for a given exposure can be queried (the engineering database, etc)?

schlafly: Correct.  I don't know how SDP sets this in the L1s but yes, it's presumably via a query of the engineering database.  I don't have a strong sense for what mode we would be in if SDP would have originally failed to set this metadata but later the engineering database would allow it to be set.

braingram: Thanks. As we've discussed before there is an expectation that we will run into something similar in the future (where new files expect some metadata that old files do not have and filling in that metadata is either overly burdensome or impossible). This seems like one of those cases. To throw out a few options for how we address this:

- accept that this is a breaking change and we can't process older files, skip the migration tests
- update the RAD schema to allow for some undefined state (`None` as you mentioned or something else), update the migration tests to fill in the undefined state
- fill in `False` but warn that we've assumed a likely correct but possibly incorrect value. Update the tests to expect this warning for old files but not current ones.
- add new API that allows users to fill in this new value, update the migration tests to use this new API (assuming that the value is False)

PaulHuwe: My vote would be for 3. 2 has the issue that this is a database'd keyword with the type `boolean`, so `None` wouldn't be acceptable (otherwise I would have used it here).

schlafly: Let's go with 3.  We can just add a warning now; if testing for that warning interferes with getting the release out, let's punt the tests to a future ticket.  Thanks.

PaulHuwe:
I re-ran the roman-edit regression tests, and they passed: https://github.com/spacetelescope/RegressionTests/actions/runs/24422091104

schlafly (approved):
Thanks Paul.

romancal/datamodels/migration.py
```diff
@@ -30,14 +33,30 @@ def update_model_version(model, *, close_on_update=False):
...
             "visit_file_sequence", None
         )
     ):
         updated_model.meta.observation.wfi_parallel = visit_file_sequence > 1
 
+    # old files (<B21) lack HGA_move
+    if isinstance(model, ImageModel) and "hga_move" not in model.meta.get(
+        "exposure", "hga_move"
+    ):
+        logging.warning(
```
braingram, line 48: Can we make this warning and catch it in the tests?

PaulHuwe: Done

braingram:
```suggestion
        warnings.warn(
```
I'm suggesting something that's a `Warning` and not a log message.

Also this log message is going to the root logger which isn't captured (so won't end up in produced files).

PaulHuwe: Okay, updating.

PaulHuwe: After discussion, @braingram and I decided that we should move this update to a separate PR as to unblock things mid-merge.

Issue #2082, opened 2025-12-02
Improve catfile parsing and error messages


braingram:
As discussed https://github.com/spacetelescope/romancal/pull/2078#discussion_r2581450361
the current `_parse_catfile` function in the tweakreg step allows parsing files with only filenames:
https://github.com/spacetelescope/romancal/blob/c9d9845038e39cd841e73fea90e21f0f8e90f8dd/romancal/tweakreg/tweakreg_step.py#L551-L552
as seen in this test:
https://github.com/spacetelescope/romancal/blob/c9d9845038e39cd841e73fea90e21f0f8e90f8dd/romancal/tweakreg/tests/test_tweakreg.py#L943-L963

However the returned result is not supported by the rest of the step (and later causes a crash).

I suggest we drop support for parsing files of this format (which do not contain required catalogs for the listed filenames) and instead of `_parse_catfile` produce an exception for this format.

stscijirahub:
_This issue is tracked on JIRA as [RCAL-1273](https://jira.stsci.edu/browse/RCAL-1273)._

mairanteodoro:
@braingram 

Should we:

1.  drop the support for catfiles or ✅ 
2. ~adopt the same behavior as for JWST~? 🚫 

~In JWST (`jwst.tweakreg.tweakreg_step`), `_parse_catfile()` reads a plain text “catfile” and returns a `dict` mapping data model filename → catalog filename (or `None`). See code implementation [here](https://github.com/spacetelescope/jwst/blob/61bd2fe478342a086411206c6de64c8f8eccbcb3/jwst/tweakreg/tweakreg_step.py#L488-L542).~

~Key behaviors in JWST:~

- ~if catfile is `None` or empty/whitespace: returns `None`;~
- ~it skips blank lines and comments (# ...);~
- ~each non-comment line is split on whitespace:~
  - ~2 columns: `data_model catalog_file`:~
    - ~stored as `{data_model: <catfile_dir>/<catalog_file>}` (catalog path made relative to the catfile’s directory);~
  - ~1 column (data model only):~
    - ~stored as `{data_model: None}` (meaning “no custom catalog provided for this model; let tweakreg generate one / skip custom for it”);~
  - ~\>2 columns:~
    - ~raises ValueError("'catfile' can contain at most two columns.").~

schlafly:
[...]

PR #2265, opened 2026-04-10
RCAL-1238: Sync up TweakReg's docs and code behavior.

### LINKED ISSUE #1990: Tweakreg documentation describes incorrect attribute
https://roman-pipeline.readthedocs.io/en/latest/roman/tweakreg/README.html#custom-source-catalogs
mentions:
> In addition to setting the meta.tweakreg_catalog attribute of input data models to the custom catalog file name, the tweakreg_step also supports two other ways of supplying custom source catalogs to the step:

However the code checks `meta.source_catalog.tweakreg_catalog` for the filename:
https://github.com/spacetelescope/romancal/blob/54595aedfe4686ae7a35f134b9c3e949af6d86e6/romancal/tweakreg/tweakreg_step.py#L407

The code also has a FIXME:
https://github.com/spacetelescope/romancal/blob/54595aedfe4686ae7a35f134b9c3e949af6d86e6/romancal/tweakreg/tweakreg_step.py#L130-L135

which suggests that setting via an association may not work. Do we test this?

Finally, there are situations where the input file will not have a `meta.source_catalog` which makes setting `meta.source_catalog.tweakreg_catalog_name` difficult:
```python
m.meta.source_catalog.tweakreg_catalog_name = "foo.ecsv"
```
would fail with:
```
AttributeError: No such attribute (source_catalog) found in node: <class 'roman_datamodels.stnode._node.DNode'>
```
It would be helpful for the documentation to clarify this situation.

mairanteodoro:
Resolves [RCAL-1238](https://jira.stsci.edu/browse/RCAL-1238)

Closes #1990 

Issue #1990 reported that docs referenced the wrong metadata path and that ASN/member-based custom catalog behavior was unclear. This PR updates docs, preserves compatibility, and adds tests to verify expected behavior.

- Updated `TweakReg` docs to consistently reference:
  - `meta.source_catalog.tweakreg_catalog_name` (file-based catalog);
  - `meta.source_catalog.tweakreg_catalog` (in-memory catalog).
- Clarified documentation for the missing `meta.source_catalog` case, including expected failure mode and how to initialize it.
- Updated ASN member handling in `ModelLibrary` so member-level `tweakreg_catalog` is mapped into:
  - `meta.source_catalog.tweakreg_catalog_name`;
  - while preserving `meta.tweakreg_catalog` for backward compatibility.
- Removed outdated `FIXME` context in `TweakRegStep` related to ASN catalog wiring.
- Added tests in `romancal/tweakreg/tests/test_tweakreg.py` for:
  - custom catalog via ASN member attribute (`tweakreg_catalog`) being used successfully;
  - explicit `AttributeError` when `meta.source_catalog` is missing.

## Regression tests
All tests are passing:
- https://github.com/spacetelescope/RegressionTests/actions/runs/25860309174 (05/14/26)

## Tasks

## Note
The description for this PR was written using AI.


romancal/datamodels/library.py
```diff
@@ -70,7 +70,16 @@ def _assign_member_to_model(self, model, member):
...
+                # leave this for backwards compatibility, but also add it to
+                # the source_catalog as a filename for the catalog
                 model.meta[attr] = member[attr]
+                if not hasattr(
+                    model.meta,
+                    "source_catalog",
+                ):
+                    model.meta["source_catalog"] = {}
+
+                model.meta.source_catalog["tweakreg_catalog_name"] = member[attr]
```
schlafly, line 79: Would we be able to simplify some of the input handling in the tweakreg step if we always assigned to meta.source_catalog['tweakreg_catalog_name'] rather than to two locations?  This is an in-memory thing here, and the model.meta[attr] locations we probably do not actually want set in the final L2 files (since they're not in the schema?).

docs/roman/tweakreg/README.rst
```diff
@@ -7,30 +7,44 @@ Description
...
 using the corrected WCS will align on the sky.
 
 Custom Source Catalogs
 ----------------------
-The default catalog used by ``tweakreg`` step can be disabled by
-providing a file name to a custom source catalog in the
-``meta.tweakreg_catalog`` attribute of input data models.
+For the ``tweakreg`` step to use user-provided input source catalogs,
+``use_custom_catalogs`` parameter of the ``tweakreg`` step must be set to
+`True`.
```
schlafly: I'm confused here.  Isn't our normal mode of operation to run tweakreg with use_custom_catalogs = False but set source_catalog.tweakreg_catalog_name to the name of our parquet file?

schlafly (commented):
Trying to make sure I understand what the code intends to do, can you correct the following?

There are four different ways to specify a catalog.  
1. in-memory via tweakreg_catalog
2. by filename tweakreg_catalog_name
3. in the association (overwriting tweakreg_catalog_name)
4. via catfile (overwriting tweakreg_catalog_name) with use_custom_catalogs.

The catfile option only happens if use_custom_catalogs is set and takes precendence in that case.  I think the association handling happens early so 3 takes precedence over 2.  I don't know what happens if an in-memory catalog is set and also a filename or a filename in the association.

mairanteodoro:
There's an issue (#2082) and a corresponding ticket ([RCAL-1273](https://jira.stsci.edu/browse/RCAL-1273)) in which Brett makes a suggestion to drop support for catfiles. For increased visibility, this is what [I asked](https://github.com/spacetelescope/romancal/issues/2082#issuecomment-4226268782) on the issue page:

> Should we:
> 
> 1.  drop the support for catfiles or ✅ 
> 2. ~adopt the same behavior as for JWST~? 🚫 
> 
> ~In JWST (`jwst.tweakreg.tweakreg_step`), `_parse_catfile()` reads a plain text “catfile” and returns a `dict` mapping data model filename → catalog filename (or `None`). See code implementation [here](https://github.com/spacetelescope/jwst/blob/61bd2fe478342a086411206c6de64c8f8eccbcb3/jwst/tweakreg/tweakreg_step.py#L488-L542).~
> 
> ~Key behaviors in JWST:~
> 
> - ~if catfile is `None` or empty/whitespace: returns `None`;~
> - ~it skips blank lines and comments (# ...);~
> - ~each non-comment line is split on whitespace:~
>   - ~2 columns: `data_model catalog_file`:~
>     - ~stored as `{data_model: <catfile_dir>/<catalog_file>}` (catalog path made relative to the catfile’s directory);~
>   - ~1 column (data model only):~
>     - ~stored as `{data_model: None}` (meaning “no custom catalog provided for this model; let tweakreg generate one / skip custom for it”);~
>   - ~\>2 columns:~
>     - ~raises ValueError("'catfile' can contain at most two columns.").~

romancal/datamodels/library.py
mairanteodoro: I'm OK with that. I just kept the two ways to avoid breaking anything at this point. We can open a new ticket to make sure that when dealing with the catalog filename, we only set/use `meta.source_catalog.tweakreg_catalog_name`.

Just to clarify the main comment (the many options to specify a catalog), currently, we don't fully support option 4 (see this for more details). So, if we drop that, the only options are either an array with the data (option 1) or a string with the catalog filename (option 2 and 3, in which case the latter takes precedence over the former).

docs/roman/tweakreg/README.rst
mairanteodoro: Yes, `use_custom_catalogs` is set to `False` by default and that's how we normally run `TweakRegStep`.

schlafly: Right, but the docs go on to say here that:
> To provide a custom catalog for a given input model, set
> ``meta.source_catalog.tweakreg_catalog_name`` to the custom catalog filename.

So it sounds to me like we get the catalog from tweakreg_catalog_name regardless of whether use_custom_catalogs is set?

romancal/datamodels/library.py
```diff
@@ -70,7 +70,16 @@ def _assign_member_to_model(self, model, member):
         # __setitem__ calls here instead of setattr
         for attr in ("tweakreg_catalog",):
             if attr in member:
+                # leave this for backwards compatibility, but also add it to
+                # the source_catalog as a filename for the catalog
```
braingram: Since we never actually supported "meta.tweakreg_catalog" I say we remove setting this attribute.

romancal/datamodels/library.py
braingram: Would it be simpler to update the tweakreg code to look at the association information directly (`lib.asn["products"][0]["members"]` etc...) rather than patching the file on open?

docs/roman/tweakreg/README.rst
mairanteodoro: Good catch — the current wording is ambiguous.

In the current implementation, `use_custom_catalogs` doesn’t "gate" whether `meta.source_catalog.tweakreg_catalog_name` is used; it only controls whether we apply/override per-model catalog metadata from external inputs like a catfile (if we decide to continue supporting it) or an association file. So if `meta.source_catalog.tweakreg_catalog_name` is already populated on a model, `TweakReg` will read from it regardless of `use_custom_catalogs`.

I’ll update the docs section to make that explicit and clarify that `use_custom_catalogs=True` is required only for catfile or ans-driven injections. Unless that's not the intended behavior, in which case I'll have to update the code, instead.

schlafly: Thanks.  Yes, updating the docs is the approach I would take here; I just couldn't figure out what this was controlling.  Bringing in association / catfile makes conceptual sense to me (though we have decided to remove catfile support in the linked issue).

It would be helpful if the docs also covered what the priority list was, i.e., if you specify an in-memory catalog, a file name, and an external catalog via use_custom_catalogs, what wins?  I guess the external catalog?  And if there is both an in-memory catalog and a file name, which wins?

In this context, do we need the --use_custom_catalogs argument at all, and can we just replace it with "use the association if present; otherwise use the in-memory option; otherwise use the file name", for example?  It seems to me if we remove catfile and make the association the way to specify that, then --use_custom_catalogs doesn't really have a role.

mairanteodoro: Yup, I agree with dropping `use_custom_catalogs` now that we'll also drop support for catfiles. I'll update the docs and the code to account for that.

romancal/datamodels/library.py
mairanteodoro: Once PR #2269 has been merged, I can take a look at how much effort this would take.

romancal/datamodels/library.py
mairanteodoro: Done.

braingram:
Thanks for making these updates. Once this is rebased can we address the association catalog handling in this PR? On main this currently does not work as expected (add "tweakreg_catalog" keys to member entries does trigger the step to use the provided catalogs). The custom catalog paragraph could also use some clarification:
> Association files can also be used as tweakreg input for custom catalogs. When an association is provided, tweakreg reads the custom catalog information for each member from that member’s meta.source_catalog.tweakreg_catalog or meta.source_catalog.tweakreg_catalog_name metadata.

The above doesn't mention how to define the custom catalog and the `meta.source_catalog.*` values are what the pipeline produced catalogs use (so in this sense aren't "custom").

mairanteodoro:
The clarification on how to set the custom catalog within an association file is being fixed here:
- https://github.com/spacetelescope/romancal/pull/2276

This PR does fix the broken branch when using an association file. Now, with this PR, we set the correct attribute (`meta.source_catalog.tweakreg_catalog_name`) that will actually be used in the tweakreg step.

braingram (approved):
Thanks for making these updates.

The `ModelLibrary` changes to avoid needing to patch the model metadata (https://github.com/spacetelescope/romancal/pull/2265/changes#r3075256150) can be handled in a different PR.

schlafly (approved):

PR #2269, opened 2026-04-14
RCAL-1273: Drop support to catfiles.

### LINKED ISSUE #2082: Improve catfile parsing and error messages

mairanteodoro:
Resolves [RCAL-1273](https://jira.stsci.edu/browse/RCAL-1273)

Closes #2082 

This PR drops support for catfiles in `romancal` by removing related code paths and any associated configuration/documentation so the package no longer advertises or maintains compatibility with catfiles.

Note that **custom catalogs can still be used**, since `TweakRegStep` does not fundamentally depend on catfile; it depends on `meta.source_catalog` being present, with either:
  - `meta.source_catalog.tweakreg_catalog_name` (path to a catalog file), or
  - `meta.source_catalog.tweakreg_catalog` (an in-memory array that can be converted to an astropy.table.Table).
 
So after dropping the support for catfiles, the way to provide a custom catalog should be through setting `meta.source_catalog.tweakreg_catalog_name` (or `tweakreg_catalog`) on each input model.

## Regression tests
All tests are passing.
- https://github.com/spacetelescope/RegressionTests/actions/runs/24473659892 (04/15/26)

## Tasks


docs/roman/tweakreg/README.rst
```diff
@@ -8,63 +8,24 @@ Overview
...
-The catalog must be in a format automatically recognized by
-:py:meth:`~astropy.table.Table.read`. The catalog must contain
+Custom source catalogs can be supplied through
+``meta.source_catalog.tweakreg_catalog`` (in-memory table) or
+``meta.source_catalog.tweakreg_catalog_name`` (file path). The catalog must
+be in a format supported by :py:meth:`~astropy.table.Table.read` (or parquet
+for ``.parquet`` files). The catalog must contain
 either ``'x'`` and ``'y'`` or ``'x_psf'`` and ``'y_psf'`` columns which
 indicate source *image* coordinates (in pixels). Pixel coordinates are
 0-indexed.
```
schlafly, line 25: Can we describe the association mechanism here as well?  I think that's the only one that works without editing the L2 file, which seems worth mentioning to me.

schlafly (changes requested):
This looks good; I approve pending one minor doc update.  Thanks!

docs/roman/tweakreg/README.rst
mairanteodoro: Added a short paragraph to describe the usage of asn files as custom catalogs input.

schlafly: Do you know how this actually works?  In the association file the member isn't meta.source_catalog.tweakreg_catalog_name?

schlafly:
Can you try updating the romanisim lower pin to 0.13.1 to fix the oldestdeps failure?
    "romanisim>=0.13.0,<0.14", ->     "romanisim>=0.13.1,<0.14",

docs/roman/tweakreg/README.rst
mairanteodoro: In PR #2265, we updated the `_assign_member_to_model()` method so that it does the following:
```python
for attr in ("tweakreg_catalog",):
    if attr in member:
        model.meta.source_catalog["tweakreg_catalog_name"] = member[attr]
```
whenever the ASN file has `member.tweakreg_catalog` set. So the input ASN structure should look like this for a custom catalog:
```json
{
    "asn_type": "image3",
    "asn_rule": "candidate_unspecified",
    "version_id": null,
    "code_version": "0.1.0",
    "degraded_status": "No",
    "program", "12345",
    "products": [
        {
            "name": "r0000101001001001001_01101_0001",
            "members": [
                {
                    "expname": "r0000101001001001001_01101_0001_wfi01_cal.asdf",
                    "exptype": "science",
                    "tweakreg_catalog": "/path/to/my_custom_catalog.ecsv"
                },
                {
                    "expname": "r0000101001001001001_01101_0002_wfi01_cal.asdf",
                    "exptype": "science",
                    "tweakreg_catalog": "/path/to/another_custom_catalog.ecsv"
                }
            ]
        }
    ]
}
```
This implementation has been tested in that PR and it works fine.

One thing, though: I think we might have to change the attribute name both in the `for` loop as well as in the ASN file to `tweakreg_catalog_name` to be consistent with the nomenclature used in `TweakReg`.

schlafly (approved):

PR #2280, opened 2026-04-21
Add python 3.14 support, add upper pins to some dependencies

braingram:
Resolves https://jira.stsci.edu/browse/RCAL-1381
Resolves https://jira.stsci.edu/browse/RCAL-1328

Add a few upper pins for:
- astropy
- photutils

Cleanup a few pyproject.toml comments.

Bump the python upper pin to `<3.15`. FWIW 3.15 beta is scheduled for early May.

Branch protections will need to be updated with merging of this PR.

Regtests with python 3.14 here: https://github.com/spacetelescope/RegressionTests/actions/runs/24738534491/job/72371767757

## Tasks


zacharyburnett (approved):

larrybradley:
Are there plans to drop support for Python 3.11?
Scientific python packages (e.g., numpy, scipy) are starting to drop 3.11.

Ref: https://scientific-python.org/specs/spec-0000/

braingram:
> Are there plans to drop support for Python 3.11? Scientific python packages (e.g., numpy, scipy) are starting to drop 3.11.
> 
> Ref: https://scientific-python.org/specs/spec-0000/

Overall I'm not in favor of dropping support for 3.11 for a few reasons:
- We are fully compatible with 3.11. If a user wants to use 3.11 they can install romancal. If we increase our lower pin they will have issues.
- We have dependencies that have a significant lag time for supporting new versions. 3.14 was stable in Oct 2025 and it has taken 6 months for us to claim support for it due to various dependencies not providing releases supporting 3.14. This lag in being able to adopt new versions means we have a much smaller allowable range. Since the motivation for spec0 is "Limiting the scope of supported dependencies is an effective way for packages to limit maintenance burden" this is already addressed by the lagging dependencies.

I'll refrain from comments on spec0 in general.

schlafly (approved):
Looks good to me, thanks!

PR #2281, opened 2026-04-23
PSF fitting updates

larrybradley:

This PR updates PSF fitting in a few ways:
* Removes the unneeded (public) `fit_psf_to_image_model` function
* Removes the docs for `fit_psf_to_image_model`
* Moves the PSF jitter information to the main source catalog docs
* Changes the PSF fitting to use a (5, 5) region instead of a (15, 15) region.  15x15 is unusually large.  Most of the PSF information is in the highest S/N central region.  
* Removes the local background estimation and subtraction from the PSF step -- relying on the global background subtraction, which is also used for aperture- and segment-based photometry.
* Fixes a circular import issue in psf_matching.py
* These changes reduce the runtime (for the entire elp) of an example dataset (r0034001001001002003_0002_wfi01_f106_cal.asdf) with 39,340 sources from ~7.5 to ~4.5 min.

## Tasks


larrybradley:
Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/24851970648

schlafly:
Thanks Larry, this looks good.  Specifically for the removal of a local background estimation in the PSF photometry, is that expensive or is there a way to configure it?  That's one where I know that people have strong feelings.

larrybradley:
When local background is turned on in PSF photometry it does the subtraction as well.  It's not a very expensive operation. In the example above with 39k sources, with local background subtraction adds about 24s (~4m 58s with vs ~4m 34s without for the entire elp run).  By default, I think we should leave it off, but I could add parameters to the source catalog step ~(inner/outer annulus radius)~ (perhaps just an on/off toggle) so that it could be turned on.

EDIT:  We do already *report* (but do not subtract) a local background measured in a circular annulus in the output catalogs.  The inner/outer radii for that is not configurable -- it's set by a reference file.  Is that sufficient?

schlafly:
Thanks Larry.  You're right, the background in the annulus could be adequate.  But to get the same answer as local-background-subtracted PSF photometry I think we'd need to also tabulate something like sum(psf) / sum(psf * psf) for the PSF stamp used for each source; usually we term this n_effective for the PSF.  Then the LBS PSF photometry would be psf_flux - n_effective * local_background, maybe with a factor of the pixel area?

I'm happy to leave it off if we have that (new?) field.

schlafly (approved):
Approving, spinning out n_effective computation to an issue.

PR #2282, opened 2026-04-27
Refactor dq_init to move logic out of dq_init_step

schlafly:
This PR simplifies the dq_init step.  The primary goal is to separate the step & reference handling logic from the more algorithmic logic; the former stays in the stpipe step infrastructure and the latter moves to dq_initialization.py.  This allows other callers (specifically, the HLIS PIT) to use this step without using the full stpipe infrastructure.

While making these changes, I also did the following:
- removed support for guider type exposures.  I do not believe these can ever go through romancal.  @tddesjardins confirmed this.  @ddavis-stsci , do you remember the history here?  I'm not sure what romancal could usefully do for these exposures.  The old code largely skipped dq_init for them, but I feel like every other step should likewise be skipped for them.
- added a feature to expand the guide-window masks by one pixel.  The HLIS PIT supports this so I didn't want to drop this mode; it's possible we'll want it as well.  Default off.
- added a test for the new guide-window expansion feature
- clipped the x_start / x_stop / y_start / y_stop pixels to no longer mask way off the image for -999999 starts / stops.
- added lazy_load = False to the initial file opening in the pipeline.  This is because (1) we need to access everything in this file anyway in dq_init, so lazy loading provides no benefit, and (2) because of the way this step interacts with TVAC files, not lazy loading leads to errors for TVAC files.  See https://github.com/spacetelescope/roman_datamodels/issues/631#issuecomment-3969512538 for more details.

## Tasks
  - [x] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job (https://github.com/spacetelescope/RegressionTests/actions/runs/25011793884/job/73249179511)

### DIFF
--- a/romancal/pipeline/exposure_pipeline.py
+++ b/romancal/pipeline/exposure_pipeline.py
@@ -84,16 +84,16 @@ class ExposurePipeline(RomanPipeline):
         log.info("Starting Roman exposure calibration pipeline ...")
 
         # determine the input type
+        # Because we're processing raw files, let's open without any
+        # laziness; we need to propagate all of the bits into the ramps
+        # anyway.  It also avoids bugs like:
+        # https://github.com/spacetelescope/roman_datamodels/issues/631
         lib, input_type = open_dataset(
             dataset,
             update_version=self.update_version,
             return_type=True,
             as_library=True,
             open_kwargs={"on_disk": self.on_disk, "lazy_load": False},
-            # Because we're processing raw files, let's open without any
-            # laziness; we need to propagate all of the bits into the ramps
-            # anyway.  It also avoids bugs like:
-            # https://github.com/spacetelescope/roman_datamodels/issues/631
         )
         return_lib = input_type in ("ModelLibrary", "asn")
 
--- a/romancal/pipeline/exposure_pipeline.py
+++ b/romancal/pipeline/exposure_pipeline.py
@@ -89,7 +89,11 @@ class ExposurePipeline(RomanPipeline):
             update_version=self.update_version,
             return_type=True,
             as_library=True,
-            open_kwargs={"on_disk": self.on_disk},
+            open_kwargs={"on_disk": self.on_disk, "lazy_load": False},
+            # Because we're processing raw files, let's open without any
+            # laziness; we need to propagate all of the bits into the ramps
+            # anyway.  It also avoids bugs like:
+            # https://github.com/spacetelescope/roman_datamodels/issues/631
         )
         return_lib = input_type in ("ModelLibrary", "asn")
 
--- a/changes/2282.dq_init.rst
+++ b/changes/2282.dq_init.rst
@@ -1,0 +1,1 @@
+Moved DQ init logic into dq_initialization.py and added expand_gw_flagging step parameter.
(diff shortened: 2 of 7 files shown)


schlafly:
Regression tests pass.  https://github.com/spacetelescope/RegressionTests/actions/runs/25011793884/job/73249179511

romancal/pipeline/exposure_pipeline.py
```diff
@@ -84,12 +84,16 @@ def process(self, dataset):
...
+        # laziness; we need to propagate all of the bits into the ramps
+        # anyway.  It also avoids bugs like:
+        # https://github.com/spacetelescope/roman_datamodels/issues/631
         lib, input_type = open_dataset(
             dataset,
             update_version=self.update_version,
             return_type=True,
             as_library=True,
-            open_kwargs={"on_disk": self.on_disk},
+            open_kwargs={"on_disk": self.on_disk, "lazy_load": False},
```
braingram, line 96: I'd like to find a way to avoid disabling lazy load since it will also impact models after dq_init (and if dq_init is skipped). Ideally we'd fix rdm but since that's out-of-scope for this build perhaps we can work-around the issue here. This seems easiest to handle in a follow-up PR. Does that work for you? If so, any additional information you can provide on what issue(s) occur with lazy_load would be helpful.

ddavis-stsci:
Early in the mission it was not clear if we would have to process guide exposures or what the processing would be. It was decided that we would process them from raw science data to ramp data with DQ values for archiving. Anything else was TBD.

ddavis-stsci (approved):
LGTM

romancal/pipeline/exposure_pipeline.py
schlafly: Thanks Brett.  I wasn't super worried about this because in the usual operations dq_init loads the raw model and returns a ramp model, and everything else sees the ramp model.  And the ramp model loads everything in the raw model, not benefiting from lazy loading.  Maybe someone could pass a ramp model to the ELP and then skip dq init, and then disabling lazy loading hurt them, but that case seemed niche enough to warrant letting lazy_load be false until we fix the underlying issue in rdm.

I think your suggestion is that we go ahead and merge this and you pursue a better approach in a separate PR?

The problem lazy loading was causing was that for TVAC data only, the from_tvac_raw (https://github.com/spacetelescope/romancal/blob/main/romancal/dq_init/dq_init_step.py#L57) wasn't materializing all of the stuff it copied into the new file.  This meant that when we closed the input model in dq_init (https://github.com/spacetelescope/romancal/blob/main/romancal/dq_init/dq_init_step.py#L135) and then eventually went to write out the file, asdf tried to load from the closed file and crashed.  This bug is hidden in main because this line (https://github.com/spacetelescope/romancal/blob/main/romancal/dq_init/dq_init_step.py#L57) overwrites the input model, so when we close the input model in dq_init, we are really closing the science raw model that the TVAC model is converted into, and not the input model.  So the underlying asdf file isn't closed for the TVAC case and life continues.  I inadvertently fixed that bug as part of this PR, causing the rdm issue to trigger.

braingram: Thanks! Yes, the suggestion is to address the `lazy_load` usage in a follow-up.

PR #2286, opened 2026-04-28
Dark and saturation step updates for HLIS pipeline

### LINKED ISSUE #2285: Update saturation and dark subtraction steps to enable better HLIS pipeline integration
The HLIS saturation pipeline expands saturated pixels to their neighbors and backs up saturation-flagged pixels by one (so that a resultant will be flagged as saturated if the following resultant crosses the saturation limit.  We should optionally support these modes.

We should also update the dark_current step to separate the step logic (input, output, reference files) from the subtraction and flagging updating, to enable use of this step without going through stpipe.

schlafly:
Closes #2285 .

This PR adds two arguments to the saturation step spec, n_pix_grow_sat and backup, which expand regions flagged by saturation.  n_pix_grow_sat is passed on to stcal, which grows saturation-flagged regions to also mark their neighbors.  backup is handled within the step, growing the saturation backward in time to flag the resultants before the saturation-flagged resultants.  Growing the saturation is pretty reasonable and should probably be our default, though RTB should decide that; when a pixel saturates the neighbors go non-linear pretty quickly as charge migration increases.  Backing up the saturation flags in time feels overly conservative to me---we try pretty hard to flag anything where any read would have saturated---but it's a reasonable conservative option.

This PR also updates the saturation tests and docs accordingly.

Lastly, it includes a minor refactoring of the dark step to keep reference file and input / output logic in the step and algorithmic logic in a separate routine called by the step.

Regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/25333817873

Following discussion with Brett, I also deleted the --dark-output option from the spec of the dark subtraction step.  For Webb, an interpolated dark cube is constructed, but for Roman we don't do this and this option just wrote the dark reference file out again with a new name.  This doesn't seem useful and has been removed.

## Tasks


romancal/saturation/tests/test_saturation.py
```diff
@@ -332,6 +332,66 @@ def test_saturation_getbestref(setup_wfi_datamodels):
...
+    # Create fresh models for the backup=1 case
+    ramp2, satmap2 = setup_wfi_datamodels(nresultants, nrows, ncols)
+    ramp2.meta.exposure.read_pattern = [[1], [2, 3], [4]]
+    ramp2.data[0, 5, 5] = 0
+    ramp2.data[1, 5, 5] = 30000
+    ramp2.data[2, 5, 5] = 62000
+    satmap2.data[5, 5] = satvalue
+
+    # With backup=1: resultant 1 (multi-read) also gets SATURATED from resultant 2
+    output2 = flag_saturation(ramp2, satmap2, backup=1)
```
braingram: To check I'm following, setting `backup=2` here should produce the same result as `backup=1` as the first element in read_pattern only contain 1 item?

If so, would you add a test for that?

schlafly: You got it.  I've updated these tests to use 4 resultants instead of three to more explicitly see that increasing backup increases flagging but doesn't propagate to the single element resultant.

braingram: Thanks! The test updates look great.

braingram:
ruff is complaining about an unused variable:
```

F841 Local variable `image_shape` is assigned to but never used
  --> romancal/dark_current/tests/test_dark.py:77:5
   |
75 |     # Set test size
76 |     ref_shape = (10, 28, 28)
77 |     image_shape = (ref_shape[1], ref_shape[2])
   |     ^^^^^^^^^^^
78 |
79 |     # Create test ramp and dark models
   |
help: Remove assignment to unused variable `image_shape`

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```
otherwise LGTM.

braingram (approved):
Thanks!

PR #2287, opened 2026-04-29
Fix dq_init model handling to allow re-enabling lazy_load for elp

braingram:
Follow up to https://github.com/spacetelescope/romancal/pull/2282

This PR:
- fixes the dq_init model handling to not close the provided input file (which may not always be appropriate)
- adds a model closing to elp (when dq_init returns a different model)
- drops use of `from_tvac_fps` and `from_science_raw` replacing these with a new method `to_ramp_model` which largely uses `create_fake_data` (for tvac/fps files) and `create_from_model` (for science raw). More on this below.
- re-enables lazy_load for elp

Regtests show unrelated failures and some differences:
https://github.com/spacetelescope/RegressionTests/actions/runs/25078786707
The differences are:
- the elp no longer produces a file with the tagged scalar TvacPrdSoftwareVersion and instead uses the correct str
- the elp no longer produces cal_step metadata with the non-existent source_detection and jump steps
(Note I will need to make a clean run for okifying as the above is a rerun).

Also memory usage of elp is improved by 1.4x (1379 MB vs 1935 MB).

## Tasks

### DIFF
--- a/romancal/dq_init/dq_initialization.py
+++ b/romancal/dq_init/dq_initialization.py
@@ -8,6 +8,48 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
+def to_ramp_model(model):
+    if isinstance(model, RampModel):
+        return model
+
+    if isinstance(model, (FpsModel | TvacModel)):
+        science_raw = ScienceRawModel.create_fake_data(model)
+        if "statistics" in science_raw.meta:
+            science_raw["extras"] = {"tvac": science_raw.meta.pop("statistics")}
+        return to_ramp_model(science_raw)
+
+    ramp_model = RampModel.create_from_model(
+        {
+            **model,
+            "data": None,
+        }
+    )
+
+    ramp_model.meta.cal_step = {}
+    for step_name in ramp_model.schema_info("required")["roman"]["meta"]["cal_step"][
+        "required"
+    ].info:
+        ramp_model.meta.cal_step[step_name] = model.meta.get("cal_step", {}).get(
+            step_name, "INCOMPLETE"
+        )
+
+    shape = model.data.shape
+    ramp_model.data = model.data.astype(np.float32)
+    ramp_model.pixeldq = np.zeros(shape[1:], dtype=np.uint32)
+
+    # check if the input model has a resultantdq from SDF
+    if hasattr(ramp_model, "resultantdq"):
+        ramp_model.groupdq = ramp_model.pop("resultantdq")
+    else:
+        ramp_model.groupdq = np.zeros(shape, dtype=np.uint8)
+
+    # check for exposure data_problem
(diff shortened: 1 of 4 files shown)


romancal/dq_init/dq_initialization.py
```diff
@@ -1,13 +1,83 @@
...
+    if isinstance(model, (FpsModel | TvacModel)):
+        science_raw = ScienceRawModel.create_fake_data(
+            {
+                **model,
+                "data": model.data.value,
+                "amp33": model.amp33.value,
+            }
+        )
+
+        # work-around issue with roman_datamodels not copying the nested read_pattern
```
braingram, line 27: I'm working on a rdm issue for this.

schlafly:
Thanks Brett, this is a clear improvement.  We would consider deleting the from_science_raw and from_tvac_raw routines in roman_datamodels in the future?

Broadly, my sense is that you prefer roman_datamodels to be strictly about reading and writing roman datamodels, in which case this code clearly belongs here.  I think our old line separating romancal and roman_datamodels was more along the lines of whether there was any real changes to the data vs. moving it around.  I'm happy moving that line so that roman_datamodels is skinnier---just confirming that something along those lines should be the new policy.

schlafly (approved):

braingram:
> Thanks Brett, this is a clear improvement. We would consider deleting the from_science_raw and from_tvac_raw routines in roman_datamodels in the future?

Yeah I think that makes sense but we should first check with @WilliamJamieson as he's been looking into a proposal for https://github.com/spacetelescope/rad/issues/869 I'll refrain from trying to fully connect the dots here but I have a hunch that https://github.com/spacetelescope/roman_datamodels/issues/655 would make handling all of this easier as we could more easily add hooks into tvac/fps handling to do simple things like strip out the tagged sub-nodes.

> Broadly, my sense is that you prefer roman_datamodels to be strictly about reading and writing roman datamodels, in which case this code clearly belongs here. I think our old line separating romancal and roman_datamodels was more along the lines of whether there was any real changes to the data vs. moving it around. I'm happy moving that line so that roman_datamodels is skinnier---just confirming that something along those lines should be the new policy.

Thanks for detailing that. Yeah that aligns with what I'm thinking. One way to look at this is that someone concerned with how romancal processes tvac data would likely first look at/track romancal to know how the data will be processed. Splitting data transformations across romancal and roman_datamodels would be harder to understand than keeping all of these transformations within romancal.

braingram:
I started a fresh set of regtests for okifying: https://github.com/spacetelescope/RegressionTests/actions/runs/25122721236

romancal/dq_init/dq_initialization.py
braingram: See https://github.com/spacetelescope/roman_datamodels/pull/659

braingram:
The fresh run shows only the expected differences for tvac data where the cal_step entries for removed steps are no longer present and the output no longer contains a tagged scalar.

PR #2293, opened 2026-05-01
Removed pixel response function convolution for ePSF

t-brandt:

Closes #

This PR removes the convolution of the ePSF from CRDS with the pixel response function.  The ePSF should, by definition, already be the PSF as sampled by the detector, and therefore should require only sampling, not interpolation.

## Tasks


schlafly:
Do you understand the test failures, e.g., https://github.com/spacetelescope/romancal/pull/2293/changes#diff-4d57ffc87f28fca9e1191deb55e443c154b99e87bb10827910f1840c23b089e5L288 ?  I think it takes the same path through the now unconvolved PSFs so I'm surprised at the difference.  It's possible that photutils doesn't like the normalization of the current CRDS reference file, and a factor of oversample ** 2 is needed?

t-brandt:
@schlafly--I think that is very likely.  The current ePSF reference files are normalized to a sum of 1, which is not the way ePSF files are constructed and delivered.  A usual ePSF file will be normalized to a sum of 1 when sampling, i.e., to a sum of oversample ** 2 when not resampling.  The difference is a factor of oversampled ** 2, or 16 in this case.  The failures seem to be from numerical comparisons of the flux and PSF shape, and if you put in a PSF with a different normalization, that will propagate to the derived fluxes.  The quickest way to check is probably to manually change the normalization by that factor of oversampled ** 2 and re-run the regression tests, verifying that they then pass.

mairanteodoro (commented):
Aside from figuring out what is causing the differences in the unit tests, the code looks good to me!

schlafly:
Test failure is due to a transient VO issue.  Regression tests here pass: https://github.com/spacetelescope/RegressionTests/actions/runs/25565540966 .

PR #2303, opened 2026-05-05
Reduce public interface of ELP steps

schlafly:
This PR refactors several steps in the exposure pipeline, aiming to:
1. reduce the amount of public API in preparation for romancal version 1.0 and introducing semantic versioning
2. makes each ELP step have a separate callable function outside the stpipe step object
3. eliminates some accumulated cruft along the way.

e.g., the flatfield.do_correction function that did nothing but call flatfield.do_flat_field has been removed.

The refpix data.py module was renamed _data.py and references to it were updated.

The ugliest bit of this was moving ols_cas22 and likely outside of the ramp fit step; that required adding a bunch of arguments to those functions that were previously accessed through the step object.

I considered removing the deprecated ols_cas22 but decided against it at this point.

Regression tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/25393049798 .

## Tasks


braingram (approved):
Thanks for the improvements!

If removing the deprecated ramp fitting makes the code simpler it might be worth considering as a separate PR (if time allows for this build) but the changes in this PR look good to me.

schlafly:
I felt like I wanted to leave the ols_cas22 as an option for commissioning, so I left it in for now.  Thanks!

PR #2306, opened 2026-05-06
Tweak to the DataModel migration function 

WilliamJamieson:
This currently RCAL is using private RDM API as part of version migration tools for DataModels. RDM intends to modify this internal API (e.g. spacetelescope/roman_datamodels#669) and is providing a public API hook for RCAL to use instead in spacetelescope/roman_datamodels#671 which will be maintained so that RCAL can be independent of some of the private internals of RDM.

This PR switches RCAL to using the public functionality from spacetelescope/roman_datamodels#671.

## Tasks


pyproject.toml
```diff
@@ -25,7 +25,8 @@ dependencies = [
     # the roman_datamodels pin to the released version. If there
     # is a need to change this to main then we need a new
     # roman_datamodels release.
-    "roman_datamodels>=0.31.0, <0.32",
+    # "roman_datamodels>=0.31.0, <0.32",
+    "roman_datamodels@git+https://github.com/WilliamJamieson/roman_datamodels.git@feature/move_basic_migrate",
```
WilliamJamieson, line 29: This change is only to demonstrate that this will work with the changes in RDM. I expect to update this before this PR is merged.

PR #2307, opened 2026-05-06
Correct amp33 for encoding offset and update docs

tddesjardins:
This PR corrects the amp33 array for the data encoding offset in the DQ initialization step. According to Tim, a DC offset should result in either a negligible or no change in the IRRC correction, but it is more correct to account for this in the amp33 array. Documentation on the step has also been updated to mention the encoding offset.

### DIFF
--- a/changes/2307.dq_init.rst
+++ b/changes/2307.dq_init.rst
@@ -1,0 +1,1 @@
+Fix application of data encoding offset to also subtract from the amp33 array.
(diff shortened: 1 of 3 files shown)


schlafly:
Started a regtest here: https://github.com/spacetelescope/RegressionTests/actions/runs/25442156254

schlafly:
(thanks!)

schlafly (approved):
Thanks!

PR #2308, opened 2026-05-06
Reduce assign_wcs public API

schlafly:
This PR continues reducing the romancal public API, for the assign_wcs step.  I also deleted some empty patch_match directories (leftover from skycells?) and updated a little bit of calling code within romancal that used routines in assign_wcs.  

Regression tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/25440309755 .

## Tasks


braingram (approved):
Changes LGTM.

One question on the various functions left public:
- assign_wcs_step.load_wcs
- pointing.v23tosky
- utils.add_s_regions

Would it be an improvement to consolidate these into one submodule (perhaps `assign_wcs.py`)? That could be done in a separate PR, in this one or after the 1.0.0.

romancal/assign_wcs/__init__.py
```diff
@@ -1,8 +1,9 @@
 from astropy.modeling.models import Mapping
 
+from .assign_wcs import load_wcs
 from .assign_wcs_step import AssignWcsStep
 
-__all__ = ["AssignWcsStep", "_distorted_to_undistorted"]
+__all__ = ["AssignWcsStep", "_distorted_to_undistorted", "load_wcs"]
 
 
 def _distorted_to_undistorted(wcsobj):
```
braingram, line 9: I didn't notice this earlier but it's presence in `__all__` seemed odd.

This is unused and untested but appears to be added based on a request from SSC:
https://github.com/spacetelescope/romancal/pull/1612

Are there tickets/plans for follow-up work?

Reading over the comments, etc it appears that:
- there is not yet a replacement for this function
- testing it locally it doesn't crash but it's unclear if it's returning the correct result

Since it doesn't look like a simple "remove unused code" it seems easier to handle this all in a separate PR and leave things as-is here (which this PR is already doing).

braingram (approved):
Thanks for consolidating these. I'm giving this another approval given those changes so this can be merged once the regtests pass.

romancal/assign_wcs/__init__.py
schlafly: Yeah, I also rediscovered this function when doing this refactoring.  There has been no follow-up conversation about this function.  It seems unlikely to me that it's used and it clearly depends on the internals of the gwcs object and will eventually break.  Nadia had plans for separately breaking out the different transforms.  I think Takahiro was the ultimate requester here; I can try to follow up but consider this to be likely dead code.

PR #2309, opened 2026-05-06
Update stcal upper pin in pyproject.toml

schlafly:
We are hoping to move to something closer  to semantic versioning with stcal.  Webb currently imposes a <2 pin.  This PR changes romancal to follow their example.


braingram (approved):
Thanks!

PR #2310, opened 2026-05-06
add CRDS to get_stpipe_loggers

braingram:
stpipe now has a viable path for logging crds log messages.

In preparation for those changes this PR adds "CRDS" to the list of stpipe-aware loggers. The name is upper-case since crds uses that particular name:
https://github.com/spacetelescope/crds/blob/8785ff0a1f5bef8a9fc629ccf19fc12d198ff4cf/crds/core/log.py#L278

Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/25522417601
and with stpipe main: https://github.com/spacetelescope/RegressionTests/actions/runs/25574428798
both all pass

Testing locally with stpipe main this works as expected and the crds log messages are logged. This was tricky to test since most crds usage (for my setup with a valid crds cache) that results in log messages are errors during pre-step code where steps are missing pars files. These aren't captured yet so I had to enable debug logging in crds which produced the expected flood of log messages.

## Tasks

### DIFF
--- a/changes/2310.stpipe.rst
+++ b/changes/2310.stpipe.rst
@@ -1,0 +1,1 @@
+Add "CRDS" to list of loggers configured by pipeline runs.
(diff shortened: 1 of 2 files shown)


schlafly (approved):
Thanks!

PR #2313, opened 2026-05-07
RCAL-1380: Populate SFD-estimated extinction in catalogs.

mairanteodoro:
Resolves [RCAL-1380](https://jira.stsci.edu/browse/RCAL-1380)

This PR adds support for populating a new `dust_ebv` extinction column in Roman source and multiband catalogs using SFD dust maps resolved through CRDS. It updates the catalog schema and source-catalog generation logic, adds tests and regression checks for the new field.

## Note
This adjusts CI/tox setup to use temporary git-based dependency installs needed for the change. ~We need to revert it before merging this PR.~ Done (unit tests will fail until the [RAD PR](https://github.com/spacetelescope/rad/pull/870) is merged).

## Regression tests
- https://github.com/spacetelescope/RegressionTests/actions/runs/25570563604 (05/08/26; all tests passing except for one that needs to have the truth file updated later).

## Tasks


romancal/source_catalog/_source_catalog.py
```diff
@@ -425,6 +433,181 @@ def image_flags(self):
...
+        ):
+            dm_headers = {"roman.meta.instrument.name": "wfi", "roman.field": field}
+            reference = getreferences(
+                dm_headers, reftypes=["dustmap"], observatory="roman"
+            )
+            map_paths[pole] = self._extract_dustmap_path(reference, field)
+        return map_paths
+
+    @staticmethod
+    def _extract_dustmap_path(reference, field):
```
schlafly: I have experience with crds.getreferences(...) failing due to an unknown reference type, but I'm unfamiliar with a mechanism where CRDS returns a dictionary but it doesn't have the key that was expected.  I think I'd delete this function and let the caller handle CRDS failures?

romancal/source_catalog/_source_catalog.py
```diff
@@ -649,7 +832,8 @@ def _assemble_catalog(self):
         for column in self.column_names:
             catalog[column] = getattr(self, column)
             definition = self.cat_model.get_column_definition(column)
-            catalog[column].info.description = definition["description"]
+            if definition is not None:
+                catalog[column].info.description = definition["description"]
```
schlafly: This is needed because the field doesn't currently exist in rad?  Is this needed once the field is added to rad?

romancal/source_catalog/_column_schema.py
```diff
@@ -248,6 +252,7 @@ def column_names(self):
             colnames.extend(skywin_colnames)
             colnames.extend(det_colnames)
             colnames.extend(flag_columns)
+            colnames.extend(dust_colnames)
```
schlafly, line 255: When we go to require this in rad products, let's require it only in the prompt & multiband products, not the forced products.

schlafly (commented):
This looks good.  A few comments inline.  Please add the new E(B-V) field to the list of fields the regression tests look for (this PR presently only adds unit tests).  Thanks!

romancal/source_catalog/_source_catalog.py
mairanteodoro: Once the field's in RAD, we can remove the guard.

schlafly: Can you remove the guard now and point pyproject.toml to your rad PR?

romancal/source_catalog/_source_catalog.py
mairanteodoro: Removed.

romancal/source_catalog/_source_catalog.py
mairanteodoro: Done.

romancal/source_catalog/_column_schema.py
mairanteodoro: OK.
Just so you know, with the current code, we partially populate the forced products:
- **`forced_full`** includes `dust_ebv` (because it shares the prompt-style column set in `CatalogSchema`).
- **`forced_det`** does not include `dust_ebv`.

In the final forced output (the join of `forced_full` + `forced_det`), we should get a `dust_ebv` column coming from `forced_full`, and it is **not** renamed to `forced_dust_ebv` by the current prefix logic.

schlafly: That's fine.  We might change the forced behavior later but as long as we're not requiring it in rad I think that's okay.

romancal/regtest/test_multiband_catalog.py
```diff
@@ -35,6 +35,7 @@
     "aper02_f213m_flux",  # DMS539 PSF-matched photometry
     "segment_f213m_flux",  # DMS539
     "kron_f213m_flux",  # DMS539
+    "dust_ebv",  # DMSXXX dust extinction values
```
schlafly: We don't need a DMSXXX for this.

schlafly (commented):
Sorry, I worry I made the dependencies more complicated.  Have we changed the procedure there?  My memory is that normally we update pyproject.toml to point to your specific rad branch that has the needed updates.  I don't understand the ci-gitdeps stuff.  Alternatively run regtests overriding the rad dependency to point to your branch and point to the successful regtest run.

mairanteodoro:
> Sorry, I worry I made the dependencies more complicated. Have we changed the procedure there? My memory is that normally we update `pyproject.toml` to point to your specific rad branch that has the needed updates.

That's usually the case for _direct_ dependencies. However, RAD is not a direct dependency of romancal and editing `pyproject.toml` won't honor the required version for RAD set in that file (it will be installed via RDM).

@zacharyburnett and I had a quick chat about this issue and he will draft a PR to try and solve this issue of RAD and RDM depending on each other while being two separate packages. For now, patching `tox.ini/tests.yml` is the temporary solution that I was able to come up with. At least it shows that the guard won't be needed when the RAD changes are in.

> Alternatively run regtests overriding the rad dependency to point to your branch and point to the successful regtest run.

The regtests are fine because we can provide additional parameters that will override the requirements:
https://github.com/spacetelescope/RegressionTests/actions/runs/25570563604

In that regtest run, the only failure is for the `test_multiband_catalog: romancal.regtest.test_multiband_catalog`, and that is because we need to update the truth file later.

romancal/regtest/test_multiband_catalog.py
mairanteodoro: Removed.

romancal/regtest/test_source_catalog.py
```diff
@@ -63,6 +63,7 @@ def fields(catalog):
         "segment_flux_err",  # DMS386 flux uncertainties
         "is_extended",  # DMS376 type of source
         "warning_flags",  # DMS387 dq_flags
+        "dust_ebv",  # DMSXXX dust extinction values
```
schlafly:
```suggestion
        "dust_ebv",
```
There's not an explicit DMS requirement here.

schlafly (commented):
Looks good to me.  One minor comment inline.  Thanks Mairan!

schlafly:
The oldestdeps python 3.11 tests were failing with NoneType in various source catalog steps.  We probably need a rad version requirement to do this; we're saying that's okay here since it should pass in the regtests with the pyproject / roman_datamodels pointing to the right rad version.

zacharyburnett:
> @zacharyburnett and I had a quick chat about this issue and he will draft a PR to try and solve this issue of RAD and RDM depending on each other while being two separate packages. For now, patching `tox.ini/tests.yml` is the temporary solution that I was able to come up with. At least it shows that the guard won't be needed when the RAD changes are in.

I drafted https://github.com/spacetelescope/roman_datamodels/pull/676 to see what making a monorepo of `rad` and `roman_datamodels` would look like, but from talking with @braingram and @WilliamJamieson I don't think that's a viable solution either unfortunately. For now we'll just have to work with this dependency chain as-is

romancal/regtest/test_source_catalog.py
mairanteodoro: Oh, that one slipped away... thanks for catching it.

schlafly (approved):
Thanks, approved.  I'll let you back out the pyproject toml changes and merge when the corresponding rad PR is in.  We'll also need to okify the differences; I'll let you handle that when you merge?

mairanteodoro:
> We'll also need to okify the differences; I'll let you handle that when you merge?

Sounds good.

braingram:
> > @zacharyburnett and I had a quick chat about this issue and he will draft a PR to try and solve this issue of RAD and RDM depending on each other while being two separate packages. For now, patching `tox.ini/tests.yml` is the temporary solution that I was able to come up with. At least it shows that the guard won't be needed when the RAD changes are in.
> 
> I drafted [spacetelescope/roman_datamodels#676](https://github.com/spacetelescope/roman_datamodels/pull/676) to see what making a monorepo of `rad` and `roman_datamodels` would look like, but from talking with @braingram and @WilliamJamieson I don't think that's a viable solution either unfortunately. For now we'll just have to work with this dependency chain as-is

The unit tests are failing with a `TypeError` because:
https://github.com/spacetelescope/romancal/blob/1937ff33a4f2891d031c5eae24e89f60c9dfc4e4/romancal/source_catalog/_source_catalog.py#L652
is attempting to get the "description" where `definition` is `None`. Merging https://github.com/spacetelescope/rad/pull/870 won't fix this as:
- romancal is requiring the released rdm version: https://github.com/spacetelescope/romancal/blob/1937ff33a4f2891d031c5eae24e89f60c9dfc4e4/pyproject.toml#L28
- the development version of rdm is using the released version of rad: https://github.com/spacetelescope/roman_datamodels/blob/210ae0eb8d44d7e16647c30176eb29e4f68c7d01/pyproject.toml#L19

We will need:
- this PR to be updated to point to rdm main
- a rdm PR setting the required rad version to rad main

@mairanteodoro would you update the rdm requirement in this PR? I'll open a rdm PR setting the rad requirement to main.

mairanteodoro:
> Merging [spacetelescope/rad#870](https://github.com/spacetelescope/rad/pull/870) won't fix this as:
> * romancal is requiring the released rdm version: https://github.com/spacetelescope/romancal/blob/1937ff33a4f2891d031c5eae24e89f60c9dfc4e4/pyproject.toml#L28
> * the development version of rdm is using the released version of rad: https://github.com/spacetelescope/roman_datamodels/blob/210ae0eb8d44d7e16647c30176eb29e4f68c7d01/pyproject.toml#L19
> 
> We will need:
> 
> * this PR to be updated to point to rdm main
> * a rdm PR setting the required rad version to rad main
> 
> @mairanteodoro would you update the rdm requirement in this PR? I'll open a rdm PR setting the rad requirement to main.

Yep, I can do that. I assumed that we would have a released version soon, so I didn't think of changing the requirements.

PR #2314, opened 2026-05-07
RCAL-1392 Refactor public vs private interface

stscieisenhamer:
Resolves [RCAL-1392](https://jira.stsci.edu/browse/RCAL-1392)

This PR clarifies the distinction between the intended public API and private API. Significant change is placing most of the utility code into private modules.

## Tasks


stscieisenhamer:
[regression test #1](https://github.com/spacetelescope/RegressionTests/actions/runs/25734032380)
result: clean

[regression test#2](https://github.com/spacetelescope/RegressionTests/actions/runs/25823083678)
reason: significant rebase
result: pending...

schlafly (approved):
Thanks Jonathan.  I'll let you merge when you're happy that the SDP code doesn't care about these changes.

schlafly (commented):
Reviewing more carefully, I think the remaining public API is:
engdb has classes, get_meta, is_alive methods
engdb_lib removed from all but did not underscore the library or classes so I think we treat this as public nominally?
stp has add_wcs, update_wcs: likely intentionally
v1_calculate has v1_calculate_from_models, v1_calculate_over_time, simplify table: do we need these?
scripts: We don't currently have a rule that 'scripts' are private; we could add that to Brett's page.  But these have things like main, deprecated_name, parse_args, and add_dva that are nominally public.  add_dva is the only one I think we're likely to consider that should actually be public?

stscieisenhamer:
**engdb has classes, get_meta, is_alive methods**
Confirm yes: The classes `ENGDB_*` are public and the indicated methods are public

**engdb_lib removed from all but did not underscore the library or classes so I think we treat this as public nominally?**
    Confirm yes: The abstract class on which the above are based. Hence, by definition, public

**stp has add_wcs, update_wcs: likely intentionally**
    Confirm yes

**v1_calculate has v1_calculate_from_models, v1_calculate_over_time, simplify table: do we need these?**
    I would say yes because these represent the programmatic api that implements the command-line script

**scripts: We don't currently have a rule that 'scripts' are private...**
    Will clean those up. I would think that the deprecation notice can be removed since the deprecation was to remove by the next major update, which this is.

schlafly:
> **v1_calculate has v1_calculate_from_models, v1_calculate_over_time, simplify table: do we need these?** I would say yes because these represent the programmatic api that implements the command-line script

Your call, but I don't think there's anything wrong with the script calling private functions.  simplify_table was the one in particular that seemed to me like if we wanted to implement something a different way that didn't need simplify table, we should be able to do that internally without worrying about external users.

PR #2316, opened 2026-05-08
Updated timing pattern

t-brandt:
Closes #

This PR improves the timing pattern for the detectors, primarily for the purpose of making the WFI18 transient correction more accurate.  The approach implemented is almost as accurate as using the guide window MA tables, but avoids having to read these reference files in for now.  There is a fudge factor of 1.5 passed to the timing pattern to account for the fact that the WFI18 transient decays slightly more quickly when reading out the guide window.

## Tasks


schlafly:
Thanks Tim.  There's a missing definition of npad that is breaking the tests.  The logic here has gotten involved enough that some testing probably makes sense but I'll defer to you, thanks.

schlafly:
Running regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/25677644827 .  They show some minor differences ~everywhere due to the dark decay step being an early step, and this timing pattern affecting the dark decay step.

schlafly:
This has minor impacts across a lot of regtest files, that get amplified by bad linearity corrections.  The only ones that are actually noticeable are on WFI18, which has the effect that depends sensitively on the timing pattern.  There we see the expected 256 pixel pattern.  Approving and merging.  
image

schlafly (approved):

t-brandt:
Thanks, Eddie.  For posterity, I am adding a couple of notes:
The definition of npad is there; it is hardcoded as (to my knowledge) it is not possible to change.
Regression test failures were from two issues: (1) not passing a frame time in tests in wfi18_transient; and (2) using a frame time shorter than is possible given that the pixel time is hardcoded in the new timing pattern.  Both are fixed with minor changes to those unit tests.

PR #2318, opened 2026-05-11
RCAL-1360: Skip Fluxless J-Band SI Objects

### LINKED ISSUE #2231: Skip Fluxless J-Band SI Objects
_Issue [RCAL-1360](https://jira.stsci.edu/browse/RCAL-1360) was created on JIRA by [Paul Huwe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=phuwe):_

Within source injection, the J-Band is used to select the faint magnitudes for injected sources. A small percentage of objects can be selected with no J-Band flux. Accordingly, the injection step should be adjusted to remove such objects from the selection sample.

PaulHuwe:
Resolves [RCAL-1360](https://jira.stsci.edu/browse/RCAL-1360)

Closes #2231

This PR ensures no entries with zero flux in the J-Band are used to create the source injection catalog.

## Tasks


PaulHuwe:
Passing regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/25673853455

romancal/source_catalog/tests/test_injection.py
```diff
@@ -224,8 +216,14 @@ def test_create_cosmoscat():
...
 
+    # Set multiplier of objects to create enough artificial objects
+    # to test that 0 J-band flux objects are removed
+    obj_fact = 100
+
     # Convert x,y to ra, dec
-    ra, dec = wcsobj.pixel_to_world_values(np.array(XPOS_IDX), np.array(YPOS_IDX))
+    ra, dec = wcsobj.pixel_to_world_values(
+        np.array(obj_fact * XPOS_IDX), np.array(obj_fact * YPOS_IDX)
+    )
```
schlafly: The approach here is that you generate 100x more sources at identical locations and hope that at least one of them has 0 J flux, and then test that none of them have 0 J flux at the end?  Then remove the stuff you added by taking items -1 through 3?  This feels weird?

Would this be better as a new test that only makes a lot of objects and checks that no fluxes are <0, rather than adding & removing these objects?

schlafly (commented):
The motivation here is that J is used for the galaxy sizes and so it's particularly important that J fluxes be non-zero?

PaulHuwe:
> The motivation here is that J is used for the galaxy sizes and so it's particularly important that J fluxes be non-zero?

Yes.

romancal/source_catalog/tests/test_injection.py
PaulHuwe: No need to hope - we pass a seed all the way through the calls to RSIM which determines the order of the COSMOS galaxy sample. At that size, there are four 0 J flux objects (iirc). 

I have no issue breaking it out into its own test, without the test input manipulation.

schlafly: Let's break it into a separate test.

PaulHuwe: Refactored.

schlafly:
Can you fix the failing test?

schlafly (approved):

PR #2322, opened 2026-05-11
Add versioning page to docs (and minor stpipe submodule cleanup)

braingram:
Let me know if it's preferred to split the docs and minor stpipe change (renaming `romancal.stpipe.utilities` to `_utilities`).

Regtests all pass: https://github.com/spacetelescope/RegressionTests/actions/runs/25690927281

docs with the added page:
https://roman-pipeline--2322.org.readthedocs.build/en/2322/roman/versioning.html

I wasn't sure where to add the page so I nested it under "Package Index".

Here is the jwst page which was used as a model: https://jwst-pipeline.readthedocs.io/en/stable/jwst/user_documentation/more_information.html#versioning-scheme

I left out the "deprecation policy" section as we have none. We could consider adding one but I'm inclined to wait until we see how the versioning scheme is to work with.

## Tasks


schlafly:
No need to split the PR, looks reasonable to me.

schlafly (approved):
Perfect, thanks!  Leaving out a deprecation policy makes sense to me for now, though I'm also happy to invent one.  If I were to invent one, I'd make it something like:
Deprecated features will continue to exist for at least six months following deprecation, but may then be removed.
But I'm making that up and I am happy to leave it unspecified for now, thanks.

PR #2324, opened 2026-05-12
Update resampled output image's background level and status

mcara:
When attempting to resample (again) an already resampled image, the resulting double-resampled image often has an unexplained background offset compared to expected (single-resampled) image. After analysis, this is caused by resample populating most meta of the output resampled image from an input cal image and not updating background-related fields.

See https://github.com/spacetelescope/jwst/pull/10525 and https://github.com/spacetelescope/stcal/pull/542 for more details/conversations.

Regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/25767553367

## Tasks


mcara:
Unfortunately the code that I intended to update initially:
```python
        output_model.meta.background.level = self.output_model.get("level", None)
        output_model.meta.background.subtracted = self.output_model.get(
            "subtracted",
            False
        )
```
does not work here, probably due to roman datamodels not storing this info for the resampled image. 

Instead I updated how level is set before passing these data to resample when an input model does not have background level computed. This is to make it more consistent with changes in `stcal` and `jwst`. It should not affect any tests.

schlafly:
I've forgotten the general flow here.  Let me try to recap, and you can correct me?

skymatch is responsible for setting the background metadata levels.  It has a handful of options.  It can either match the sky levels to one another, leading the matched images, after subtracting the levels, to all have the same non-zero sky background, or it can zero the sky, leading the matched images to all have zero sky background after subtraction.  These levels are recorded in 'level' in the individual metadata.  It can also choose whether these levels are subtracted (subtracted = True) or not.  If not, it's up to the outlier detection or resample step to subtract these levels before comparing pixels or coadding.

In that context, I did not think that Roman's coadd images had background metadata, unless they go through sky match.  I don't remember if that leads resample to fail or just to skip any sky subtraction.  In principle, I could imagine that you could pass resampled images through both sky match and outlier and resample, and that could work.  In that case sky match would again populate the relevant metadata and conceptually sensible values would be populated and used.

I think this PR is trying to get the right metadata without going through skymatch.  For the case where skymatch has zeroed the sky, I think that makes sense conceptually; since the sky has been zeroed in each resampled image, you want to skip the background subtraction step, and you certainly don't want background metadata from the first input image (?) being treated as the sky background on the resampled image.  For the case where skymatch has matched the sky between the input images, but intentionally left on some amount of average sky, it's less clear to me what you want; you want to know those average sky levels for each input resampled image, and put in new metadata that adjusts them so that they're consistent.  Running skymatch on the resampled images would achieve that.

I take the associated Webb PR as trying to fix the ~bug that the background metadata from the first input exposure is used as the background metadata for the resampled image.  I may gather from this PR that Roman doesn't have the same issue, and so doesn't really have an associated bug?  It looks to me like this PR changes what happens for an image that doesn't have background metadata, which used to default to 'I have been background subtracted, by zero' and is now 'I have not been background subtracted, by None.'  In either case this triggers a no-op here at present, but you want the romancal default to follow the Webb default.  I'm okay with that.
https://github.com/spacetelescope/stcal/blob/main/src/stcal/resample/resample.py#L820-L821

I checked one example Roman coadd and it did not contain any offending 'background' metadata, so I think that for Roman this default block would fire and no sky subtraction would be done.  If that's right, there's no bug in romancal, and this PR only aims to change what the 'don't do a correction' ~placeholder values are.  Is that right?

For romancal we build a table with the values of the backgrounds for each input image and store it in the coadd.  The former default of 0 was slightly nice in that that table has integers; with the new default of None that table has strings.  But the new version more clearly signifies something went wrong so I'll go along with it.

Can you confirm that romancal isn't affected by the linked bug and that this is just a 'don't do anything' flag update?

mcara:
I am somewhat confused about this statement:

> In that context, I did not think that Roman's coadd images had background metadata, unless they go through sky match. I don't remember if that leads resample to fail or just to skip any sky subtraction. In principle, I could imagine that you could pass resampled images through both sky match and outlier and resample, and that could work. In that case sky match would again populate the relevant metadata and conceptually sensible values would be populated and used.

1. _"Roman's coadd images had background metadata, unless they go through sky match"_ I could not find where in the code is the background level of the _resampled image_ (coadds?) is set. Maybe I am missing something but **I think it is never set**. That's OK. Unfortunately JWST datamodels always have these metadata and so my JWST PR is trying to fix that.
2. This PR as it is now (after 3 commits), is simply changing how missing sky level data in *input* images are interpreted and passed to `stcal.resample`: Should we assign level 0 and say that we already subtracted this level OR should we assign `None` (meaning we never computed it) and that this non-existent level was not subtracted. Functionally, `stcal.resample` treats both cases the same: both `level==None` or `subtraced is True`. So this PR does not change the behavior and only "semantics".

In my [first commit in this PR](https://github.com/spacetelescope/romancal/pull/2324/changes/d365a67d3aab2f4db1c12a824b40a29a3a7e7a9b) I tried to set backgound level of resampled images but it was failing because `meta` does not have `background` attribute. So I removed what I thought would be a "fix" but if it is never set for Roman, then it is not a bug and no fix is needed.

> I take the associated Webb PR as trying to fix the ~bug that the background metadata from the first input exposure is used as the background metadata for the resampled image. I may gather from this PR that Roman doesn't have the same issue, and so doesn't really have an associated bug? It looks to me like this PR changes what happens for an image that doesn't have background metadata, which used to default to 'I have been background subtracted, by zero' and is now 'I have not been background subtracted, by None.' In either case this triggers a no-op here at present, but you want the romancal default to follow the Webb default. I'm okay with that.
https://github.com/spacetelescope/stcal/blob/main/src/stcal/resample/resample.py#L820-L821

> I checked one example Roman coadd and it did not contain any offending 'background' metadata, so I think that for Roman this default block would fire and no sky subtraction would be done. If that's right, there's no bug in romancal, and this PR only aims to change what the 'don't do a correction' ~placeholder values are. Is that right?

That's _exactly_ right!

> Can you confirm that romancal isn't affected by the linked bug and that this is just a 'don't do anything' flag update?

This PR is just a _"'don't do anything' flag update"_. Based on my analysis, it looks like Roman does not assign background level to resampled datamodels (but I am not 100% sure) and so, IMO, Roman resampled images are not affected by this JWST bug.

schlafly:
> 1. _"Roman's coadd images had background metadata, unless they go through sky match"_ I could not find where in the code is the background level of the _resampled image_ (coadds?) is set. Maybe I am missing something but **I think it is never set**. That's OK. Unfortunately JWST datamodels always have these metadata and so my JWST PR is trying to fix that.

I agree, we don't set that; only sky match sets background levels, so if you want background levels in coadds, you would have to pass the coadds through sky match.  We never actually do this because we haven't explored coadding coadds, but conceptually it could work.  I kind of think this is just right and that the Webb code is "wrong"; no background metadata should exist in the coadds unless they have gone through a sky matching process, IMO.  For the Webb case I would worried about what happens when skymatch is run in a mode where it is not trying to set the sky background to zero.  It seems likely to me that your two resampled images will have different sky levels in that case and won't coadd nicely.

> This PR is just a _"'don't do anything' flag update"_. Based on my analysis, it looks like Roman does not assign background level to resampled datamodels (but I am not 100% sure) and so, IMO, Roman resampled images are not affected by this JWST bug.

Good, we're on the same page there, I was confused by the linked bug report.

schlafly:
@mcara , I'm happy to merge this when / if you think the discussion of the right defaults on Webb have resolved.  The current Roman defaults (subtracted = True, level = 0) doubly turn off any additional subtraction behavior, so I think they're "fine" but I agree that lining up with Webb is preferable.  I'm not going to consider this something we need to get into B22 though (I'm hoping to green light that today).

mairanteodoro (commented):
Looks good to me! Thanks, @mcara!

mcara:
can this be merged?

schlafly (approved):

PR #2325, opened 2026-05-13
Make source cataloging only use DO_NOT_USE and ignore other DQ flags

schlafly:
This PR makes the source catalog step ignore all bits DQ bits other than DO_NOT_USE.  Old CRDS reference files contained substantial numbers of problematic bits that were not flagged.  New reference files should address this issue so that only DO_NOT_USE pixels are in fact problematic.

On the way I also changed some test code to use GW_AFFECTED_DATA rather than PERSISTENCE, since PERSISTENCE may change names.  I also changed a doc NON_LINEAR reference to NONLINEAR, since the latter is the actual name of the pixel in question.

Regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/25806174508
They show small improvements to the Gaia astrometric RMS, a pleasant surprise; presumably some of the formerly flagged pixels are now being used.  I am starting a fresh run to bring in the dust map updates.

## Tasks

### DIFF
--- a/changes/2325.source_catalog.rst
+++ b/changes/2325.source_catalog.rst
@@ -1,0 +1,1 @@
+Source catalog DQ masking now excludes only pixels with DO_NOT_USE set, rather than any set DQ bit.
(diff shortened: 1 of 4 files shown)


mairanteodoro (approved):
Looks good to me. Thanks, Eddie!

PR #2327, opened 2026-05-13
Update handling of different exposure types.

### LINKED ISSUE #2311: WFI_LOLO data is supposed to process all the way through L2?
Tyler just told me that we should include WFI_LOLO here?  @tddesjardins 

https://github.com/spacetelescope/romancal/blob/a15bfab0fa1165d007369e640312c40dd5daf69b/romancal/pipeline/exposure_pipeline.py#L145

Tyler, it would be valuable to run down each of the exposure types and how far we should process them.
  - WFI_IMAGE
  - WFI_SPECTRAL
  - WFI_IM_DARK
  - WFI_SP_DARK
  - WFI_FLAT
  - WFI_LOLO
  - WFI_WFSC

schlafly:
Closes #2311 .

This PR aligns the handling of different exposure types with the prescription of #2311.  The new goal is:
- all images get processed, at least including all steps before fielding
- image, lolo, wfsc images proceed up to before source cataloging
- only images get source cataloging and tweakreging

It also changes tweakreg so that anything that has a source catalog gets tweakreg'd, rather than requiring images.  Since tweakreg doesn't actually look at the images this seems sensible, and the old spectral exposures won't have source catalog so the old behavior is retained with one fewer check.

It also changes the behavior of a few steps when run individually, which formerly checked to make sure that they were being run on the right type of image, and now just try to proceed.  The rationale here is that if the user specifically asks to run a step on an image, we should do it (and potentially crash and burn due to a missing reference file, e.g., when flattening a dark).

I took the opportunity to simplify the handling of exposures including an SCA which was all saturated.  The all saturated SCA skips processing as usual, but the other SCAs go through as normal and get tweakreg'd as usual.  This makes more sense to me and made for simpler code.

I updated the corresponding docs and removed a few tests that no longer made sense (mostly checking that we bailed out in the right way when running an individual step on a file of the wrong type, which we are now doing when the user explicitly requests it).

#2311 includes a request to run the photom step on spectroscopic images, which still don't get flat fielding.  To make the flow clean, I now moved photom ahead of flat fielding.  photom doesn't touch the pixels so these steps commute.  I updated the docs correspondingly.

Regression tests are here:
https://github.com/spacetelescope/RegressionTests/actions/runs/25826379623/job/75880473052
and show small metadata differences reflecting the fact that now the photom step is performed on spectral data, filling out the values from the photom reference file.  That means that the step is now marked complete, metadata contains a reference to the CRDS file name, and the metadata values from the photom file (conversions to megajanskies, pixel areas) are now filled out to the photom file value (None for spectral cases).

## Tasks


schlafly:
@braingram and @ddavis-stsci , mind taking a look at this one?  Brett, there was some refactoring of the main exposure_pipeline path that maybe you'd like to see, including changing the handling of the annoying all saturated case.  @ddavis-stsci , this changes some policies for different steps that go back a long time (e.g., allowing flat field to run on spectroscopic data if the user explicitly runs that) that you might want to look at.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -111,19 +111,14 @@ def process(self, dataset):
         imcats = []
         with images:
             for i, image_model in enumerate(images):
-                exposure_type = image_model.meta.exposure.type
-                if exposure_type != "WFI_IMAGE":
-                    log.info("Skipping TweakReg for spectral exposure.")
+                source_catalog = getattr(image_model.meta, "source_catalog", None)
```
braingram, line 114: I think this should work with the changes in https://github.com/spacetelescope/romancal/pull/2265 (which we should get into this build since that fixes the custom catalog feature).

braingram (approved):
Code changes and regtest differences look good to me.

The exptype conditional logic in the elp looks like it would be better handled by:
- pars files for the different exptypes, skipping the appropriate steps
- elp calling `run` on all steps (where the pars file will configure stpipe to skip the steps where so configured).

Given the upcoming release and the need for new pars files this would be a reasonable temporary measure until pars files can be made.

schlafly:
> The exptype conditional logic in the elp looks like it would be better handled by:
> 
> * pars files for the different exptypes, skipping the appropriate steps
> * elp calling `run` on all steps (where the pars file will configure stpipe to skip the steps where so configured).
> 
> Given the upcoming release and the need for new pars files this would be a reasonable temporary measure until pars files can be made.

Thanks, agreed on both counts.

mairanteodoro (approved):
Looks good to me! Thanks, Eddie!

PR #2329, opened 2026-05-14
RCAL-1396 no authentication header if no token and velocity checking

stscieisenhamer:
Resolves [RCAL-1396](https://jira.stsci.edu/browse/RCAL-1396)

This PR addresses two bugs:
- Using MAST, the authentication header was generated whether there is a token or not
- Observatory velocity check too strict and inconsistent

Fixes include:
- If no token, no authentication is generated
- Velocity constraint loosened and check is more consistent.

## Tasks

### DIFF
--- a/romancal/orientation/_lib.py
+++ b/romancal/orientation/_lib.py
@@ -23,9 +23,9 @@ FGS_DEFAULT_QUATERNION = np.array(
     [-0.18596734175399293, 0.6837984564491885, -0.1800546332580956, 0.6822141509826322]
 )
 
-# Maximum absolute speed of the observatory. Used for sanity check is defined
-# as the sum of the absolute components of the velocity.
-MAX_OBSERVATORY_SPEED = 150
+# Maximum absolute speed of the observatory. Used for sanity check for any
+# individual component of the velocity.
+MAX_OBSERVATORY_SPEED = 500  # km/s
 
 # Common conversions.
 # Yes, these are available in a number of packages. However, they are explicitly defined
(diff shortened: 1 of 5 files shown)


stscieisenhamer:
[regression test #1](https://github.com/spacetelescope/RegressionTests/actions/runs/25839728475)
result: clean

romancal/orientation/_lib.py
```diff
@@ -23,9 +23,9 @@
     [-0.18596734175399293, 0.6837984564491885, -0.1800546332580956, 0.6822141509826322]
 )
 
-# Maximum absolute speed of the observatory. Used for sanity check is defined
-# as the sum of the absolute components of the velocity.
-MAX_OBSERVATORY_SPEED = 150
+# Maximum absolute speed of the observatory. Used for sanity check for any
+# individual component of the velocity.
+MAX_OBSERVATORY_SPEED = 500  # km/s
```
schlafly, line 28: I just have to ask---earth around the sun is ~30 km / s, satellite in low earth orbit is ~10 km / s.  My sense is that the typical velocity on the way to L2 is ~1 km / s.  When is a velocity of 150 km / s or 500 km / s valid?

schlafly (approved):

romancal/orientation/_lib.py
stscieisenhamer: Never, but this is more a check against the default value, with is like 9999. Did not want to be exact, originally had the 150, but a recent MRT (or just some other recent data) produced data that had ~250, which isn't right but should otherwise have processed anyways. So just something really to keep the "obviously silly" out of the picture.

PR #2342, opened 2026-05-22
Refactor tweakreg's logic for flagging step SKIPPED, COMPLETE, or FAILED.

### LINKED ISSUE #1677: Record when tweakreg fails?
We currently record "complete" for the tweakreg step regardless of whether the step was successful.
https://github.com/spacetelescope/romancal/blob/dad22703b046ecbaac63c0f2f67e4a2d0277f63e/romancal/tweakreg/tweakreg_step.py#L232-L238
There's some indication as to whether the step was successful in the presence or absence of the wcs_fit_results metadata, though that will not be an obvious signal to most people.

It's unclear to me what the right signal should be.  We don't have a cal_step status that is "FAILED", though we could conceptually add it.  We could leave the step as "incomplete" or set it to "skipped", though usually those mean that we haven't run the step yet or that we intentionally skipped it, not that we tried to run the step but failed for some reason.

I encountered this in the context of the flatsat data, which contains test pattern data.  We find some sources and try to match them to Gaia, and fail to get many matches, given the fact that the test pattern data has no relationship to the sky.  Then tweakreg says "couldn't match Gaia, step complete", and everything proceeds.  In some way this is "good"---for the current tests they want things to run through to exercise the system, even though the data is garbage and doesn't make much sense.  In another sense this scares me a lot, in that I don't want us to process weeks' worth of data and then learn that we had accidentally been failing tweakreg in every one but not noticing it.

Thoughts on the correct behavior here?

mairanteodoro:
Resolves [RCAL-1379](https://jira.stsci.edu/browse/RCAL-1379)

Closes #1677 

This PR updates tweakreg failure signaling and fit-result metadata so attempted-but-unsuccessful runs are clearly visible and diagnosable. It addresses the behavior discussed in #1677 ([RCAL-1379](https://jira.stsci.edu/browse/RCAL-1379)): tweakreg could report `COMPLETE` even when no successful WCS update occurred, and `wcs_fit_results` could be missing in failure paths.

The tweakreg step was update to account for the following:
- refactored finalize logic to set step status by per-model fit outcome:
    - `SKIPPED`: model was not attempted (e.g., no source catalog),
    - `COMPLETE`: successful fit and WCS update,
    - `FAILED`: attempted but no successful WCS solution.
- removed early-return behavior that left attempted models in implicit `INCOMPLETE` on absolute-alignment errors.
- added unified fit-result serialization for attempted models:
    - always populates `meta.wcs_fit_results`,
    - preserves available `nmatches` (falls back to NaN if unavailable),
    - adds `n_detector` (number of detector images used in the solve),
    - normalizes `numpy` types for metadata safety,
    - removes internal/non-user-facing fit keys defensively.

The relevant docs and unit tests were also updated accordingly.

### Regression tests
- https://github.com/spacetelescope/RegressionTests/actions/runs/26311853561 (05/22/26)

## Tasks

### DIFF
--- a/romancal/tweakreg/tweakreg_step.py
+++ b/romancal/tweakreg/tweakreg_step.py
@@ -580,7 +580,7 @@ def _serialize_wcs_fit_results(
         fit_results["status"] = "FAILED"
 
     if fit_results.get("nmatches") is None:
-        fit_results["nmatches"] = np.nan
+        fit_results["nmatches"] = 0
 
     fit_results["n_detector"] = n_detector
     return fit_results
--- a/romancal/tweakreg/tweakreg_step.py
+++ b/romancal/tweakreg/tweakreg_step.py
@@ -177,6 +177,8 @@ class TweakRegStep(RomanStep):
 
         # run alignment only if it was possible to build image catalogs
         if len(imcats):
+            absolute_alignment_failed = False
+
             # extract WCS correctors to use for image alignment
             if len(images.group_indices) > 1:
                 try:
@@ -188,18 +190,31 @@ class TweakRegStep(RomanStep):
                 self.do_absolute_alignment(ref_image, imcats)
             except TweakregError as e:
                 log.warning(str(e))
-                return images
+                absolute_alignment_failed = True
 
             # finalize step
             with images:
                 for imcat in imcats:
                     image_model = images.borrow(imcat.meta["model_index"])
-                    image_model.meta.cal_step.tweakreg = "COMPLETE"
+                    fit_info = imcat.meta.get("fit_info")
+                    fit_status = "" if fit_info is None else str(fit_info.get("status", ""))
+                    fit_succeeded = (
+                        not absolute_alignment_failed and "SUCCESS" in fit_status
+                    )
+
+                    image_model.meta["wcs_fit_results"] = _serialize_wcs_fit_results(
+                        fit_info=fit_info,
+                        n_detector=len(imcats),
+                        force_failed_status=absolute_alignment_failed,
+                    )
+
                     # remove source catalog
-                    del image_model.meta["tweakreg_catalog"]
+                    if "tweakreg_catalog" in image_model.meta:
+                        del image_model.meta["tweakreg_catalog"]
 
                     # retrieve fit status and update wcs if fit is successful:
-                    if "SUCCESS" in imcat.meta.get("fit_info")["status"]:
+                    if fit_succeeded:
+                        image_model.meta.cal_step.tweakreg = "COMPLETE"
                         # Update/create the WCS .name attribute with information
                         # on this astrometric fit as the only record that it was
                         # successful:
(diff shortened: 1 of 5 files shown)


romancal/tweakreg/tweakreg_step.py
```diff
@@ -542,6 +537,66 @@ def _validate_catalog_columns(catalog) -> bool:
...
+    ]:
+        fit_results.pop(key, None)
+
+    if force_failed_status:
+        fit_results["status"] = "FAILED"
+    elif not fit_results.get("status"):
+        fit_results["status"] = "FAILED"
+
+    if fit_results.get("nmatches") is None:
+        fit_results["nmatches"] = np.nan
```
schlafly: Let's do -1 or 0 here since it feels like nmatches should be an integer.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -542,6 +537,66 @@ def _validate_catalog_columns(catalog) -> bool:
...
+
+    fit_results["n_detector"] = n_detector
+    return fit_results
+
+
+def _to_python_scalar_or_list(value):
```
schlafly, line 589: I hadn't thought about this.  Yeah, there are some rotations / shifts / matrices in here that probably get serialized by default as numpy arrays.  That is a little annoying in that it probably triggers extra binary extensions for very small arrays.  OTOH, it's not something that I had ever noticed until thinking about it now.  What problem case are you trying to solve here?

romancal/tweakreg/tweakreg_step.py
```diff
@@ -542,6 +537,66 @@ def _validate_catalog_columns(catalog) -> bool:
...
+        fit_results = {k: _to_python_scalar_or_list(v) for k, v in fit_info.items()}
+
+    # remove unwanted keys from WCS fit results
+    for key in [
+        "eff_minobj",
+        "matched_ref_idx",
+        "matched_input_idx",
+        "fit_RA",
+        "fit_DEC",
+        "fitmask",
```
schlafly, line 573: Are these fields in our input today?  What triggers them to exist, such that they need to be removed?  I tend to agree that we don't want them by default.

schlafly (commented):
Looks good.  A few minor comments inline.  Thanks!

schlafly:
Regtests show only new n_detector field; good.  https://github.com/spacetelescope/RegressionTests/actions/runs/26311853561/job/77895494812

romancal/tweakreg/tweakreg_step.py
mairanteodoro: Everything in `fit_info` is provided by the tweakwcs package after a successful alignment:
https://github.com/spacetelescope/tweakwcs/blob/0cccdd62230e755d484e1eac043aa329cf19b5be/tweakwcs/wcsimage.py#L1379-L1398

romancal/tweakreg/tweakreg_step.py
mairanteodoro: Done; using 0.

romancal/tweakreg/tweakreg_step.py
mairanteodoro: No specific case, I just thought it'd be better to make sure we're preventing serialization issues with non-native types in the metadata (as required by ASDF).

romancal/tweakreg/tweakreg_step.py
schlafly: Sorry, I had missed that this was a code refactor rather than something new.

schlafly (approved):

PR #2345, opened 2026-05-26
Save cat/segm files in ELP only if requested by self.save_results.

mairanteodoro:
Resolves [RCAL-1139](https://jira.stsci.edu/browse/RCAL-1139)

Closes #

This PR makes `ExposurePipeline` honor `save_results` for source-catalog sidecar products `*_cat.parquet` and `*_segm.asdf`. These files are now created only when `save_results = True`.

### What changed

- Updated `romancal/source_catalog/_save_utils.py` so cat/segm sidecar files are only written when save output is enabled;
    - added a sidecar-save gate via `save_catalog_results` (defaults to `save_results`);
    - segm is now saved only when that gate is `True`;
    - explicit cat save for `return_updated_model = True` is also gated;
    - when not saving sidecars and `return_updated_model = True`, the step now stores an in-memory catalog at `meta.source_catalog.tweakreg_catalog` (instead of a filename), so tweakreg can still consume it without on-disk cat;
- updated `romancal/pipeline/exposure_pipeline.py` to keep pipeline behavior clean:
    - `self.source_catalog.return_updated_model = True` (existing behavior).
    - `self.source_catalog.save_results = False` to avoid writing `*_sourcecatalog.asdf`.
    - `self.source_catalog.save_catalog_results = self.save_results` so cat/segm follow pipeline `save_results`.

### Tests updated

- `romancal/pipeline/tests/test_exposure_pipeline.py`
  - `test_elp_save_results` now expects:
    - `save_results = True`: `test_cal.asdf`, `test_wcs.asdf`, `test_cat.parquet`, `test_segm.asdf`
    - `save_results = False`: no outputs in the output dir.
- `romancal/source_catalog/tests/test_source_catalog.py`
  - updated `test_l2_source_catalog_keywords` and `test_l3_source_catalog_keywords` expectations for `save_results = False` to no cat/segm outputs.
  - added explicit assertions that `*_cat.parquet` and `*_segm.asdf` existence matches `save_results`.
  - added `test_l2_return_updated_model_uses_in_memory_catalog_when_not_saving` to verify:
    - in-memory `tweakreg_catalog` is populated,
    - no `tweakreg_catalog_name`,
    - and no cat/segm files are created.

## Tasks

### DIFF
--- a/romancal/source_catalog/_save_utils.py
+++ b/romancal/source_catalog/_save_utils.py
@@ -147,7 +147,9 @@ def save_all_results(
 
     # save the segmentation image only when results are requested
     if save_catalog_results:
-        save_segment_image(self, segment_img, cat_model, output_filename, save_debug_info)
+        save_segment_image(
+            self, segment_img, cat_model, output_filename, save_debug_info
+        )
 
     # Update the source catalog filename metadata
     self.output_ext = "parquet"
--- a/romancal/source_catalog/_save_utils.py
+++ b/romancal/source_catalog/_save_utils.py
@@ -102,7 +102,7 @@ def save_all_results(
     """
     Return and save the results of the source catalog step.
 
-    The segmentation image is always saved.
+    The segmentation image is saved only when ``save_results`` is True.
 
     If ``save_results`` is True, then either the input model with
     updated metadata or the source catalog model is saved/returned
@@ -143,20 +143,22 @@ def save_all_results(
     output_filename = (
         self.output_file if self.output_file is not None else cat_model.meta.filename
     )
+    save_catalog_results = getattr(self, "save_catalog_results", self.save_results)
 
-    # always save the segmentation image
-    save_segment_image(self, segment_img, cat_model, output_filename, save_debug_info)
+    # save the segmentation image only when results are requested
+    if save_catalog_results:
+        save_segment_image(self, segment_img, cat_model, output_filename, save_debug_info)
 
     # Update the source catalog filename metadata
     self.output_ext = "parquet"
     output_catalog_name = self.make_output_path(basepath=output_filename, suffix="cat")
     self.output_ext = "asdf"
 
-    # Always save the source catalog, but don't save it twice.
-    # If save_results=False or return_updated_model=True, we need to
-    # explicitly save it.
+    # Explicitly save the source catalog only when sidecar saving is enabled and
+    # return_updated_model=True. In this case the returned object is the
+    # input model, so we must write the catalog product here.
     return_updated_model = getattr(self, "return_updated_model", False)
-    if not self.save_results or return_updated_model:
+    if save_catalog_results and return_updated_model:
         self.output_ext = "parquet"
         self.save_model(
             cat_model,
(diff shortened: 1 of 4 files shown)


schlafly (commented):
I don't completely follow this PR.  The most important question is where does the catalog go when save_results is not True?  I also don't have a good picture in my head of what a sidecar is.

I think there's a world where this makes sense, but it seems to me to need to be part of a somewhat larger refactoring.  In that world:
- source catalog step always returns a source catalog and a segmentation image
- stpipe is responsible for writing those
- I think the source catalog step needs to get some new routine to teach stpipe how to write the (source catalog, segmentation map) tuple it returns.  I only remember that one of the jwst steps has an example of doing  that but I don't remember which one.
- ELP would be responsible for writing them or otherwise communicating them to tweakreg, and updating the source catalog file name, getting rid of return_updated_model

mairanteodoro:
When `save_results=False` in this PR, the catalog is not written to disk. In the ELP, it is kept in-memory on the model as `meta.source_catalog.tweakreg_catalog` so tweakreg can consume it in the same run. The segmentation map is similarly not persisted.  

When `save_results=True`, we write `*_cat.parquet` and `*_segm.asdf` and set `meta.source_catalog.tweakreg_catalog_name` to the catalog path.

With respect to the wording “sidecar,” I meant “auxiliary output files” written alongside `*_cal.asdf` (the catalog + segmentation outputs).

I also agree with your larger design direction:
•  `SourceCatalogStep` returns catalog + segmentation result data;
•  `stpipe` handles writing that return type;
•  ELP handles passing data to tweakreg and we remove `return_updated_model`.

I scoped this PR narrowly to fix the immediate behavior bug (cat/segm honoring `save_results`) with minimal changes to the API. I’m fine with either opening a follow-up refactor issue/PR for the broader cleanup you described or just holding off on this PR.

romancal/source_catalog/_save_utils.py
```diff
@@ -177,11 +181,16 @@ def save_all_results(
...
-                "tweakreg_catalog_name": output_catalog_name
-            }
+            if save_catalog_results:
+                input_model.meta.source_catalog = {
+                    "tweakreg_catalog_name": output_catalog_name
+                }
+            else:
+                input_model.meta.source_catalog = {
+                    "tweakreg_catalog": cat_model.source_catalog.as_array()
+                }
```
schlafly, line 193: Thanks, I missed this.

schlafly (commented):
Thanks, I missed adding the catalog to model.meta.source_catalog.  But we still drop the segmentation map on the ground?

Adding the source catalog to the metadata was the old behavior.  I think that's a bad approach; these catalogs can be quite large and really aren't really metadata.  The as_array() bit will also lose metadata that we usually keep, introducing a new 'kind' of source catalog.

I think there's a way to do this, but we always want the step to produce a source catalog and the associated segmentation map, and we don't want to stuff them into the images.  For an ELP run I don't know that we have any alternative to writing them out at present; I think that source_catalog bit is the only existing programmatic approach we have to do this (but you would know better).  So we would need tweakreg to fail or something when save_results is False, or add a new mechanism into tweakreg that lets us pass the catalogs.  To be clear, we don't need that in ops today so I don't think we should add that new mechanism, and an ELP run should probably have a default of source_catalog.save_results = True or otherwise raise an issue if tweakreg isn't skipped and it's not asked to make a catalog or something.

I don't immediately see what the whole path would look like there.  If you think you're on the trail and can scope it out though, great.

mairanteodoro:
I agree the bigger cleanup is the right long-term direction. Nevertheless, for this PR, we could focus on the requested behavior correctness, which so far is that '_"cat" and "segm" files are only created if `save_results=True`_'. So we could revert this PR to no in-memory catalog stuffing into `model.meta.source_catalog`, keeping tweakreg input file-based (tweakreg_catalog_name) only. In an ELP run, when tweakreg is enabled, require an on-disk source catalog (i.e., ensure `source_catalog.save_results = True`, or raise a clear error if not). Or, we can just hold off on this PR until we have a clear picture of the requirements. What do you think?

schlafly:
I think we should delay until a more complete cleanup is possible.  If we revert the model.meta.source_catalog part of this PR, then source_catalog.save_results = False means "do the source catalog step, and then throw the results away" which is too weird of a behavior.

PR #2352, opened 2026-06-04
Fix save_results handling (remove custom source catalog saving)

### LINKED ISSUE #1960: elp runs with save_results disabled produces segm and cat files
From: https://github.com/spacetelescope/romancal/pull/1959#discussion_r2355419374 in response to "would you consider this a bug?"
> I hadn't, but I agree that's another case where we're doing something that isn't a good match to the stpipe assumptions.
>
> Some background: I see the role of the source catalog step as being to produce source catalogs and ancillary data products that contain things like the segmentation map. In that context, maybe the least surprising approach would be for the source catalog step to return a tuple of catalog-related products and have some special saving routines to know how to save them if requested; you pointed me to related code in a Webb step once.
>
> The current pipeline has a single input - single output flow that doesn't really work well with this. But you could imagine the ELP gathering the catalogs and models and then sending them both as arguments to the tweakreg step. Instead the current design is that the catalog step writes out catalog files and updates the images to contain references to those catalog files, which tweakreg reads back in. That doesn't really work well with the save_results concept but is reasonable behavior for the pipeline, insofar as it produces the files we need and contains references to them so that we know what was used in tweakreg. A former pipeline flow put the catalogs in the image metadata, which worked well with the pipeline flow but is a bit weird in that the metadata becomes very large; most users would not expect to find catalogs "inside" an image; and people who want only catalogs should not need to download the associated images.
>
> So since I don't really see a world where we don't want to save the catalogs, I have not seen this as a bug. However, it does seem like we could build a catalog pipeline in which one block would be the current catalog step, and later blocks might do things like the photometric redshifts, etc.. The current behavior would not support that kind of composition well. We currently control that kind of behavior via arguments like --fit_psf, but maybe the suggestion is that in a better designed pipeline that could be a series of steps which would enable better configurability from stpipe or swapping in & out user steps. I see that more as a theoretical statement than something where I actually see a lot of action happening, though, and would tend to not make changes here.

This issue can be used to discuss options. Ideally a solution that is compatible with stpipe/strun parameters (like `save_results`) which are well documented and standard across the pipelines and steps.

braingram:
Fixes https://github.com/spacetelescope/romancal/issues/1960

Fix save_results for source_catalog/multiband_catalog/elp/mosaic pipeline. On main all of these (if the source_catalog step is run) always write out `cat` and `segm` files. With this PR those steps only write out results if save_results=True.

To accomplish this the source_catalog and multiband_catalog steps were modified to return a tuple of (catalog_model, segmentation_map_model). The pipelines were updated to accommodate this new return value and
handle updating the processed model and catalogs.

During an elp run, tweakreg no longer modifies the provided catalog and instead the catalog is updated in the elp (so the result is the same). The reason for this is that:
- on main the catalog update required that a file on disk existed
- the catalog was round-tripped via parquet bypassing validation

This PR also provides a few benefits to python usage of romancal where in addition to respecting save_results:
- SourceCatalogStep/MultibandCatalogStep.call now returns the catalog and the segmentation map model
- elp.call now returns the catalogs and segmentation map model (in addition to the image model)
- mosaic.call now return the catalogs and segmentation map model (in addition to the mosaic model)

The modifications above allowed removal of:
- SourceCatalogStep.return_updated_model (and removal of the tests the exercised this)
- source_catalog._save_utils

Regression tests show 1 difference:
https://github.com/spacetelescope/RegressionTests/actions/runs/27376974541
where the output file for `test_pipeline_suffix` which [skips twearkeg](https://github.com/braingram/romancal/blob/26409040b8e777a633ae345281509397bdf22184/romancal/regtest/test_wfi_image_pipeline.py#L341) no longer contains a meta.source_catalog.tweakreg_catalog entry.

## Tasks


romancal/multiband_catalog/multiband_catalog_step.py
```diff
@@ -99,7 +114,27 @@ def process(self, dataset):
         if msg is None:
             segment_img, cat_model = results
         else:
-            return save_empty_results(self, *results, msg=msg)
```
schlafly, line 102: Do you think it makes sense to do something more like return create_empty_results(...), just to try to isolate this logic from the main path?

romancal/multiband_catalog/multiband_catalog_step.py
```diff
@@ -130,6 +165,40 @@ def process(self, dataset):
...
+            if hasattr(segment_img, "si_segment_img"):
+                segmentation_model["si_data"] = segment_img.si_segment_img.data.astype(
+                    np.uint32
+                )
+            if hasattr(segment_img, "si_detection_image"):
+                segmentation_model["si_detection_image"] = (
+                    segment_img.si_detection_image
+                )
+
+        return cat_model, segmentation_model
```
schlafly, line 172: It looks like you wrote this in a pretty defensive way such that maybe it could share code with the create_empty_results path?

romancal/pipeline/exposure_pipeline.py
```diff
@@ -109,28 +106,70 @@ def process(self, dataset):
...
+
+                if not self.tweakreg.skip and len(catalog.source_catalog):
+                    # attach the catalog to the model so tweakreg can see it
+                    if "source_catalog" not in model.meta:
+                        result.meta["source_catalog"] = {}
+                    result.meta.source_catalog.tweakreg_catalog = catalog.source_catalog
+
                 lib.shelve(result, model_index)
+                catalogs.append(catalog)
+                segmentations.append(segmentation)
```
schlafly, line 129: Noting: now the ELP always returns the group of three libraries: images, catalogs, segmentation maps.  When the latter two don't make sense they're a bunch of empty catalogs and empty segmentation maps.

romancal/tweakreg/tweakreg_step.py
```diff
@@ -273,6 +273,28 @@ def save_model(self, result, *args, **kwargs):
...
+
+        for x_col, y_col, ra_col, dec_col in updates:
+            if any(c not in catalog.colnames for c in (x_col, y_col, ra_col, dec_col)):
+                # Only update existing columns to preserve the file schema.
+                continue
+            catalog[ra_col], catalog[dec_col] = tweaked_wcs.pixel_to_world_values(
+                catalog[x_col], catalog[y_col]
+            )
+
     def update_catalog_coordinates(self, tweakreg_catalog_name, tweaked_wcs):
```
schlafly, line 297: We have update_catalog_coordinatse and _update_catalog_coordinates.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -109,28 +106,70 @@ def process(self, dataset):
...
 
+        # tweakreg was run, update catalog positions
+        with lib:
+            for model_index, model in enumerate(lib):
+                if model.meta.cal_step.tweakreg == "COMPLETE":
+                    catalog = catalogs[model_index]
+                    self.tweakreg._update_catalog_coordinates(
+                        catalog.source_catalog, model.meta.wcs
+                    )
+                lib.shelve(model)
```
schlafly, line 151: It's now the ELP's job to do the coordinate updating.

schlafly (commented):
This looks like a good approach, thanks Brett!

I think providing a tweakreg catalog name when possible is helpful.  The role here is to say what was used to do the registration, assuming that most folks will be sending a catalog from disk or making one (as we do in the ELP).  I'd rather not drop that.   Guessing the future name in the ELP doesn't sound crazy to me if this is straightforward.  For the general case of 'I was passed a catalog in memory' I don't think we need to handle this.

I also think that most folks will want to update ra / dec following tweakreg.  Conceptually, the source catalog step measures x & y in an image and ra / dec are only really known after tweakreg (though they have an initial guess value from the initial WCS).  So in my head the default is to update.  But I'm happy doing this only when the ELP runs (as you have implemented) and providing some kind of update routine as you do here.

For the case of images that don't get catalogs---are we now writing empty catalogs and segmentation files, etc. for them?  I think I see that with this PR we are now always returning all three.  We definitely don't want ops to see new files for a spectroscopic / lolo run.  I think we need to make sure that the files do not get produced at all for those cases, distinguishing between 'the catalog step was skipped' and 'we ran the catalog but something about the image was so bad that we got an empty catalog out.'

braingram:
> This looks like a good approach, thanks Brett!

Thanks for taking a look at this approach.

> I think providing a tweakreg catalog name when possible is helpful. The role here is to say what was used to do the registration, assuming that most folks will be sending a catalog from disk or making one (as we do in the ELP). I'd rather not drop that. Guessing the future name in the ELP doesn't sound crazy to me if this is straightforward. For the general case of 'I was passed a catalog in memory' I don't think we need to handle this.
> 
> I also think that most folks will want to update ra / dec following tweakreg. Conceptually, the source catalog step measures x & y in an image and ra / dec are only really known after tweakreg (though they have an initial guess value from the initial WCS). So in my head the default is to update. But I'm happy doing this only when the ELP runs (as you have implemented) and providing some kind of update routine as you do here.

On main the coordinates are only updated for elp runs since the default for `TweakRegStep.update_source_catalog_coordinates` is `False`:
https://github.com/spacetelescope/romancal/blob/67d9afce95fac9c5448d34cf106877065eaa9200/romancal/tweakreg/tweakreg_step.py#L74

I aimed to keep that default unchanged in this PR but still keep it as an option. We could change the default and document that this only works for:
- catalogs provided via `tweakreg_catalog_name`
- catalogs that are in parquet format (and possibly only those produced by romancal)

> For the case of images that don't get catalogs---are we now writing empty catalogs and segmentation files, etc. for them? I think I see that with this PR we are now always returning all three. We definitely don't want ops to see new files for a spectroscopic / lolo run. I think we need to make sure that the files do not get produced at all for those cases, distinguishing between 'the catalog step was skipped' and 'we ran the catalog but something about the image was so bad that we got an empty catalog out.'

I'll take a look. Is the goal here to have the elp return only a model if run with spectroscopic/lolo data (compared to the (model, catalog, segmentation) output for imaging data)? What should happen if a user skips tweakreg, should it also return just the model?

romancal/tweakreg/tweakreg_step.py
braingram: Yeah, the current `update_catalog_coordinates` function only supports updating a parquet file from disk. I can update it to use `_update_catalog_coordinates` once we've defined the goals of what needs to update catalogs and when.

romancal/multiband_catalog/multiband_catalog_step.py
braingram: I copied this from `_save_utils.py`:
https://github.com/spacetelescope/romancal/blob/67d9afce95fac9c5448d34cf106877065eaa9200/romancal/source_catalog/_save_utils.py#L82-L87
I suspect it was written this way due to the complicated way that these modules were interacting. This step could likely be simplified by just creating the segmentation_model earlier and adding to it directly rather than attaching stuff to the segment_img which I don't think photutils needs.

romancal/multiband_catalog/multiband_catalog_step.py
braingram: I'll take a look. Given https://github.com/spacetelescope/romancal/pull/2352#discussion_r3375659975 I think it's worth doing some more refactoring in this step.

braingram:
Looking at main, LOLO has the source catalog step run:
https://github.com/spacetelescope/romancal/blob/67d9afce95fac9c5448d34cf106877065eaa9200/romancal/pipeline/exposure_pipeline.py#L170-L172
so I would expect that `cat.parquet` and `segm.asdf` files are produced for that exptype. Is that a bug on main?

schlafly:
> Looking at main, LOLO has the source catalog step run:

Thanks.  I should have said no catalogs for dark / flat / spectroscopic / wfsc.  Indeed, for LOLO we want the catalogs.  In the context of this PR, I want to continue to not produce (empty) catalogs for data products where we've been asked not to produce catalogs.

schlafly:
> On main the coordinates are only updated for elp runs since the default for `TweakRegStep.update_source_catalog_coordinates` is `False`:
> 
> https://github.com/spacetelescope/romancal/blob/67d9afce95fac9c5448d34cf106877065eaa9200/romancal/tweakreg/tweakreg_step.py#L74
> 
> I aimed to keep that default unchanged in this PR but still keep it as an option. We could change the default and document that this only works for:
> 
> * catalogs provided via `tweakreg_catalog_name`
> * catalogs that are in parquet format (and possibly only those produced by romancal)

That's perfect.

> > For the case of images that don't get catalogs---are we now writing empty catalogs and segmentation files, etc. for them? I think I see that with this PR we are now always returning all three. We definitely don't want ops to see new files for a spectroscopic / lolo run. I think we need to make sure that the files do not get produced at all for those cases, distinguishing between 'the catalog step was skipped' and 'we ran the catalog but something about the image was so bad that we got an empty catalog out.'
> 
> I'll take a look. Is the goal here to have the elp return only a model if run with spectroscopic/lolo data (compared to the (model, catalog, segmentation) output for imaging data)? What should happen if a user skips tweakreg, should it also return just the model?

Broadly, the goal is not to produce an empty file when we don't want to make a catalog in the first place.  So yes, I would say if the user skips the source catalog step we don't want catalog / segmentation files to be created.  I don't think tweakreg should control this, but skipping source cataloging should.  And thanks, I was confused again about LOLO; LOLO and imaging get source catalogs and segmentation maps in addition to images, but everything else (dark / spectral / wfsc) do not.

I don't know what mechanisms are available to achieve this, but I'd be happy, e.g., with the following type of approach:
- right now you add an empty segmentation map / catalog to the library if the step skips
- instead your could imagine adding None
- when writing out a library with None, don't write out

I expect that doesn't actually work (None isn't a valid enough entry in a library?), but that feels like it could work with your current approach.

romancal/multiband_catalog/multiband_catalog_step.py
```diff
@@ -58,8 +60,21 @@ class MultibandCatalogStep(RomanStep):
         inject_seed = integer(default=None)   # RNG seed for injected sources
         save_debug_info = boolean(default=False)
                                    # Include image data and other data for testing
+        output_use_model = boolean(default=True)  # When saving use `DataModel.meta.filename`
```
braingram, line 63: This is overriding the default `output_use_model=False` to match the current behavior in `_save_utils`.

romancal/multiband_catalog/multiband_catalog_step.py
braingram: This step has seen some additional simplification. Would you give the current state a look and see if this still applies?

romancal/multiband_catalog/multiband_catalog_step.py
braingram: This was simplified in the latest code. I'll resolve this comment but please unresolve this if the latest code doesn't address the issue.

romancal/pipeline/exposure_pipeline.py
braingram: This is slightly different in the current code. Now if the source_catalog step is not run, no catalog or segmentation map will be produced and the pipeline will return `None` values for those items.

romancal/pipeline/exposure_pipeline.py
braingram: For the current code it's still possible to update the catalog coordinates with tweakreg (as long as the catalog is saved to a parquet file). However the elp does not exercise this option (since there is no parquet file at the time of running tweakreg) and instead corrects the catalog in-memory.

romancal/pipeline/exposure_pipeline.py
```diff
@@ -109,28 +102,78 @@ def process(self, dataset):
...
+
         log.info("Roman exposure calibration pipeline ending...")
 
         # return a ModelLibrary
         if return_lib:
-            return lib
+            return (
+                lib,
+                ModelLibrary([c for c in catalogs if c is not None]),
```
braingram, line 159: For an elp run using an association the pipeline returns 3 libraries: models, catalogs, segmentation maps.

There is one nuance where if the association contains mixed exptypes that result in some members skipping source_catalog the libraries will have unequal numbers of members (since ModelLibrary can't contain `None` members). We could convert these `None`s into empty catalogs/segmentation maps or perhaps disallow associations with mixed exptypes? Perhaps there is a better option for this PR but I'm ok with the current behavior given the unique conditions required for this corner case.

romancal/pipeline/mosaic_pipeline.py
```diff
@@ -66,16 +89,23 @@ def process(self, dataset):
         self.outlier_detection.resample_on_skycell = self.resample_on_skycell
         self.resample.resample_on_skycell = self.resample_on_skycell
 
+        if self.save_results:
+            # only set if save_results is True since setting
+            # output_file will trigger results to be saved
+            self.output_file = library.asn["products"][0]["name"]
```
braingram, line 95: This tripped me up so I'm calling it out. Setting `output_file` overrides `save_results` so we need to gate the assignment of `output_file` in this pipeline.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -125,8 +123,7 @@ def image_model():
 
 def test_forced_catalog(image_model, function_jail, ignore_parquet_metadata_paths):
     output_filename = "force_cat.parquet"
-    step = SourceCatalogStep()
-    _ = step.call(
+    _ = SourceCatalogStep.call(
```
braingram, line 126: I updated these as I saw them. `call` is a [class method](https://github.com/spacetelescope/stpipe/blob/416bcf9492c4fbe452a127767a9c908074b6d587/src/stpipe/step.py#L698) which makes a new step instance. For this example creating `step` is unnecessary and misleading (usage like this in jwst led to confusion about how `call` works).

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -480,117 +481,6 @@ def test_do_psf_photometry_column_names(function_jail, image_model, fit_psf):
         assert len(psf_colnames) == 0
 
 
-@pytest.mark.parametrize(
-    "snr_threshold, npixels, nsources, save_results, return_updated_model, expected_result, expected_outputs",
```
braingram, line 484: `return_updated_model` is removed in this PR so this test is no longer needed.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -480,117 +481,6 @@ def test_do_psf_photometry_column_names(function_jail, image_model, fit_psf):
...
-        # none_cat.parquet), this test needs to know how to translate that back
-        # to the equivalent segmentation file.
-        basefilename = result.meta.filename.split("_")[0]
-        filepath = Path(function_jail / f"{basefilename}_{suffix}.{ext}")
-        assert filepath.exists()
-
-        if suffix == "cat":
-            # the catalog is saved as a parquet file
-            tbl = pyarrow.parquet.read_table(filepath)
-            assert isinstance(tbl, pyarrow.Table)
```
braingram, line 589: This test duplicates checks in `test_l2_source_catalog` so this PR removes this redundant test.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -628,150 +518,3 @@ def fail_getreferences(*args, **kwargs):
...
-            False,
-            MosaicSourceCatalogModel,
-            {
-                "cat": MosaicSourceCatalogModel,
-                "segm": MosaicSegmentationMapModel,
-            },
-        ),
-    ),
-)
-def test_l3_source_catalog_keywords(
```
braingram, line 687: This test is redundant to test_l3_source_catalog.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -628,150 +518,3 @@ def fail_getreferences(*args, **kwargs):
...
-            True,
-            ImageModel,
-        ),
-        (
-            False,
-            ImageSourceCatalogModel,
-        ),
-    ),
-)
-def test_l2_source_catalog_return_updated_model_attribute(
```
braingram, line 753: This PR removes `return_updated_model` so this test was removed.

romancal/stpipe/core.py
```diff
@@ -98,6 +99,19 @@ def finalize_result(self, model, reference_files_used):
...
             is the reftype code, the second element is the filename.
         """
+        if model is None:
+            # None can be passed to this function if a step or pipeline returns
+            # None inside a tuple of results (for example SourceCatalogStep).
+            return
+
+        if isinstance(model, ModelLibrary):
+            # This works around what is likely an issue in how stpipe handles
+            # tuples that contain ModelLibrary instances: https://github.com/spacetelescope/stpipe/issues/322
```
braingram, line 109: As noted in the comment there is an upstream issue for this https://github.com/spacetelescope/stpipe/issues/322

romancal/multiband_catalog/multiband_catalog_step.py
schlafly: This looks great.

romancal/multiband_catalog/multiband_catalog_step.py
schlafly: This duplicates the stpipe docstring, but I don't understand what this option does.

romancal/pipeline/exposure_pipeline.py
schlafly: I agree that this is a weird corner case and I don't feel strongly about what the pipeline does it; current implementation is good, thanks!

schlafly (approved):
This looks good to me, thank Brett.  Additionally the regtest difference makes sense.  Any reasons we shouldn't merge this ~now?

I left one comment inline where I didn't understand what the new output_use_model = boolean(True) option to MultibandCatalogStep was doing.

romancal/multiband_catalog/multiband_catalog_step.py
braingram: output_use_model generates the output filename based on the meta.filename of the output model. It's a pretty confusing collection of code that generates the output filename but it involves the following.

When a step completes and returns a result, if save_results is True, stpipe will generate a filename for the result. To start at the call to save_model:
https://github.com/spacetelescope/stpipe/blob/34bc943c52398ea1f54b5688cf7b8fc0ef298da9/src/stpipe/step.py#L622
this is called without providing a filename (only the model is provided). stpipe then starts to try and figure out the filename by:
- [checking Step.output_file](https://github.com/spacetelescope/stpipe/blob/34bc943c52398ea1f54b5688cf7b8fc0ef298da9/src/stpipe/step.py#L1098) which might be set by the user
- [checking meta.filename if output_use_model is enabled](https://github.com/spacetelescope/stpipe/blob/34bc943c52398ea1f54b5688cf7b8fc0ef298da9/src/stpipe/step.py#L1136)
- [finally the filename creation will fallback to parent output_file or a default filename](https://github.com/spacetelescope/stpipe/blob/34bc943c52398ea1f54b5688cf7b8fc0ef298da9/src/stpipe/step.py#L1200)

On romancal main in _save_utils much of this logic is bypassed since an output_file is provided to save_model:
https://github.com/spacetelescope/romancal/blob/41dd6a9fc08bfd5b4a31d81c3fad0799b492cc2e/romancal/source_catalog/_save_utils.py#L93
The output_file is determined from:
https://github.com/spacetelescope/romancal/blob/41dd6a9fc08bfd5b4a31d81c3fad0799b492cc2e/romancal/source_catalog/_save_utils.py#L143
- Step.output_file or
- model.meta.filename

which most closely matches the stpipe behavior when output_use_model is True.

PR #2355, opened 2026-06-10
Rcal-1395 Make association public/private and remove unused modules

### LINKED ISSUE #2326: Make association public/private and remove unused modules
_Issue [RCAL-1395](https://jira.stsci.edu/browse/RCAL-1395) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

Keep asn_from_list.py, multiband_asn.py, and skycell_asn.py as public methods and moving the other to private status. Also remove any unused modules.

ddavis-stsci:
Resolves [RCAL-1395](https://jira.stsci.edu/browse/RCAL-1395)

Closes #2326

This PR addresses keeps asn_from_list.py, multiband_asn.py, and skycell_asn.py as public methods and moves the remainder to private. It also removes skycell_asn.py and skycell_asn.py modules and updates the documentation.
reopening this as a new PR.

Tasks

  

## Tasks
      - ]x] **after the reviewer has approved these changes**, run `okify_regtests` to update the truth files

the regression tests are at https://github.com/spacetelescope/RegressionTests/actions/runs/27294549405


romancal/associations/__init__.py
```diff
@@ -22,12 +22,9 @@ def libpath(filepath):
...
-from .load_asn import load_asn
-from .main import *
-from .pool import *
-from .registry import *
+from ._association import _Association
+from ._association_io import AssociationNotValidError, json
+from ._generate import _generate
+from ._load_asn import load_asn
+from ._pool import _AssociationPool
+from ._registry import _AssociationRegistry
```
mairanteodoro, line 30: This looks like an API-breaking change for downstream and users of `romancal.associations`. Should we add compatibility aliases (e.g. `from ._association import _Association as Association`) and then deprecate them gradually?

romancal/associations/mk_skycell_asn_from_skycell_list.py
mairanteodoro, line 1: `pyproject.toml` needs to be updated to reflect the removal of this module. Currently, it contains:
```python
[project.scripts]
asn_from_list = "romancal.associations.asn_from_list:_cli"
mk_skycell_asn_from_skycell_list= "romancal.associations.mk_skycell_asn_from_skycell_list:_cli"
mk_skycell_list = "romancal.associations.mk_skycell_list:_cli"
```

romancal/associations/mk_skycell_list.py
mairanteodoro, line 1: `pyproject.toml` needs to be updated to reflect the removal of this module. Currently, it contains:
```python
[project.scripts]
asn_from_list = "romancal.associations.asn_from_list:_cli"
mk_skycell_asn_from_skycell_list= "romancal.associations.mk_skycell_asn_from_skycell_list:_cli"
mk_skycell_list = "romancal.associations.mk_skycell_list:_cli"
```

mairanteodoro (changes requested):
Thanks for the refactoring, Dave!

Overall, the changes are fine, but I worry that it might be an API-breaking change for downstream users importing from `romancal.associations`. If that's indeed the intention here, then I suggest that we document it somewhere (in the change log entry, maybe?).

Other than that, we need to update `[project.scripts]` in `pyproject.toml` to reflect the removal of `mk_skycell_asn_from_skycell_list.py` and `mk_skycell_list.py`.

romancal/associations/mk_skycell_list.py
ddavis-stsci: Thanks, I forgot about these in the pyproject file.

romancal/associations/__init__.py
ddavis-stsci: Do we know of anyone using these so that it would be an issue? Most of the use for these are for generating associations from the pool files, which we are not using. 

If someone is using this I'm happy to have the aliases but I think this might be a good time to make these private before launch.

romancal/associations/mk_skycell_asn_from_skycell_list.py
ddavis-stsci: Thanks, I forgot about these in the pyproject file.

romancal/associations/__init__.py
mairanteodoro: Thanks, Dave!
I'm fine with that too. Just wanted to make sure it's not being used anywhere else. I know [`roman_photoz` uses `load_asn`](https://github.com/spacetelescope/roman_simulate_dr/blob/a304b37ac3d6824f756d1d9f46ee49ebc31ca990/roman_simulate_dr/scripts/visualization_utils/visualize_generic_coadd.py#L31), but that's such a small change that it shouldn't be a problem updating that repo.

mairanteodoro (approved):
Looks good to me! Thanks, Dave!

schlafly (commented):
Thanks Dave.  I like the removal of the skycell list stuff and the deletion of the unused decorators.

Is there a rule you followed for deciding which portions of lib/ to make private?  e.g., acid / association_rules / constraint / decorators / diff / dms_base / ... remain public following this change, but counter / ioregistry / keyvalue_registry / process_list are made private?

I did briefly try to see how I would have done this refactor (with Claude).  There I sent association.py to _association.py but did not send Assocation to _Association. I guess that's mostly just an arbitrary choice but it feels like that injects more churn.  Likewise I sent callback registry -> _callback_registry but did not do CallbackRegistry to _CallbackRegistry.  I took something more like the approach you took in counter, renaming counter -> _counter but not Counter to _Counter.  That branch is here
https://github.com/schlafly/romancal/tree/make-associations-private
and includes making the rest of lib/ private, which I hadn't previously done.

re load_asn, it looks like now you re-export that so that e.g. roman_simulate_dr will need to see no changes.  That seems like a good choice.  @mairanteodoro , is that how you see this too?

ddavis-stsci:
> Thanks Dave. I like the removal of the skycell list stuff and the deletion of the unused decorators.
> 
> Is there a rule you followed for deciding which portions of lib/ to make private? e.g., acid / association_rules / constraint / decorators / diff / dms_base / ... remain public following this change, but counter / ioregistry / keyvalue_registry / process_list are made private?
> 
> I did briefly try to see how I would have done this refactor (with Claude). There I sent association.py to _association.py but did not send Assocation to _Association. I guess that's mostly just an arbitrary choice but it feels like that injects more churn. Likewise I sent callback registry -> _callback_registry but did not do CallbackRegistry to _CallbackRegistry. I took something more like the approach you took in counter, renaming counter -> _counter but not Counter to _Counter. That branch is here https://github.com/schlafly/romancal/tree/make-associations-private and includes making the rest of lib/ private, which I hadn't previously done.
> 
> re load_asn, it looks like now you re-export that so that e.g. roman_simulate_dr will need to see no changes. That seems like a good choice. @mairanteodoro , is that how you see this too?

I left acid / association_rules etc public so if anyone wants to add new rules for an association they have access to these modules. 

The Association and CallbackRegistry should only be used internally thus the moving them to private. 

We can make everything in the lib directory private but it will make it less transparent for folks making any custom associations.

schlafly:
> I left acid / association_rules etc public so if anyone wants to add new rules for an association they have access to these modules.
> 
> The Association and CallbackRegistry should only be used internally thus the moving them to private.
> 
> We can make everything in the lib directory private but it will make it less transparent for folks making any custom associations.

Doesn't use of these custom rules require things like an association pool file, etc., that we don't really make or support right now?  If I try to imagine what a custom association rule might look like, it would be something like a GBTDS parallax-sensitive coadd that includes every other GBTDS season in a stack.  In practice I would imagine now telling folks to write whatever code they want to generate the list of every-other-season and then use asn_from_list to bundle that as an association, and I would not refer them to acid / association_rules / ....  Do you think that something in those modules is a better starting point for making a custom association?

ddavis-stsci:
> > I left acid / association_rules etc public so if anyone wants to add new rules for an association they have access to these modules.
> > The Association and CallbackRegistry should only be used internally thus the moving them to private.
> > We can make everything in the lib directory private but it will make it less transparent for folks making any custom associations.
> 
> Doesn't use of these custom rules require things like an association pool file, etc., that we don't really make or support right now? If I try to imagine what a custom association rule might look like, it would be something like a GBTDS parallax-sensitive coadd that includes every other GBTDS season in a stack. In practice I would imagine now telling folks to write whatever code they want to generate the list of every-other-season and then use asn_from_list to bundle that as an association, and I would not refer them to acid / association_rules / .... Do you think that something in those modules is a better starting point for making a custom association?

We don't use pool files for the Roman associations but they can still be used. 
If you want all the criteria used for associations to be external from the association code that is fine. 
So all lib modules should move to be private?

schlafly:
> We don't use pool files for the Roman associations but they can still be used. If you want all the criteria used for associations to be external from the association code that is fine. So all lib modules should move to be private?

I don't currently foresee us using these and so don't think we should have it as part of the public API.  Yes, let's make them private.

ddavis-stsci:
The associations/lib modules are now private and the unit tests and regression tests are passing.
https://github.com/spacetelescope/RegressionTests/actions/runs/28873319244

schlafly (approved):
Thanks, this looks good to me.  I'm happy to merge; are you ready?

PR #2360, opened 2026-06-12
Fix read noise sqrt(2) factor in likely ramp fitting; flat field dumo

### LINKED ISSUE #2354: Flat field step needs to flat field dumo as well
We added the 'dumo' field to ramp fitting, which is the difference between the uniform and optimal ramp fits.  These values need to be adjusted when we do the flat fielding, since both the uniform & optimal images need to be flattened.

We don't have separate uncertainties on the dumo quantities so this is only a single line parallel to this one.
https://github.com/spacetelescope/romancal/blob/main/romancal/flatfield/flat_field.py#L84

dumo is itself rather small and the flat field isn't far from one, but there's no reason not to do this.

@t-brandt , do you see this differently from me?

### LINKED ISSUE #2358: stcal assumes read noise is CDS read noise
romancal assumes that read noise reflects how far an individual read is from truth, while stcal's ramp fitting assumes that read noise is the typical difference between two reads of the same quantity (CDS read noise).  These two quantities are conceptually different by the square root of two; stcal converts the CDS read noise to the ordinary read noise here:
https://github.com/spacetelescope/stcal/blob/main/src/stcal/ramp_fitting/likely_fit.py#L74
romancal currently sends the ordinary read noise instead of the CDS read noise, meaning that this conversion leads us to use too little read noise.

We should fix this by sending stcal the CDS read noise rather than the ordinary read noise.

schlafly:
Closes #2358
Closes #2354

This PR fixes two issues: first, the 'dumo' field needs to be corrected with the flat field, and second, romancal and stcal were using different conventions for the readnoise, leading to confusion.  romancal now multiplies its read noise reference file by sqrt(2), converting from per-read uncertainties to CDS uncertainties.

The biggest updates by lines are to the tests which hard-code a particular set of return values to ramp fitting; those values get updated to the currently returned values.  I added a new test that aims to validate correctness, demanding that when ramps are sent in with known amounts of noise that the scatter in the slopes match the estimated uncertainties, that the difference between the measured slopes and the true rate is a small fraction of the uncertainty on average, and that (measured - true) / uncertainty has a standard deviation of 1.

We intend this PR to be in a patch for B22.

This is a bugfix and only requires updating the 'patch' version.

There will need to be a fair amount of okifying, since this will change data, masking, and uncertainties following ramp fitting.  Regression test results here:
https://github.com/spacetelescope/RegressionTests/actions/runs/27550058944/job/81433653869#step:40:3489
Note the mild improvement (1%) in image RMS, likely due to moving from optimal-with-the-wrong-read-noise to optimal-with-the-right-read-noise weights.

## Tasks
- [X] if a JIRA ticket exists, [make sure it is resolved properly](https://github.com/spacetelescope/romancal/wiki/How-to-resolve-JIRA-issues)


schlafly:
@t-brandt , this PR addresses the two issues we talked about yesterday.  Mostly just to alert you; there's not a lot here to look at.

stscieisenhamer (approved):

ddavis-stsci (approved):

PR #2368, opened 2026-06-27
RCAL-1342: Source Injection Nan Pixel Support

### LINKED ISSUE #2022: Revise Source Injection NaN Handling
We currently use NaN to avoid populating injected sources in regions where there is no imaging, but we probably do want to populate sources where only a small region is NaN (e.g., due to a cosmic ray).

PaulHuwe:
Resolves [RCAL-1342](https://jira.stsci.edu/browse/RCAL-1342)

Closes #2022

This PR adds support for injecting sources onto NaN pixels. Previously, if an object was to be injected at the location of a NaN pixel, it would be skipped. This PR will inject the source at a location with a single NaN at the grid point, using an interpolated pixel value for injection. If a source is to be injected in a NaN region, it is skipped. 

NOTES:
1. The injection is coded in such a way that it can be readily expanded to a region of NaN pixels that are some fraction of the source size, if desired. 
2. This PR adds some code duplication in the test file, which was needed to avoid changing other tests. I want to fully clean up this test file in a separate PR.

## Tasks


romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -315,6 +344,57 @@ def test_multiband_catalog_some_invalid_inputs(
     assert np.all(np.isnan(cat["segment_f184_flux_err"]))
 
 
+def make_si_nan_test_image():
```
schlafly: Can the new test use the old test image or the old test use the new test image?  It just isn't obvious to me why two different sets of test patterns are needed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -315,6 +344,57 @@ def test_multiband_catalog_some_invalid_inputs(
...
+    rng = np.random.default_rng(seed=RNG_SEED)
+    noise_scale = 0.01 * MEANFLUX
+    noise = rng.normal(loc=MEANFLUX, scale=noise_scale, size=data.shape)
+    data += noise
+    err = noise_scale * np.ones_like(data)
+
+    return data, err
+
+
+def mosaic_si_nan_model(shape=(500, 500)):
```
schlafly, line 375: Could we reuse the code from mosaic_model or mosaic_si_model here?  I don't immediately spot the differences.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +543,147 @@ def test_multiband_source_injection_catalog(
...
+    # Filter by a maximum match radius (one pixel)
+    match_mask = d2d < 0.1 * u.arcsec
+
+    matched_si_g = results["NoNan"].source_injection_catalog[match_mask]
+    matched_grid = results["Grid"].source_injection_catalog[idx[match_mask]]
+
+    # Ensure all objects matched
+    assert len(matched_grid) == len(results["NoNan"].source_injection_catalog)
+
+    # Test that all matched objects have similar magnitudes
```
schlafly: Does this test depend on the fact that the NaNs are far from the sources?  Broadly I think it's fine if the magnitudes are wrong if you NaN out the center of a source; I don't think we want this test to depend on how the fake sources are located relative to the NaNs.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +543,147 @@ def test_multiband_source_injection_catalog(
...
+    coords_block = SkyCoord(
+        results["Block"].source_injection_catalog["ra"],
+        results["Block"].source_injection_catalog["dec"],
+    )
+    coords_si_bl = SkyCoord(
+        results["NoNan"].source_injection_catalog["ra"],
+        results["NoNan"].source_injection_catalog["dec"],
+    )
+
+    # Find indices in GridSI that best match OrigSI
```
schlafly:
```suggestion
    # Find indices in BlockSI that best match OrigSI
```
GridSI should be BlockSI, if I've followed correctly?

romancal/multiband_catalog/_multiband_catalog.py
```diff
@@ -580,13 +581,82 @@ def make_source_injected_library(library, seed=None):
                 si_cat["y_pos"] = si_y_pos
                 si_cat["label"] = np.arange(len(si_x_pos))
 
+            # Handle NaN regions
```
schlafly, line 583: I'm missing the point of this block.

If you do not include the new lines 584 through 648, what happens?  It's possible injection fails because it doesn't know how to handle adding to NaN, so you're working around it by first "un-NaN-ing" the image.  If that's it, the logic could be:
data[nanmask] = 0
inject
data[nanmask] = nan
?

romancal/multiband_catalog/_multiband_catalog.py
```diff
@@ -580,13 +581,82 @@ def make_source_injected_library(library, seed=None):
...
+
+                    # Set temporary interpolation value
+                    si_model.data[
+                        min(y_stamp) : max(y_stamp) + 1, min(x_stamp) : max(x_stamp) + 1
+                    ][mask_stamp] = interpolated_matrix[mask_stamp]
+
+                # Remove grid points in extended NaN regions
+                if idx_bad:
+                    nanmask[idx_bad] = False
+                    si_nan_cat.remove_rows(nanmask)
```
schlafly: Good, I think the logic here is 'don't bother injecting sources in fully masked regions.'  I think the added code complexity is high enough though that we might as well just inject the sources anyway and not expect to find them, if that enables a simpler code flow.

schlafly (commented):
Thanks.  I identified some of the same code duplication in the testing that you did.  I agree that it makes sense to clean this up; I don't think we want this PR to introduce more duplication in preparation for a later PR that removes it?

I left a few other in line comments about the tests.

More importantly, I'm not sure I really get what the purpose of the new block in _multiband_catalog is.  If the PR were only the deletion of 302-306 in _injection, what would go wrong?

romancal/multiband_catalog/_multiband_catalog.py
PaulHuwe: Do we anticipate a use case where someone would override the defaults to add a large number of sources to inject? That would be my one concern with removing this.

schlafly: I think anyone injecting large number of sources is prepared to handle the volume of the large number of sources they're injecting.

My comment is predicated on my sense that most of the new code in this PR is trying to avoid injecting in completely empty regions and thinking that we can avoid that complexity.  If that's not right and its trivial to avoid injecting in fully masked regions then I think there's some argument for avoiding it.

If we don't care about this does the code reduce to just lines 302-306?  In which case, yes, let's simplify the code and let folks generating large numbers of simulated sources get large numbers of simulated results back.

romancal/multiband_catalog/_multiband_catalog.py
PaulHuwe: Removed as discussed offline.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Refactored to use the old image.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Refactored to use mosaic_si_model.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: The NaNs are only at the center of the injected sources.

romancal/multiband_catalog/_multiband_catalog.py
PaulHuwe: Removed.

PaulHuwe:
Regression tests pass: https://github.com/spacetelescope/RegressionTests/actions/runs/29050165165

romancal/multiband_catalog/tests/test_multiband_catalog.py
schlafly: Did I misunderstand this?  This comment still looks wrong.

romancal/multiband_catalog/_multiband_catalog.py
```diff
@@ -580,13 +580,28 @@ def make_source_injected_library(library, seed=None):
...
+            # Handle NaN regions
+            y_pos_idx, x_pos_idx = (
+                np.round(si_y_pos).astype(int),
+                np.round(si_x_pos).astype(int),
+            )
+            nanmask = np.isnan(si_model.data[y_pos_idx, x_pos_idx])
+
+            # Temporarily set NaNs to zero for injections
+            si_model.data[y_pos_idx[nanmask], x_pos_idx[nanmask]] = 0
+            si_model.var_poisson[y_pos_idx[nanmask], x_pos_idx[nanmask]] = 0
```
schlafly: Per our discussion Wednesday, the ratio here is used to determine the exposure time, so we should fill with nanmedian or something rather than just zeros (I don't know what exposure time 0/0 would correspond to).

romancal/multiband_catalog/tests/test_multiband_catalog.py
schlafly: Thanks.  What is the typical dispersion here?  If I understand right, this is matching the recovered magnitude of the source on a partially NaNed image to the recovered magnitude of the same source, on an image without NaNs.  Conceptually I expect the recovered magnitude to depend on whether the the image has NaNs.  For stars, I expect it to depend a lot because the stars are small and NaNing their centers makes a big impact on the flux estimate you would get.  For galaxies, it depends on how big the galaxy is and things are less clear to me.  2 mmag is a very tight tolerance and I worry that this is a sign that something is not working right.  What am I missing?

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +487,142 @@ def test_multiband_source_injection_catalog(
...
+    matched_si_bl = res_cat["NoNan"].source_injection_catalog[match_mask]
+    matched_block = res_cat["Block"].source_injection_catalog[idx[match_mask]]
+
+    # Test that all matched objects have similar magnitudes
+    assert np.allclose(
+        matched_block["kron_f158_abmag"], matched_si_bl["kron_f158_abmag"], rtol=3e-2
+    )
+    assert np.allclose(
+        matched_block["kron_f184_abmag"], matched_si_bl["kron_f184_abmag"], rtol=3e-2
+    )
```
schlafly, line 725: Similarly to the comment above... this is for the case where large chunks of the image are NaN.  So we expect some sources to not be recovered at all.  I guess in this section there's no assertion that the lengths are identical, reflecting that.  But the requirement that the kron mags match at 0.03 mag means that no sources are allowed to be injected near the edge?

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: I applied the change, but it got lost in a git merge. Trying again.

romancal/multiband_catalog/_multiband_catalog.py
PaulHuwe: When data and var_poisson is zero, exposure time is 0/1 - the method changes any 0 var_poisson to 1. I was going for the least bespoke change, but I am happy to switch to nanmedian.
https://github.com/spacetelescope/romanisim/blob/main/romanisim/l3.py#L172C1-L177C27

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Source are injected near the edge, but are not detected. 

Injected sources for NaN Block:
image

Zoom of the center NaN Block corner:
image

PaulHuwe: Your suspicions are well founded - during PR revisions, the NANPOINTS drifted. I am rewriting that section (and tests) to be less brittle.

romancal/multiband_catalog/_multiband_catalog.py
schlafly: Let's fill with the nanmedians of the whole image so that these sources have sane exposure times.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: I believe this is addressed with the new changes.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -415,22 +485,7 @@ def test_multiband_source_injection_catalog(
...
-        (65, 18),
-        (41, 43),
-        (17, 53),
-        (65, 71),
-        (20, 80),
-        (85, 88),
-    ]
-    for modx, mody in cat[["x_mod", "y_mod"]]:
-        assert (modx, mody) in gocj_locs
+    assert len(cat) == 112
```
schlafly, line 488: My sense is that the old version of this had 175 = 7 * 25, for your seven sources tiled 5x5, and so it made sense to assert this number (we detected all of the sources).  The new number is opaque to me.  Then the test went on to check that the sources were also found at the correct locations.

I think I understand that you added additional complexity to the test image (though I'm not immediately sure what the motivation was), so the location test became complicated, and somewhat surprisingly (?) that also led to the number of sources detected dropping dramatically.

I'd propose not including that in this PR in order to simplify things.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+        results["NoNan"].si_detection_image[~block_nan_stamp_mask],
+        results["Block"].si_detection_image[~block_nan_stamp_mask],
+        atol=MEANFLUX * EXPTIME,
+    )
+
+    # Catalog tests to ensure that the sources injected at
+    # NaN pixel center locations are reasonably close
+
+    # Create catalogs of "best" selected recovered sources (only one source per injection)
+    nonan_rs_no_dlbs = results["NoNan"].recovered_sources[
```
schlafly: What's dlbs?

romancal/multiband_catalog/_multiband_catalog.py
```diff
@@ -580,13 +580,32 @@ def make_source_injected_library(library, seed=None):
...
+
             # Inject sources into the detection image
             si_model = inject_sources(
                 si_model,
                 si_cat,
                 seed,
             )
 
+            # Reinstate NaNs after injections
+            # Intentionally leave var_poisson at median for NaN data
```
schlafly: What value did var_poisson formerly have?  Also NaN?  It strikes me as conceptually nicer to restore both the data & the variance; what motivates leaving it at the median?

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+
+    # Create three sets of libraries with two copies of the same mosaic:
+    # NoNan - original mosaics
+    # Grid - NoNan mosaics with the center locations of injected
+    #     sources on a diagonal set to NaN
+    # Block - Grid mosaics with a quadrant also set to NaN
+    libmods = libraries_si_nan
+
+    # Obtain source injection locations
+    with libmods["NoNan"]:
```
schlafly, line 534: Can we avoid repeating the make_source_grid(...) invocation?  between libraries_si_nan and here?  Don't we store an extension with the locations of the injected sources, so we can just read in the injected locations rather than regenerating them?

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+        nonan_rs_no_dlbs["ellipticity"][ngce_idx],
+        atol=0.3,
+    )
+    ngce_idx = ngce_idx[bad_idx_mask]
+    gce_idx = gce_idx[bad_idx_mask]
+
+    # Grid catalog tests
+
+    # Ensure Grid has NaN flux within a 0.4" aperture of NaN pixels
+    assert np.all(np.isnan(grid_rs_no_dlbs["aper04_f158_flux"][gce_idx]))
```
schlafly: I don't think we need to assert this behavior; let's drop this test and equivalents.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+    grid_inter_ce = list(set(nonan_ce) & set(grid_ce))
+    block_inter_ce = list(set(nonan_ce) & set(block_ce))
+
+    # Create numpy indices for each matched catalog object
+    ngce_idx = np.array(
+        [
+            np.where(nonan_rs_no_dlbs["best_injected_index"] == ce)[0][0]
+            for ce in grid_inter_ce
+        ]
+    )
```
schlafly, line 634: I think this is something like 
``
np.flatnonzero(np.isin(nonan_rs_no_dlbs['best_injected_index'], grid_inter_ce))
``
Code along those lines would be more readable.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+    bce_idx = np.array(
+        [
+            np.where(block_rs_no_dlbs["best_injected_index"] == ce)[0][0]
+            for ce in block_inter_ce
+        ]
+    )
+
+    # Remove obvious object mismatch
+    # (SI can have multiple matches for one object, and it chooses the closest.
+    #  NaNs can cause a mismatch.)
```
schlafly, line 656: This is done only for grid and not block indices?

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+    assert np.allclose(
+        results["NoNan"].si_detection_image[~block_nan_stamp_mask],
+        results["Block"].si_detection_image[~block_nan_stamp_mask],
+        atol=MEANFLUX * EXPTIME,
+    )
+
+    # Catalog tests to ensure that the sources injected at
+    # NaN pixel center locations are reasonably close
+
+    # Create catalogs of "best" selected recovered sources (only one source per injection)
```
schlafly, line 602: We have totally blasted the flux away in these regions.  A large galaxy that had a single NaN injected in its vicinity should end up reasonably close; a low SNR star whose core is removed should be pretty bad.  I don't think we want the tests to try to have that level of fidelity and determine values for the tolerances that pass that are quite sensitive to the details of the SI SNRs.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+    block_nan_stamp_mask[240:, 240:] = True
+
+    # Compare data
+
+    # Grid vs NoNan
+    # All pixels far from the NaN injected sources should be close
+    assert np.allclose(
+        results["NoNan"].si_detection_image[~grid_nan_stamp_mask],
+        results["Grid"].si_detection_image[~grid_nan_stamp_mask],
+        atol=MEANFLUX * EXPTIME,
```
schlafly, line 587: Do you remember the units of the detection image?  This absolute tolerance seems weird to me.

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -461,6 +516,215 @@ def test_multiband_source_injection_catalog(
...
+            snr_threshold=15,
+            npixels=4,
+            fit_psf=False,
+            deblend=True,
+            inject_sources=True,
+            inject_seed=RNG_SEED,
+            save_results=False,
+            save_debug_info=True,
+        )
+
```
schlafly, line 568: Starting around here there's a lot of logic involving matching the somewhat complicated NaN assignment in the fixture to the locations where sources should be missing / bad.  I worry this is fragile.  We should try to remove any of that detail from this function.

I think we should end up making this set of tests much simpler.  e.g.: 
1. currently, we always inject sources, so we can test that len(segmentation_map.injected_sources) = 400 for all 3 cases?
2. we can test that len(recovered_sources) is ~75% of the value for the nonan case, due to the missing block?

I don't think we want any tests of the level of damage the NaNs did to the recovered sources; that depends too much on the details of the injected sources and is hard to reason about.

I don't think this is needed, but an additional test along the lines of 'sources away from NaNs look the same in the different injection cases' is reasonable, in particular if you've encountered cases where this doesn't work and so this test would be doing real work.  If you decide we do need such a test, it could look like:
```
distance_to_nan = distance_transform_edt(~np.isfinite(data))
far_from_nan = distance_to_nan[recovered['y'].astype('i4'), recovered['x'].astype('i4')] > X
recovered = recovered[far_from_nan]
match(recovered, nonan_recovered)
assert(close(recovered[segment_flux][match_recovered], nonan_recovered[segment_flux][match_nonan_recovered]))
```
That would at least remove a lot of the specific details of where the NaNs are, but is still a fair amount, and it's hard for me to say what the tolerances should be.  Maybe assert(difference < 3 * uncertainty) would be better.

Is there a clear case you are trying to catch here that can go wrong?  If not, let's simplify this dramatically and just do something like 1 & 2 above.

schlafly (commented):
Sorry, still have a lot of comments.  The code in _injection looks much simpler, great.  I think I have lost much of the thread on the testing code.  I have some suggestions in line but we should probably discuss in person.

I also noticed while doing this that we store a copy of the whole catalog following source injection here:
https://github.com/spacetelescope/romancal/blob/main/romancal/multiband_catalog/multiband_catalog_step.py#L161
which you use in these tests.  Does this catalog make it into the parquet files?  The original doc calls for only storing the recovered sources close to the injected sources (to avoid making another copy of the full catalog).  We should probably remove this and adjust test code to compensate.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and removed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and refactored.

romancal/multiband_catalog/_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: Discussed offline and addressed.

PaulHuwe:
New regression tests pass:
https://github.com/spacetelescope/RegressionTests/actions/runs/32094968167

romancal/multiband_catalog/tests/test_multiband_catalog.py
```diff
@@ -382,26 +405,68 @@ def mosaic_si_model(shape=(500, 500)):
...
 
 
 @pytest.mark.parametrize("fit_psf", (True, False))
 @pytest.mark.parametrize(
     "snr_threshold, npixels, save_results",
     (
-        (3, 10, False),
-        (2, 4, True),
+        (7, 5, False),
+        (15, 4, True),
```
schlafly, line 461: Remind me what triggered these changes?

schlafly (commented):
Thanks, this looks good.  One minor question in line.

romancal/multiband_catalog/tests/test_multiband_catalog.py
PaulHuwe: I changed the flux scale for the test images. Accordingly, I generated a quasi-mcmc script to determine new parameters.

schlafly (approved):
Thanks Paul, looks good to me.  Can you file an issue with a MWE demonstrating the problematic behavior when there is a large masked region, and I'll try to fix it?  Thanks!

PR #2390, opened 2026-07-17
Add read pattern test

schlafly:
This PR adds a new test to make sure that read_pattern is propagated into catalogs / segmentation maps.  It is a nested arrays and so there is some roman_datamodels handling that can cause problems.

Only adds a unit test (which should fail).  Regtests pointing to https://github.com/spacetelescope/roman_datamodels/pull/692 should pass:
https://github.com/spacetelescope/RegressionTests/actions/runs/29610396350

Marking this no changelog entry needed since it only changes test code.  I'll hold this in draft until the corresponding roman_datamodels PR is merged.

## Tasks

### DIFF
--- a/romancal/source_catalog/tests/test_source_catalog.py
+++ b/romancal/source_catalog/tests/test_source_catalog.py
@@ -361,6 +361,35 @@ def test_source_catalog_populates_dust_ebv(model_fixture, request, function_jail
     assert cat["dust_ebv"].dtype == np.float32
 
 
+def test_nested_metadata_propagated_to_catalog_and_segmentation(
+    image_model, function_jail
+):
+    """
+    Nested (list-of-list) metadata such as ``meta.exposure.read_pattern``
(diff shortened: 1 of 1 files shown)


schlafly:
@braingram or @WilliamJamieson , any interest reviewing?  This PR just adds a unit test that currently fails; when copying metadata into the source catalog or segmentation map from the L2 image, some array fields do not survive.  This test demonstrates that problem.  https://github.com/spacetelescope/roman_datamodels/pull/692 intends to fix the issue.  That does work; the regression test run
https://github.com/spacetelescope/RegressionTests/actions/runs/29610396350
passes.  But I haven't looked carefully at the roman_datamodels code I'm changing.

braingram:
Should we consider https://github.com/spacetelescope/romancal/pull/2237 so we can compare the truth metadata to the regtest results?

schlafly:
Ugh, sorry I lost #2237.  Let me respond in the thread there about that one...

PR #2401, opened 2026-07-27
Initial implementation of healpix-aggregated sky summary arrays.

### LINKED ISSUE #2362: Add new step to ELP to compute sky estimates in healpixels in L2 images
@darioflute has a new sky matching algorithm with code available at: https://github.com/darioflute/wfiskymatch/tree/main .  The basic input to this code is estimates of the background level over a large number of healpix pixels, roughly one for every hundred L2 pixels.  Dario computes these around here: https://github.com/darioflute/wfiskymatch/blob/21cda10abb8256458991cd34dc78ab5feecb7074/wfiskymatch/tools.py#L533 .

This code uses background estimates made following source extraction so that sources can be masked.  If we follow that approach, then this step would need access to the SourceCatalog's segment map for masking, in addition to the L2 image.
https://github.com/darioflute/wfiskymatch/blob/21cda10abb8256458991cd34dc78ab5feecb7074/wfiskymatch/tools.py#L414

The basic flow is to mask the sources and then to compute the mean of the unmasked portion of the image in blocks: https://github.com/darioflute/wfiskymatch/blob/main/wfiskymatch/tools.py#L565-L571 , and then to get the value of the block-meaned image at the locations of certain healpixels:
https://github.com/darioflute/wfiskymatch/blob/main/wfiskymatch/tools.py#L607-L608

This is actually rather different than I expected; it's taking the estimates of the 'sky' interpolated at the centers of the healpixels rather than estimating the ~median of the background within the healpixels.  Conceptually I like controlling the measurements to use exactly the same pixels here.

Then the thing that finally gets saved is the value of the mean and error at the healpix indices.
https://github.com/darioflute/wfiskymatch/blob/main/wfiskymatch/tools.py#L617-L619

Dario also stores sparser coverage indices, though it's unclear to me whether those are conceptually necessary, and they can be easily computed from the healpix indices.

My initial proposal would be to add these to the segmentation image computed during the L2 source catalog, since this needs the segmentation image anyway.  Other options include the L2 images themselves, the WfiWcs images, or a new image.  After discussions with Brett, asdf + lazy tree should be able to ~efficiently read these from S3, reading only a header block, a block index block, and the block that actually includes the data.

This ticket is initially to hash out any differences with Dario about how to implement this in romancal.

Dario, can I get your feelings on the following:
- do you think it's important to use your segmentation map rather than the one that comes out of the source catalog step?  I don't want to introduce two of these.
- do you think your approach of interpolating off of the block-reduced background image is better than literally taking the mean of the (masked) pixels that land in each healpix?
- you don't have special handling of edge healpixels; do you want that?
- by default I would include only an analog of your hpimage array, and not your coverage array (since the latter can be computed from the former via something like np.unique(hpx // 2**N).  Am I missing some bit of the importance of that array?

Thanks!

mairanteodoro:
Resolves [RCAL-1420](https://jira.stsci.edu/browse/RCAL-1420)

Closes #2362 

## Summary

This PR implements HEALPix-aggregated sky summary arrays for L2 segmentation-map products, as part of the work to support the downstream sky-matching workflow described in issue #2362.

## Regression tests
- ~https://github.com/spacetelescope/RegressionTests/actions/runs/30657302559/job/91244815212 (07/31/26 - all tests passing)~
- https://github.com/spacetelescope/RegressionTests/actions/runs/33429264503/job/99610377877 (08/31/26 - all tests passing)

### New functionality
- Adds a new `_skyvals.py` module in `source_catalog` with a `compute_skyvals()` function that computes per-HEALPix sky statistics from masked image pixels. For each fine-resolution HEALPix cell ($N_{\text{side}} = 10^{17}$, NEST ordering), it stores:
  - `healpix17`: the NEST-ordered HEALPix index
  - `data`: nanmedian of unmasked, source-free pixel values within the cell
  - `err`: nanmedian of the corresponding error values
  - `covfrac`: fractional pixel coverage of the HEALPix cell by the image footprint
- Also computes `healpix11_cov`: a coarse-resolution ($N_{\text{side}} = 10^{11}$) list of unique HEALPix coverage indices for efficient downstream footprint lookups.
- Handles sparse/edge cases:
  - uses source-free pixels first: `(~bad_pixel_mask) & (segmentation == 0)`
  - falls back to `~bad_pixel_mask` if source-free set is empty
  - returns empty arrays only if no usable pixels remain at all

### Integration with the source catalog step
- Adds a new `compute_skyvals` step parameter (default: `True`) to `SourceCatalogStep` that controls whether sky summary arrays are computed and attached.
- Sky values are computed and attached to the L2 segmentation map model in all code paths (successful detection, no sources detected, and all-pixels-masked fallback), but only for `ImageModel` inputs (i.e., L2, not L3/mosaic).

### Refactoring
- Extracts a shared `get_pixel_area_sr()` utility into `_utils.py`, used by both `RomanSourceCatalog` and the new `compute_skyvals` module.

### Dependencies
- Adds `astropy-healpix >= 1.1.2` as a new package dependency.

### Tests
- Adds three new test cases covering: presence of `skyvals`/`healpix11_cov` arrays in the output; correct behavior when the feature is disabled; and validation of statistical values and coverage fractions on a controlled synthetic image.

### Documentation
- Updates `docs/roman/pipeline/exposure_pipeline.rst`, `docs/roman/source_catalog/arguments.rst`, and `docs/roman/source_catalog/main.rst` to document the new arrays and the `compute_skyvals` parameter.

## Tasks

### DIFF
--- a/romancal/source_catalog/_skyvals.py
+++ b/romancal/source_catalog/_skyvals.py
@@ -36,32 +36,53 @@ def compute_skyvals(input_model, segmentation, bad_pixel_mask):
         2D segmentation labels; source pixels are non-zero.
     bad_pixel_mask : np.ndarray
         Boolean 2D mask where True means "exclude pixel from sky calculation".
+
+    Returns
+    -------
+    skyvals : np.ndarray
+        Structured array with one row per NEST-ordered HEALPix cell at
+        ``nside=2**17``. Its ``healpix17``, ``data``, ``err``, and
+        ``covfrac`` fields contain the cell index, median image value, median
+        error value, and image-footprint coverage fraction, respectively.
+    healpix11_cov : np.ndarray
+        Unique NEST-ordered HEALPix cell indices at ``nside=2**11`` that
+        contain at least one pixel not rejected by the bad-pixel mask,
+        regardless of its segmentation label.
     """
     data = input_model.data
     err = input_model.err
     wcs = input_model.meta.wcs
     pixel_area_sr = get_pixel_area_sr(input_model)
 
+    # Footprint coverage includes all usable pixels, including source pixels.
+    coverage = ~bad_pixel_mask
+    if not np.any(coverage):
+        log.warning(
+            "No usable pixels for skyvals after bad-pixel masking; "
+            "returning empty skyvals/healpix11_cov arrays."
+        )
+        return np.empty((0,), dtype=SKYVALS_DTYPE), np.empty((0,), dtype=np.int64)
+
     # Valid sky pixels are those not rejected by upstream masks and not part
     # of detected sources.
-    valid = (~bad_pixel_mask) & (segmentation == 0)
+    valid = coverage & (segmentation == 0)
     if not np.any(valid):
         # Fallback: if every source-masked pixel is rejected, use all unmasked
         # pixels so we still produce per-exposure sky estimates.
-        valid = ~bad_pixel_mask
-        if not np.any(valid):
-            log.warning(
-                "No usable pixels for skyvals after bad-pixel masking; "
-                "returning empty skyvals/healpix11_cov arrays."
-            )
-            return np.empty((0,), dtype=SKYVALS_DTYPE), np.empty((0,), dtype=np.int64)
+        valid = coverage
 
-    yy, xx = np.nonzero(valid)
-    ra, dec = wcs.pixel_to_world_values(xx, yy)
+    coverage_yy, coverage_xx = np.nonzero(coverage)
+    ra, dec = wcs.pixel_to_world_values(coverage_xx, coverage_yy)
 
     # Convert sky coordinates to NEST-ordered HEALPix indices at nside=2**17.
     healpix = HEALPix(nside=HEALPIX17_NSIDE, order="nested", frame="icrs")
-    healpix17 = healpix.lonlat_to_healpix(ra * u.deg, dec * u.deg).astype(np.int64)
+    coverage_healpix17 = healpix.lonlat_to_healpix(ra * u.deg, dec * u.deg).astype(
+        np.int64
+    )
+    is_sky_pixel = valid[coverage_yy, coverage_xx]
+    yy = coverage_yy[is_sky_pixel]
+    xx = coverage_xx[is_sky_pixel]
+    healpix17 = coverage_healpix17[is_sky_pixel]
 
     vals_data = data[yy, xx]
     vals_err = err[yy, xx]
@@ -98,6 +119,10 @@ def compute_skyvals(input_model, segmentation, bad_pixel_mask):
     skyvals["covfrac"] = covfrac
 
     # Coarse (nside=2**11) coverage index list for quick downstream footprint use.
-    healpix11_cov = np.unique(unique_hp // HEALPIX11_DIVISOR).astype(np.int64)
+    # Source pixels remain part of the footprint even though they are excluded
+    # from the sky-statistic reductions above.
+    healpix11_cov = np.unique(coverage_healpix17 // HEALPIX11_DIVISOR).astype(
+        np.int64
+    )
 
     return skyvals, healpix11_cov
--- a/romancal/source_catalog/_skyvals.py
+++ b/romancal/source_catalog/_skyvals.py
@@ -23,6 +23,9 @@ def _estimate_pixel_area_sr(wcs, shape):
     """
     Estimate image pixel area in steradians at the image center.
     """
+    # Use a local finite-difference estimate at the image center.
+    # This mirrors existing source-catalog conventions that treat pixel area
+    # as a representative scalar when a per-pixel map is not available.
     ycen = (shape[0] - 1) / 2.0
     xcen = (shape[1] - 1) / 2.0
 
@@ -41,6 +44,7 @@ def _get_pixel_area_sr(input_model):
     """
     Return pixel area in steradians from model metadata or estimate from WCS.
     """
+    # Prefer photom-provided pixel area metadata when available and valid.
     if (
         "photometry" in input_model.meta
         and "pixel_area" in input_model.meta.photometry
@@ -48,6 +52,7 @@ def _get_pixel_area_sr(input_model):
         and input_model.meta.photometry.pixel_area > 0
     ):
         return float(input_model.meta.photometry.pixel_area)
+    # Fallback for missing/placeholder metadata values.
     return _estimate_pixel_area_sr(input_model.meta.wcs, input_model.data.shape)
 
 
@@ -69,6 +74,8 @@ def compute_skyvals(input_model, segmentation, bad_pixel_mask):
     wcs = input_model.meta.wcs
     pixel_area_sr = _get_pixel_area_sr(input_model)
 
+    # Valid sky pixels are those not rejected by upstream masks and not part
+    # of detected sources.
     valid = (~bad_pixel_mask) & (segmentation == 0)
     if not np.any(valid):
         return np.empty((0,), dtype=SKYVALS_DTYPE), np.empty((0,), dtype=np.int64)
@@ -76,12 +83,14 @@ def compute_skyvals(input_model, segmentation, bad_pixel_mask):
     yy, xx = np.nonzero(valid)
     ra, dec = wcs.pixel_to_world_values(xx, yy)
 
+    # Convert sky coordinates to NEST-ordered HEALPix indices at nside=2**17.
     healpix = HEALPix(nside=HEALPIX17_NSIDE, order="nested", frame="icrs")
     healpix17 = healpix.lonlat_to_healpix(ra * u.deg, dec * u.deg).astype(np.int64)
 
     vals_data = data[yy, xx]
     vals_err = err[yy, xx]
 
+    # Sort/group by healpixel so per-healpixel reductions are contiguous.
     order = np.argsort(healpix17)
     hp_sorted = healpix17[order]
     data_sorted = vals_data[order]
@@ -93,11 +102,13 @@ def compute_skyvals(input_model, segmentation, bad_pixel_mask):
 
     med_data = np.empty(unique_hp.size, dtype=np.float32)
     med_err = np.empty(unique_hp.size, dtype=np.float32)
+    # Use robust medians for sky and error estimates within each healpixel.
     for i, (start, count) in enumerate(zip(starts, counts, strict=True)):
         section = slice(start, start + count)
         med_data[i] = np.nanmedian(data_sorted[section]).astype(np.float32)
         med_err[i] = np.nanmedian(err_sorted[section]).astype(np.float32)
 
+    # Normalize contributing-pixel counts to fractional HEALPix coverage.
     healpix_area_sr = 4.0 * np.pi / (12.0 * HEALPIX17_NSIDE * HEALPIX17_NSIDE)
     expected_pixels = healpix_area_sr / pixel_area_sr
     covfrac = (counts.astype(np.float64) / expected_pixels).astype(np.float32)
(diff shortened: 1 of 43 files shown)


romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,125 @@
...
+    [
+        ("healpix17", np.int64),
+        ("data", np.float32),
+        ("err", np.float32),
+        ("covfrac", np.float32),
+    ]
+)
+
+
+def _estimate_pixel_area_sr(wcs, shape):
```
schlafly: Can you look if this duplicates code in, e.g., source catalog somewhere?  Or some WCS routine?  It feels like we must already have this.  The logic seems fine; some mildly different choices are possibly if one wants to handle non-orthogonal axes, but that's minor.

romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,125 @@
...
+    ray, decy = wcs.pixel_to_world_values(xcen, ycen + 1.0)
+
+    c0 = SkyCoord(ra0, dec0, unit="deg")
+    cx = SkyCoord(rax, decx, unit="deg")
+    cy = SkyCoord(ray, decy, unit="deg")
+
+    return (c0.separation(cx).radian * c0.separation(cy).radian).item()
+
+
+def _get_pixel_area_sr(input_model):
```
schlafly: I feel like the source catalog code might have this too.

romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,125 @@
...
+    data = input_model.data
+    err = input_model.err
+    wcs = input_model.meta.wcs
+    pixel_area_sr = _get_pixel_area_sr(input_model)
+
+    # Valid sky pixels are those not rejected by upstream masks and not part
+    # of detected sources.
+    valid = (~bad_pixel_mask) & (segmentation == 0)
+    if not np.any(valid):
+        return np.empty((0,), dtype=SKYVALS_DTYPE), np.empty((0,), dtype=np.int64)
```
schlafly: I can't remember what Dario did for this case; can you check?  If this occurs we are unable to estimate backgrounds for this exposure, which feels bad.

romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,125 @@
...
+        hp_sorted, return_index=True, return_counts=True
+    )
+
+    med_data = np.empty(unique_hp.size, dtype=np.float32)
+    med_err = np.empty(unique_hp.size, dtype=np.float32)
+    # Use robust medians for sky and error estimates within each healpixel.
+    for i, (start, count) in enumerate(zip(starts, counts, strict=True)):
+        section = slice(start, start + count)
+        med_data[i] = np.nanmedian(data_sorted[section]).astype(np.float32)
+        med_err[i] = np.nanmedian(err_sorted[section]).astype(np.float32)
```
schlafly, line 85: Can you profile this to make sure it's not adding an appreciable amount of time to the step?  It seems likely fine to me, but I don't have a strong sense.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -383,6 +393,57 @@ def test_l2_input_model_unchanged(image_model, function_jail):
...
+        fit_psf=False,
+        compute_skyvals=False,
+    )
+
+    assert isinstance(result_segmentation_map, SegmentationMapModel)
+    assert "skyvals" not in result_segmentation_map
+    assert "healpix11_cov" not in result_segmentation_map
+
+
+def test_l2_skyvals_empty_when_segmentation_masks_everything(image_model):
```
schlafly: I actually don't know what behavior we want in this case; let's drop this test.

romancal/source_catalog/tests/test_source_catalog.py
```diff
@@ -383,6 +393,57 @@ def test_l2_input_model_unchanged(image_model, function_jail):
     assert_equal(original_err, image_model.err)
 
 
+def test_l2_segmentation_contains_skyvals(image_model):
```
schlafly, line 395: It would be good to have a test that tests the values.  I forget how realistic the image in image_model is.  But, e.g., if the image were randn() + offset, checking that the skyvals were roughly equal to the offset would be good.  Some covfracs should be ~1; all should be between 0 and 1.

schlafly (commented):
This looks good to me.  I have some comments inline.  When you're ready it would be good to run your data release simulation with this turned on and point Dario at the results so that he can take a look to see if things are behaving sanely.

romancal/source_catalog/_skyvals.py
mairanteodoro: I found this functionality in `RomanSourceCatalog._pixel_area`. I removed the duplicated code in this PR.

romancal/source_catalog/_skyvals.py
mairanteodoro: Yup.

romancal/source_catalog/tests/test_source_catalog.py
mairanteodoro: Done.

romancal/source_catalog/tests/test_source_catalog.py
mairanteodoro: Done.

romancal/source_catalog/_skyvals.py
mairanteodoro: Profiling results on my local machine for compute_skyvals/SourceCatalogStep overhead. Executed by AI.

Setup:
- synthetic L2-like `ImageModel` (4088 x 4088)
- background N(100, 1) + a few compact Gaussian sources
- `fit_psf=False` for `SourceCatalogStep` timing to isolate `skyvals` overhead
- `SourceCatalogStep` params: `bkg_boxsize=256`, `kernel_fwhm=2.0`, `snr_threshold=5.0`, `npixels=25`

Results:
- `compute_skyvals` alone (avg of 3 runs, after warmup): 3.036 s
- output sizes: `skyvals=65162` rows, `healpix11_cov=30` rows
- `SourceCatalogStep.call(..., compute_skyvals=False)` (avg of 2 runs): 2.514 s
- `SourceCatalogStep.call(..., compute_skyvals=True)` (avg of 2 runs): 5.603 s
-  Increment from `skyvals`:
   - +3.089 s absolute
   - +122.9% relative to `compute_skyvals=False` in this synthetic benchmark

Notes:
- This is a synthetic-image micro-benchmark and should be treated as order-of-magnitude guidance.
- The absolute delta is consistent with the helper-only timing (~3.0 s), so most added cost appears to be from `compute_skyvals` itself.

romancal/source_catalog/_skyvals.py
mairanteodoro: It seems that Dario's implementation only masks out point sources (plus bad pixels, of course) and proceeds with estimating the background. This is the [snippet that does that](https://github.com/darioflute/wfiskymatch/blob/21cda10abb8256458991cd34dc78ab5feecb7074/wfiskymatch/tools.py#L564-L578) (mask, block-reduce, and inpaint):
```python
    # 2. Mask point sources and bad pixels
    mask = maskSources(data, dq=dq, fwhm=1, areathreshold=1000)    
    # 3. Image and errors are block reduced by a factor nblock
    from astropy.nddata import block_reduce
    data[mask] = np.nan
    nblock = 7*2
    data = block_reduce(data, nblock, func=np.nanmean)
    err = block_reduce(err, nblock, func=np.nanmean)
    # Impainting
    from maskfill import maskfill
    idx = ~np.isfinite(data)
    if np.sum(idx) > 0:
        mask1 = np.zeros(np.shape(data))
        mask1[idx] = 1
        data,_ = maskfill(data, mask1, operator='median' )
```
After that, the code proceeds to calculate the Healsparse map.

I made a small change to the current implementation to have a similar behavior to Dario's by falling back to exclude only the masked pixels (i.e. `~finite`, nonpositive err, `DQ DO_NOT_USE`). Only if all pixels are bad do we return an empty array.
```python
    # Valid sky pixels are those not rejected by upstream masks and not part
    # of detected sources.
    valid = (~bad_pixel_mask) & (segmentation == 0)
    if not np.any(valid):
        # Fallback: if every source-masked pixel is rejected, use all unmasked
        # pixels so we still produce per-exposure sky estimates.
        valid = ~bad_pixel_mask
        if not np.any(valid):
            log.warning(
                "No usable pixels for skyvals after bad-pixel masking; "
                "returning empty skyvals/healpix11_cov arrays."
            )
            return np.empty((0,), dtype=SKYVALS_DTYPE), np.empty((0,), dtype=np.int64)

```

romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Parameters
+    ----------
+    input_model
+        Roman image model containing data/err arrays and metadata.
+    segmentation : np.ndarray
+        2D segmentation labels; source pixels are non-zero.
+    bad_pixel_mask : np.ndarray
+        Boolean 2D mask where True means "exclude pixel from sky calculation".
+    """
```
schlafly, line 51: Can we add to the docstring the things that are returned and what they are?

romancal/source_catalog/_skyvals.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    skyvals = np.empty(unique_hp.size, dtype=SKYVALS_DTYPE)
+    skyvals["healpix17"] = unique_hp
+    skyvals["data"] = med_data
+    skyvals["err"] = med_err
+    skyvals["covfrac"] = covfrac
+
+    # Coarse (nside=2**11) coverage index list for quick downstream footprint use.
+    healpix11_cov = np.unique(unique_hp // HEALPIX11_DIVISOR).astype(np.int64)
```
schlafly: This is all sane, but could you check that your and Dario's code match in the following respect: this code provides a healpix11_cov index for every healpix11 pixel that contains a 'valid' pixel.  If there are pixels that are good but have sources in them, they don't count as valid and don't count as possible 'coverage' pixels.  That's a little weird for a 'coverage' map.  It's fine if Dario's code is the same but we should make sure that's also the behavior he has.

schlafly (commented):
Looks good, a couple more minor comments.

romancal/source_catalog/_skyvals.py
schlafly: Okay, thanks, your handling looks reasonable.

romancal/source_catalog/_skyvals.py
mairanteodoro: Dario's implementation uses the [coarse footprint to calculate the coverage](https://github.com/darioflute/wfiskymatch/blob/21cda10abb8256458991cd34dc78ab5feecb7074/wfiskymatch/tools.py#L584-L625):
```python
    hsp_map = hspm.HealSparseMap.make_empty(nside_coverage, nside_sparse, dtype=np.float64)
    [...]
    px_num = hp.ang2pix(hsp_map._nside_sparse, ra, dec, nest=True, lonlat=True)
    upx = np.unique(px_num)
    [...]
    hsp_map.update_values_pix(upx, values.astype(np.float64))
    [...]
    idcoverage = np.where(hsp_map.coverage_map>0)[0]  # Select covered tiles
    [...]
```
The current implementation was using only the samples used for sky statistics, which would cause the bug that you just mention (where a 'valid' pixel with a source in it would be ignored). So, I aligned our implementation with Dario's and now `healpix11_cov` comes from all non-bad pixels only, while `skyvals.data`, `err`, and `covfrac` still use source-free pixels.

romancal/source_catalog/_skyvals.py
mairanteodoro: Done.

schlafly (approved):

schlafly:
Can you give me an updated regtest and I'll merged?  Can't imagine it will make a difference but July is a long time ago.

PR #2407, opened 2026-08-04
Implement source catalog reference prefetch.

### LINKED ISSUE #2338: Source catalog step reference file usage & prefetching
Eric Younkin noticed that reference files needed only by the source catalog step were not being prefetched by CRDS.  I suspect that the prefetching mechanism is triggered by lines like these:
https://github.com/spacetelescope/romancal/blob/main/romancal/saturation/saturation_step.py#L33
which the source catalog step does not use.  The source catalog step uses the aperture reference file, the dust map reference file, and the PSF reference file.  We should try to add those to the step analogously to their use in the saturation step.

@stscieisenhamer , does that look like the right mechanism to you?

mairanteodoro:
Resolves [RCAL-1401](https://jira.stsci.edu/browse/RCAL-1401)

Closes #2338

Depends on the CRDS ticket: spacetelescope/crds#1227.

## Summary (co-authored by Copilot)
This PR adds source-catalog reference prefetch support so required references are validated earlier and dust-map lookup is reused during catalog generation.

## What changed
1. `SourceCatalogStep.reference_file_types` now includes both `apcorr` and `epsf`.
2. Added `SourceCatalogStep._prefetch_dust_map_paths()` to resolve and validate north/south dustmap CRDS references up front.
3. Added `SourceCatalogStep.get_ref_override()` logic to skip ePSF prefetch (`"N/A"`) for non-imaging exposures.
4. `RomanSourceCatalog` now accepts optional pre-resolved `dust_map_paths` and uses them in `dust_ebv` before falling back to CRDS lookup.
5. Converted `RomanSourceCatalog._get_sfd_map_paths()` to a classmethod and updated internals accordingly.
6. Wired prefetched dust-map paths into both normal and forced catalog generation flows.
7. Added tests for:
   - reference file type list includes `epsf`
   - SFD map path resolution for north/south poles
   - dustmap prefetch path validation
   - behavior when dustmap prefetch returns `None` (catalog still produced with finite `dust_ebv`)

## Notes
- Dustmap prefetch failures are handled with a warning and safe fallback behavior.
- ePSF prefetch is limited to imaging exposures to avoid unnecessary/non-applicable reference access.

## Tasks


braingram:
@mairanteodoro I see a request for review but this is draft. Is this ready for review?

mairanteodoro:
@braingram I was waiting for the regtests to finish, but then we got some issues going on with them. ~So, this is good for review (meaning that all unit tests passed) pending passing the regtests.~ Spoke too soon... there's one failing unit test.

romancal/source_catalog/source_catalog_step.py
```diff
@@ -114,8 +116,64 @@ def save_model(self, model, **kwargs):
 
         return super().save_model(model, **kwargs)
 
+    def _prefetch_dust_map_paths(self):
```
braingram, line 119: Is there something preventing overriding _get_crds_parameters and using the existing prefetch?

mairanteodoro: Not really. I will reformat it to override that.

braingram:
Looking at this in detail there are a few issues to contend with.

1) stpipe only actually prefetches for pipelines ([not steps](https://github.com/spacetelescope/stpipe/blob/f33cf4c78960979185a32c5e22ead27685ba323a/src/stpipe/step.py#L839)) so we could match this behavior with the custom prefetching if it fetches the dustmap and epsf files for pipeline runs (not step runs).

2) The prefetching code in stpipe expects a 1-to-1 mapping of reftype and file. This doesn't work for dustmap since we have a 1-to-2 mapping.

3) Some consideration of log messages is likely called for as those will be visible to users/logs.

4) Some consideration of override handling is related (this doesn't work on main and perhaps we just document that the dustmap reference file cannot be overridden).

If we take the typical approach of adding dustmap to the reference_file_types we'll see the following log message:
```
2026-08-11 12:38:40,613 - stpipe.pipeline - INFO - Prefetch for DUSTMAP reference file is 'N/A'.
```
for an elp run since the default `_get_crds_parameters` doesn't contain the required ngp/sgp pole. We can't just override this as the default is needed for the other reftypes (and stpipe only calls this once per prefetch, not once per reftype).

One possible option might be to override `prefetch` in RomanStep to:
- check if `dustmap` is in `self.reference_file_types`, accept that this will result in a `N/A` message as above (or monkeypatch `reference_file_types` to exclude dustmap but that seems less-than-ideal)
- separately fetch the `dustmap` references with code matching what's called in `Pipeline._precache_references`, emitting log messages for the fetched reference files

I'd like to understand why `epsf` requires additonal custom code here checking for `exp_type`. Why isn't this handled by CRDS (with an exptype selector)?

stscieisenhamer:
The dustmap reftype has no dependency on exptype, so there should be nothing to do with respect to exptype?

Otherwise, as stated, the issue is the 1-exposure-to-many-reffiles-per-reftype requirement. This is actually a CRDS issue with only returning one reference per reftype.

Suggestion is the following plan:
- file a CRDS issue to consider how to return multiple files per reftype/selector match
- until the above is resolved, do not worry about the prefetch

The prefetch is a convenience to help ensure that all references are available when the pipeline is run. If there is overwhelming desire to prefetch, then the explicit requests will need be made.

mairanteodoro:
@stscijgbot-rstdms

mairanteodoro:
Cross-linking this PR to the CRDS ticket: spacetelescope/crds#1227.

schlafly:
FWIW, I consider prefetching to be a nice-to-have rather than a need-to-have.  So I'm content with pre-fetching not working for the dust map pending some eventual CRDS upgrade.

Issue #1386, opened 2024-08-27
Add end-to-end, SDP-lite regtest


schlafly:
In order to check out the full SDP interface, we could make a new regtest / smoketest that runs all of the ELP / HLP steps.

This is annoying in that it would be slow.  We would want to do something like the following:
- take two overlapping uncal files (e.g., existing 0001 and 0002 imaging files) for WFI01
- run the ELP on both through source catalogs (i.e., skipping tweakreg, as the SDP does), creating "2a" files that aren't archived
- make an association file for the "2a" files
- run tweakreg / the ELP on the "2a" files to produce the final cal files
- make an association file for the cal files for a particular skycell
- run the mosaic pipeline on the association

It's not obvious to me that I would want to do very much with the resulting files (e.g., compare_asdf, etc.); I might continue to leave that to the existing set of regtests.  But it's at least important to know that all of those steps run and to have a place with a SDP-lite pipeline that the real SDP pipeline can emulate.  This would make us more aware when the SDP commands need to change slightly to accommodate changes to the pipeline.

schlafly:
While checking a new version of the release I did the following as a kind of SDP-lite set of invocations:
```
# make level "2a" files
strun roman_elp r0000101001001001001_01101_0001_WFI01_uncal.asdf --steps.tweakreg.skip=True --suffix=sourcedetection
strun roman_elp r0000101001001001001_01101_0002_WFI01_uncal.asdf --steps.tweakreg.skip=True --suffix=sourcedetection
# make association file linking these
asn_from_list *_WFI01_sourcedetection.asdf --product-name aoeu -o asn_sourcedetection.json
# make final cal files
strun romancal.step.TweakRegStep asn_sourcedetection.json --suffix=cal
# make association file for cal files
asn_from_list *_cal.asdf --product-name=r0000101001001001001_F158_visit_r274dp63x32y80 -o mosaic.json
# this file then needs to get hand edited to add the target skycell name to the "target" bit of the asn file
# i.e., add target="r274dp63x32y80" to the json file
# make final mosaic product
strun roman_mos mosaic.json
```

This makes the following files:
```
asn_sourcedetection.json  # association file we made for source detection files, could be ignored?
mosaic.json  # association file we made for cal files, could be ignored?
r0000101001001001001_01101_0001_WFI01_cal.asdf  # L2 file
r0000101001001001001_01101_0001_WFI01_cat.asdf  # corresponding cat file
[...]

PR #2408, opened 2026-08-05
Optionally populate tweakreg_catalog in associations made with asn_from_list

schlafly:
This PR adds a feature to asn_from_list that makes it easy for SDP to make associations including tweakreg source catalog metadata in the way that they're used to.  asn_from_list gains a new argument that populates tweakreg_catalog in an association with a name inferred from the image file names using our usual approach (roughly, _cal.asdf -> _cat.parquet), so that associations have both images and corresponding catalogs.

In B23 we reworked how the source catalog is returned (https://github.com/spacetelescope/romancal/pull/2352) and this impacts how SDP does tweakreg for the 18 detector case (see e.g. discussion around https://github.com/spacetelescope/romancal/issues/1386#issuecomment-4927778009).  We didn't fully resolve that discussion and I'm putting in this change as a PR to smooth things over for SDP in B23; we should still consider better longer-term solutions.

Regression test here: https://github.com/spacetelescope/RegressionTests/actions/runs/31015799188
All green.

## Tasks

### DIFF
--- a/changes/2408.associations.rst
+++ b/changes/2408.associations.rst
@@ -1,0 +1,2 @@
+Add a provisional ``asn_from_list`` option that records source catalog names in
+association members, so that tweakreg can be run on already calibrated images.
--- a/romancal/associations/asn_from_list.py
+++ b/romancal/associations/asn_from_list.py
@@ -67,16 +67,13 @@ def _add_tweakreg_catalogs(asn):
     """Add a source catalog name to every member of an association.
 
     ``tweakreg`` uses a member's ``tweakreg_catalog`` attribute as the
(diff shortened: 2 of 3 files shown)


stscieisenhamer (approved):
Approving since it is a stop-gap.

For future discussion: Would it be an issue to put the catalog in the L2 itself, as opposed to/in addition to having a separate file?

schlafly:
Thanks Jonathan.

> For future discussion: Would it be an issue to put the catalog in the L2 itself, as opposed to/in addition to having a separate file?

Most users of catalogs don't also want the images, so it's awkward to bundle them together.  That was however the early design.

PR #2419, opened 2026-08-18
convert "all saturated" regtests to a unit test

braingram:
Replace regression tests that check ELP behavior for all saturated input with a unit test that does the same.

This test runs in a few seconds and this replacement allows cleaning up:
- the regtest file generation script
- removal of a regtest input and truth file
- cleanup of the elp regtest submodule

Regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/32162305695

## Tasks


schlafly (approved):
Thanks Brett.  I agree this looks better as a unit test.  I worried that there might be DMS artifacts that ~demanded a regression test, but we don't link any and they would be dumb artifacts, so I'm going ahead and merging.  Thank you!

PR #2426, opened 2026-08-27
Source catalog pixel area

schlafly:
The source catalog converted surface brightness to flux, and pixel measurements to sky units, using a single image-wide pixel area and a single north position angle taken at the image center.  This PR changes that handling to instead use per-source pixel scales and directions to north.

For L2 images the pixel scale varies by roughly 1% across an image, so this PR fixes a ~1% error in the photometry.  As a demonstration of this, the following image shows a comparison of the derived PSF fluxes wrt truth for a romanisim simulation.  
image
Using main, there is a clear gradient across the field of roughly 1%; with this branch, that gradient disappears.

I also updated the apertures to use apertures of the same size on the sky across the image, accounting for the different pixel scale, rather than using the fixed pixel scale.  I did not make the apertures elliptical (i.e., I did not account for the fact that pixels are ~rectangular rather than square), but the apertures now have the same size on the sky regardless of where they are in the image.  This is similarly only a ~1% effect.  That change is a little annoying because photutils can't vectorize a large number of slightly differently sized apertures in a good way, so I instead defined a number of apertures of slightly different sizes and used the aperture that was closest in size to the needed one.  I capped the number of bins at 64, and the spread in pixel size is 2% for typical WFI detector, so the resulting apertures vary in size by about 0.1%.  

Importantly, our current angle computations misbehaved at the pole; this is now fixed.  We work in 3D cartesian coordinates for the spherical geometry stuff in order to avoid issues.  The spherical_geometry package doesn't have the right vectorization for us to be able to efficiently bring anything in from there, unfortunately.

I replaced some source catalog pixel area, scale, and wcs angle helper functions with new per-source versions.  I consolidated some test WCS construction code and added tests, mostly of the new wcs angle routines.  Docs were updated and extended.  

Regression tests are here: https://github.com/spacetelescope/RegressionTests/actions/runs/33097046230
They show only very small changes to the WCS fit results due to now accounting for pixel scale variations.
New run following review: https://github.com/spacetelescope/RegressionTests/actions/runs/34978372700

## Tasks


schlafly:
@larrybradley , would you mind reviewing this one?

docs/roman/source_catalog/main.rst
```diff
@@ -101,6 +125,16 @@ and the effect of the image drizzling on the PSF. Finally, the PSF is
 azimuthally averaged to remove any azimuthal signatures, which will be
 different in the coadded product than in the individual input exposures.
 
+Surface Brightness to Flux Conversion
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Level 2 and Level 3 pixel values are surface brightnesses, not fluxes.
+Recovering the flux entering a pixel therefore requires multiplying
+its surface brightness by the solid angle of that particular pixel.
+The solid angle is computed directly from the WCS.    Note that the
```
larrybradley:
```suggestion
The solid angle is computed directly from the WCS. Note that the
```

romancal/source_catalog/_aperture.py
```diff
@@ -294,27 +378,40 @@ def calc_aperture_photometry(self, subtract_local_bkg=False):
...
+                CircularAperture(self.xypos_finite[idx], radius / scale)
+                for radius in radii_arcsec
+            ]
+            phot = aperture_photometry(self.model.data, apertures, error=self.model.err)
+            for i, aperture in enumerate(apertures):
+                values = phot[f"aperture_sum_{i}"]
+                if subtract_local_bkg:
+                    # Subtract the local background measured in the annulus
+                    areas = aperture.area_overlap(self.model.data)
+                    values = values - self.aper_bkg_flux[idx] * areas
```
larrybradley:
```suggestion
                    values = values - (self.aper_bkg_flux[idx] * self._source_pixel_area[idx] * areas)
```
`aper_bkg_flux` has SB units. This is a pre-existing bug (but `subtract_local_bkg` is never actually True so this branch is not currently used).

romancal/source_catalog/_source_catalog.py
```diff
@@ -189,7 +189,9 @@ def __init__(
             self.l2_to_sb = self.model.meta.photometry.conversion_megajanskys
         else:
             self.l2_to_sb = 1.0
-        self.sb_to_flux = (1.0 * (u.MJy / u.sr) * self._pixel_area).to(self.flux_unit)
+        self.sb_to_flux = (1.0 * (u.MJy / u.sr) * self._pixel_area_map).to(
+            self.flux_unit
+        )
```
larrybradley: `sb_to_flux` is currently used only in `_validate_and_convert_units`. We could save some memory (e.g., for multiband catalogs) by calculating it in that function and not storing this (now array) in the catalog instance.

romancal/source_catalog/_aperture.py
```diff
@@ -37,9 +39,10 @@ class ApertureCatalog:
         coordinates. Non-finite positions must already have been
         replaced with finite placeholders by the caller.
 
-    pixel_scale : `~astropy.units.Quantity`
-        The pixel scale (with arcsec-compatible units), used to convert
-        the configured radii from arcsec to pixels.
+    pixel_area_map : `~astropy.units.Quantity`
+        2D array of per-pixel solid angles matching the shape of
+        ``model.data``. Used to convert the annulus background back to
+        a surface brightness.
```
larrybradley: It's also used to set the per-source aperture radii.

romancal/source_catalog/tests/test_wcs_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    Sizes must be converted with the pixel area at each source, not a
+    single image-wide value.
+
+    Two identical sources are placed where the pixel area differs by 10%;
+    their reported sizes must differ by the corresponding 5% in linear
+    measure, and their pixel-frame sizes must be identical.
+    """
+    from photutils.segmentation import SegmentationImage
+
+    from romancal.source_catalog._segment import SegmentCatalog
```
larrybradley: Move these imports to the top of the file.

larrybradley (changes requested):
@schlafly Overall this looks good! There's one (pre-existing) bug on unused code (worth fixing here), one memory suggestion, and a couple of nits.

romancal/source_catalog/_aperture.py
schlafly: Thanks, I followed your suggestion.

larrybradley (approved):
Thanks, LGTM!

schlafly:
Merging and okifying, thanks Larry.

PR #2294, opened 2026-05-01
Update refpix.py - bad reference pixel masking

protonchain:
⚠️ ON HOLD: Pending RMO approval (expected by Monday).

Updating refpix to handle masking bad reference pixels, which will be useful to correct WFI03 striations and set up for accepting reference pixel DQ flagging

Resolves [RSOC-DR462](https://jira.stsci.edu/browse/RSOCDR-462)

Closes #

This PR addresses ...

## Tasks
- [ ] **request a review from someone specific**, to avoid making the maintainers review every PR
- [ ] add a build milestone, i.e. `24Q4_B15` (use the [latest build](https://github.com/spacetelescope/romancal/milestones) if not sure)
- [ ] Does this PR change user-facing code / API? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see [changelog readme](https://github.com/spacetelescope/romancal/blob/main/changes/README.rst) for instructions)
    - if your change breaks existing functionality, also add a `changes/<PR#>.breaking.rst` news fragment
  - [ ] update or add relevant tests
  - [ ] update relevant docstrings and / or `docs/` page
  - [ ] [start a regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) and include a link to the running job ([click here for instructions](https://github.com/spacetelescope/RegressionTests/blob/main/docs/running_regression_tests.md))
    - [ ] Do truth files need to be updated ("okified")?
      - [ ] **after the reviewer has approved these changes**, run `okify_regtests` to update the truth files
- [ ] if a JIRA ticket exists, [make sure it is resolved properly](https://github.com/spacetelescope/romancal/wiki/How-to-resolve-JIRA-issues)

PR #2427, opened 2026-08-28
Reduce refpix memory usage

### LINKED ISSUE #2429: When should border_ref_pix and amp33 be updated?
We currently populate the border pixels with copies of the edges of the detector in dq_init.  I would have said that we then never touched them and passed them through the remainder of the steps, but in the reference pixel correction we update the left, right, and amp33 pixels:
https://github.com/spacetelescope/romancal/blob/fec57c963e7a1d3be941cd9dbaa0b850bc86a9bf/romancal/refpix/_data.py#L108-L109
while leaving the top and bottom ones alone.  This seems inconsistent enough to me that I'm removing it in #2427, but it's worth discussing what's desired.  I think there are two sane approaches:
1. pass the border pixels & amp33 pixels untouched all the way through
2. treat the border pixels & amp33 pixels as views of the data, and make all corrections on them (wfi18_transient, dark_decay, ...) until they get peeled off and fixed in ramp fitting.

Somewhat relatedly, the way reference pixel updates the left and right pixels is as views of data, so today the left and right pixels get mutated by steps like wfi18_transient and dark_decay, but the top, bottom, and reference pixels do not.  #2427 stops that so that they're at least consistently treated.

@tddesjardins , is it clear to you which option is desired?

schlafly:
The refpix step is currently the high water mark for memory usage in exposure level pipeline runs; see e.g. this plot
image
from Dave's memory tests.  It doesn't need to be, however.  This PR changes the processing to go one resultant at a time instead of processing all of the resultants simultaneously.  The only step that couples the different resultants together is the offset computation, which can be performed early and saved without requiring holding multiple copies of all of the resultants in memory.

For an example 55 resultant TVAC file, this reduces memory usage from 30 GB to 5 GB.  For 5 resultants the difference is more like 5 GB -> 2 GB.

I stumbled onto an issue while doing this involving the reference pixels.  We save copies of the border pixels to separate arrays during the dq_init step.  The reference pixel correction was updating the values of the left and right border pixels but not the top and bottom ones.  It also updated the amp33 pixels, but left them as amp33.  I've changed this to leave all border pixels untouched.  This doesn't affect other things in the pipeline but is worth a separate discussion.  As a separate issue, the way the left and right border pixels were being assigned before this PR were as views into data, and so these arrays were getting mutated by later steps like dark subtraction and linearity.  It's honestly not clear to me what is desired and I've filed #2429 to try to resolve this; we don't use these saved border pixel arrays elsewhere.  

Resolves #2429.

One unrelated fix: a new argument was added to run_steps in #2294, but run_steps(...) callers were called with all positional arguments and wasn't updated, so some non-default parameter settings were getting jumbled arguments.  I updated some callers to use keyword arguments instead.

Regression test started here: https://github.com/spacetelescope/RegressionTests/actions/runs/33206169144
Tests show the expected changes in the left & right border pixels, as well as amp33 for TVAC data.  No changes to data.

## Tasks

### DIFF
--- a/changes/2427.refpix.rst
+++ b/changes/2427.refpix.rst
@@ -1,7 +1,3 @@
-Correct the reference pixels one resultant at a time rather than holding the
-whole ramp in the padded channel view, cutting peak memory usage of the step
-from roughly 6.5x the size of the ramp to well under 0.5x (29.7 GB to 5.4 GB
-for a 55-resultant TVAC ramp) while leaving the results bit-for-bit unchanged.
-Also fix the arguments passed from ``RefPixStep`` to ``run_steps``, which were
-each shifted by one position, so that (for example) ``fft_interpolate=False``
-no longer turns off the cosine interpolation instead.
+Reduce peak memory usage of the reference pixel correction step dramatically by
+processing resultants individually. Also fix the arguments passed from
+``RefPixStep`` to ``run_steps``, which were each shifted by one position.
--- a/changes/2427.refpix.rst
+++ b/changes/2427.refpix.rst
@@ -1,0 +1,7 @@
+Correct the reference pixels one resultant at a time rather than holding the
+whole ramp in the padded channel view, cutting peak memory usage of the step
+from roughly 6.5x the size of the ramp to well under 0.5x (29.7 GB to 5.4 GB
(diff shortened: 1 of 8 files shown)


schlafly:
@WilliamJamieson , mind taking a look?

WilliamJamieson (approved):

PR #2428, opened 2026-08-28
RCAL-1435 Add crds reference read for BAM to set_telescope_pointing

stscieisenhamer:
Resolves [RCAL-1435](https://jira.stsci.edu/browse/RCAL-1435)

This PR addresses the following: Original design for set_telescope_pointing was to read the Boresight Adjustment Matrix (BAM) from the engineering telemetry. During final testing, it was discovered that the relevant mnemonics are only present in the sporadic engineering stream. This stream is not available in real-time to the ground system while processing exposures.  The solution was to instead place these values in a CRDS reference file, the "bam" reftype. `set_telescope_pointing` now uses this reference. If unable to retrieve the reference, the pre-launch values are used.

## Tasks

### DIFF
--- a/romancal/orientation/tests/data/bam_ref_alternate.asdf
+++ b/romancal/orientation/tests/data/bam_ref_alternate.asdf
@@ -1,0 +1,26 @@
+#ASDF 1.0.0
+#ASDF_STANDARD 1.6.0
+%YAML 1.1
+%TAG ! tag:stsci.edu:asdf/
+--- !core/asdf-1.1.0
+asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
+  name: asdf, version: 5.3.1}
+history:
+  extensions:
+  - !core/extension_metadata-1.0.0
+    extension_class: asdf.extension._manifest.ManifestExtension
+    extension_uri: asdf://asdf-format.org/core/extensions/core-1.6.0
+    manifest_software: !core/software-1.0.0 {name: asdf_standard, version: 1.5.0}
+    software: !core/software-1.0.0 {name: asdf, version: 5.3.1}
+fgs_tbl: {Qb1: 0.6822141509826322, Qb2: -0.1800546332580956, Qb3: 0.6837984564491885,
+  Qb4: -0.18596734175399293}
+meta:
+  author: T. Sohn, T. Desjardins
+  description: Quaternion values for the alignment matrix between the Roman observatory
+    boresight and fine guidance system (FGS).
+  instrument: {name: WFI}
+  origin: STSCI/SOC
+  pedgriee: GROUND
+  reftype: bam
+  useafter: '2020-01-01T00:00:00'
+...
--- a/romancal/orientation/tests/data/bam_ref_alternate.asdf
+++ b/romancal/orientation/tests/data/bam_ref_alternate.asdf
@@ -20,7 +20,8 @@ roman:
                  boresight and fine guidance system (FGS).
     instrument: {name: WFI}
     origin: STSCI/SOC
-    pedgriee: GROUND
-    reftype: bam
+    pedigree: GROUND
+    reftype: BAM
+    telescope: ROMAN
     useafter: '2020-01-01T00:00:00'
 ...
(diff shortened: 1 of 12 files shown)


romancal/orientation/tests/data/bam_ref_alternate.asdf
```diff
@@ -0,0 +1,26 @@
...
+history:
+  extensions:
+  - !core/extension_metadata-1.0.0
+    extension_class: asdf.extension._manifest.ManifestExtension
+    extension_uri: asdf://asdf-format.org/core/extensions/core-1.6.0
+    manifest_software: !core/software-1.0.0 {name: asdf_standard, version: 1.5.0}
+    software: !core/software-1.0.0 {name: asdf, version: 5.3.1}
+fgs_tbl: {Qb1: 0.6822141509826322, Qb2: -0.1800546332580956, Qb3: 0.6837984564491885,
+  Qb4: -0.18596734175399293}
+meta:
```
braingram: Shouldn't this be nested under a roman key?

stscieisenhamer: True, it will be in a schema eventually. Will fix and make appropriate changes in the reference file specification.

stscieisenhamer:
[regtest #1](https://github.com/spacetelescope/RegressionTests/actions/runs/33203553858)
Result: pass

[regtest #2](https://github.com/spacetelescope/RegressionTests/actions/runs/33676200201)
reason: new reference format
Result: cannot run due to unrelated issue apparently with roman_photoz.

romancal/orientation/tests/data/bam_ref_alternate.asdf
```diff
@@ -0,0 +1,26 @@
...
+  fgs_tbl: {Qb1: 0.6, Qb2: -0.1, Qb3: 0.6, Qb4: -0.1}
+  meta:
+    author: SCSB with modified values for testing
+    description: Quaternion values for the alignment matrix between the Roman observatory
+                 boresight and fine guidance system (FGS).
+    instrument: {name: WFI}
+    origin: STSCI/SOC
+    pedgriee: GROUND
+    reftype: bam
+    useafter: '2020-01-01T00:00:00'
```
schlafly, line 29:
```suggestion
roman:
  fgs_tbl: {Qb1: 0.6, Qb2: -0.1, Qb3: 0.6, Qb4: -0.1}  # though an ndarray with a specified dtype seems like it would make more sense.
  meta:
    author: SCSB with modified values for testing
    description: Quaternion values for the alignment matrix between the Roman
                 observatory boresight and fine guidance system (FGS).
    instrument: {name: WFI}
    origin: STSCI/SOC
    pedigree: GROUND
    reftype: BAM
    telescope: ROMAN
    useafter: !time/time-1.4.0 2020-01-01T00:00:00.000
```

I don't think this really matters, but some things will come up when we try to actually add files to rad.  telescope: ROMAN is needed; other reftypes have all-caps BAM, pedgriee -> pedigree typo fix, time preix on the time string.  But since the code path really only needs to exercise af['roman']['fgs_tbl']['Qb1'], this is window dressing.  If I were designing the file I would make the fgs_table an array rather than a dictionary (!core/ndarray-1.1.0 {data: [0.6, -0.1, 0.6, -0.1], datatype: float64, shape: [4]}) since that seems more natural for a quaternion, but I'll defer to you.  It's also worth noting that this quaternion is not normalized so it doesn't correspond to a pure rotation.

I'll let you decide what you want to do here, thanks.

schlafly (approved):
I left a comment in line about the file structure; I think we could imagine different forms there, and some metadata is weird, but I don't think it needs to block the PR and I'll let you decide what you want to do.

romancal/orientation/tests/data/bam_ref_alternate.asdf
stscieisenhamer: About the caps/syntax: will fix (the pedigree misspell was identified in a crds test earliear).

About the dictionary: A, admittedly very short, discussion was there an the knee-jerk decision was to make a dictionary to be clear between what the values and how they would be found in the telemetry. Will bring up in the crds issue.

schlafly: I'll defer to you; two cents only.

stscieisenhamer: After discussions, the format will be changed to use an array.

PR #2430, opened 2026-08-31
avoid a ramp-sized temporary in the uniform-weighted slopes

schlafly:
One more memory-reduction PR: the new uniform weight computation computed a product of the double precision floats with the single precision ramp values, causing a copy of the full ramp to be made at double precision.  That can be avoided by replacing the sum(weights * data) with a np.einsum("i,ijk->jk", weights, data).  This only affects dumo which gets cast down to float16.  The einsum call still accumulates internally as double precision before having the difference wrt optimal weights cast as float16, but the order of operations is different than in the sum(weights * data) case, so some numerical precision differences are possible.

In concert with https://github.com/spacetelescope/stcal/pull/577, this removes the spike in memory usage at the end of the ramp fitting step (green line vs. blue line).
image

Regression here, all green: https://github.com/spacetelescope/RegressionTests/actions/runs/33394254962

## Tasks

### DIFF
--- a/changes/2430.ramp_fitting.rst
+++ b/changes/2430.ramp_fitting.rst
@@ -1,0 +1,2 @@
+Reduce peak memory usage of ramp fitting by not materializing a ramp-sized
+temporary while computing the uniform-weighted slopes.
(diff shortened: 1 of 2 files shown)


schlafly:
@t-brandt , mind reviewing?  Together with #2427 this was part of an effort to push down the peak memory use.

t-brandt (approved):
This is a very simple change.  It looks good to me.

schlafly:
Thanks, merging!

PR #2431, opened 2026-08-31
remove maintainers team from codeowners entries

zacharyburnett:
remove @spacetelescope/romancal-maintainers from most manually-defined modules in `.github/CODEOWNERS` so the maintainers team isn't pinged for every single pull request

## Tasks

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,26 +11,26 @@
 /.github/						@zacharyburnett @jhunkeler
 
 # steps and packages
-/romancal/assign_wcs/			@spacetelescope/romancal-maintainers @ddavis-stsci
-/romancal/associations/			@spacetelescope/romancal-maintainers @ddavis-stsci @mairanteodoro
-/romancal/dark_current/			@spacetelescope/romancal-maintainers @ddavis-stsci
-/romancal/datamodels/			@spacetelescope/romancal-maintainers @PaulHuwe
-/romancal/dq_init/				@spacetelescope/romancal-maintainers @ddavis-stsci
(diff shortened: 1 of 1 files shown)


braingram:
Are the branch protections set to require an approval from a codeowner?

If so, won't the changes in this PR make it so maintainers approval is insufficient for PRs that match the modified lines?

zacharyburnett:
> Are the branch protections set to require an approval from a codeowner?
> 
> If so, won't the changes in this PR make it so maintainers approval is insufficient for PRs that match the modified lines?

Yes approval from a codeowner is required in the ruleset, so approval from a maintainer who isn't also a defined codeowner would be insufficient to have the "Merge" button be green. However, maintainers have bypass permissions, so they would just need to check the "bypass approval" box to merge. It would look like this:
image

braingram:
Thanks. That seems like a pattern we don't want to reinforce as it bypasses all checks? Why have them if we're going to bypass them?

zacharyburnett:
we can also uncheck the "require review from Code Owner" box in the ruleset
[image](https://github.com/spacetelescope/romancal/settings/rules/7658093)

Then, anyone with write permission can approve a PR and the "Merge" button will turn green. I think that is the limit of granularity between the two options however

schlafly:
Thanks, following up from a conversation Zach and I had---I asked about this in the context of more reviewers than I expected being added onto PRs like this one: https://github.com/spacetelescope/romancal/pull/2401 .  I meant that to mean 'why are we spamming Zach and Joe' but Zach understood it as 'why are we spamming Eddie.'  I do want to keep getting these requests so I don't think we want this PR.

PR #2435, opened 2026-09-08
Replace source catalog source detection step with a template-based algorithm and deblender

schlafly:
This PR overhauls the romancal 'prompt' catalog source detection, with a few aims:
- increase the depth achieved
- make more compact sources in crowded fields
- don't break anything

The approach taken is to run a series of matched filters over the image; the templates are Gaussians of [0.2", 0.6", 2.4"] in FWHM.  Images are first background-subtracted with a background scale ~12x larger than the template, so these templates are looking at excesses over a kind of local background and point source sensitivity is not biased high when riding on the wing of a large galaxy.  We estimate the signal and the noise in these convolved matched filter images to construct an SNR image.  The SNR image uses the image noise estimate from the 'err' field in the images.  To address any calibration errors in that image, the SNR image is then normalized by the observed RMS in the SNR image; for small kernels this empirical normalization tends to be around 1 but for large profiles it is increasingly contaminated by source noise.

A 'max' is taken over the normalized SNR images for the different templates.  The resulting max SNR image is deblended, defining the initial candidate source catalog.  The segments in this source catalog are then looped over one-by-one from most significant to least significant.  The template each segment comes from is identified according to which template SNR image contributed the peak SNR.  Each segment is narrowed to the footprint of SNR image from which it came; this leads segments for point sources to not include large wings coming from the convolution with the wide Gaussian template.  Additionally segments are dilated by one pixel; any pixels already claimed by a more significant segment are removed; the contiguous region with the peak SNR is taken; and the segment is dropped if it contains fewer than 'npixels' remaining pixels.  Broadly, this procedure does less than it sounds; because segments are already disjoint on the max image and are only narrowed by the intersection with the significant regions of the per-template SNRs, the initial segment list is quite disjoint.  The 1 pixel widening is to allow a single 5 sigma pixel segment to gain enough surrounding pixels to get aperture flux measurements, and the other gymnastics allows that mild widening to be done in a way that doesn't produce weird invalid segments.

With the segment list in hand source catalog measurements can be performed as usual.  The one additional modification is that the various moments the source catalog step computes are run on a convolved image.  We now have several convolved images; we use the PSF-convolved SNR image for these computations in this PR.

This PR improves performance of the source catalog step dramatically for a range of different environments.  For example, here are images of a series of fields with the Kron radii of detected sources in the original and new code.
image
image
image
image
image

The new catalogs are in my opinion in all cases better, though they are not perfect.  In the first galaxy field the new catalog is deeper (more faint, real sources) and avoids shredding the core of the large galaxy, but it fails to deblend one source at the top of the field as well as the old catalog did.  The new star field is strictly better, avoiding some aggregate sources that were present in the old catalog (due to turning deblending on by default).  The simulated HLWAS L2 and L3 images essentially just show that the new catalogs detect significantly more faint objects---one of the chief motivations for this effort, because we were not going deep enough to detect all of the sources where the dispersed spectroscopy algorithm expects to find and photometer sources.  The new branch is also much better behaved in the bulge, where the old algorithm linked all of the stars into large clumps that spidered through the entire image, resulting in ridiculous segments and moments.  The new source detection is not great---it's not a crowded field photometry system---but it is much better.

I added a new max_sources argument and set its value to 30k; only the 'max_sources' brightest sources are measured.  This is to prevent spending too much time in the bulge field and to mitigate one performance issue in photutils that I'll fix in a later PR.

I changed the ~meaning of a few additional arguments in a way that is worth calling out:
- kernel_fwhm: this is now in arcsec rather than in pixels so that the same arguments mean the same things for L2 and L3 images
- snr_threshold: before this was the per-pixel SNR, and now it's the template SNR, so SNR = 5 corresponds to one of the profiles being detected at 5 sigma
- npixels: this is now default 9; it's the smallest number of pixels that a segment to have for us to accept it.  We internally use npixels = 1 (accept segments even if they have only one pixel at >5 sigma in the SNR image) but these get expanded by 1 pixel in all directions, so we'll detect an isolated 5 sigma source.  If that 5 sigma source is blended with neighbors though the segments get too small and this will throw it away.  Too small values will lead to a lot of garbage as moments get computed from nearly empty segments.

Two ~bugs also got fixed:
- we now use mask & coverage_mask more correctly in RomanBackground.  Prior to this PR the SNR image got NaNed due to using coverage_mask rather than mask, and this led sources to get split up into lots of pieces in regions of L2 images with lots of masked pixels.
- replaced xypos with xypos_finite in the guess to the PSF fitting; the PSF fitting crashes on non-finite guesses and this avoids that.

Regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/34353055207
Small changes to the tweakwcs results due to slightly different PSF positions, presumably because slightly different segments leads to slightly different initial guesses for the PSFs and slightly different final positions.  MAE improves to 0.109 from 0.114; not a meaningful difference.

This PR initially had a bug that led it to work fine with recent data but crash and burn on old data.  The issue was that the convolutions do FFTs and single crazy values can propagate everywhere.  So a cut at SNR = 10000 is performed to avoid letting any pixel totally trash the convolutions.

@darioflute , I promised you I would send you this branch; would you mind taking a look to see if you like this approach?

## Tasks


schlafly:
@larrybradley , would you mind taking a look at this one too?  I think this approach will let me harmonize the multiband and single band source catalog approaches; I'll take that up in a follow-up PR.

PR #2446, opened 2026-09-16
RCAL-1438: Fix multiband catalog L3 metadata blending

### LINKED ISSUE #2425: Data Release Product Filenames
_Issue [RCAL-1438](https://jira.stsci.edu/browse/RCAL-1438) was created on JIRA by [Tyler Desjardins](https://jira.stsci.edu/secure/ViewProfile.jspa?name=desjard):_

Copy/pasting from a Slack thread in #roman_data_release_dev:

From [Scott Fleming](https://jira.stsci.edu/secure/ViewProfile.jspa?name=fleming) :
the multi-band file `r00342_ref-hlwas-deep-26a_full_007m42x23y55_cat.parquet` is showing up as an optical element of F184 in AUI, if it's indeed multi-band I would assume we'd want that to either be empty or the full list of filters that went into its construction?
 
I see Segment number is being set to "1" but not sure if that's well-defined for DR products that combine multiple observations together?

defined in the L3 ASDF file:

observation: {execution_plan: 1, exposure: null, exposure_grouping: null, observation: 1,
      pass: null, program: 342, segment: 1, visit: null}

From [Tyler Desjardins](https://jira.stsci.edu/secure/ViewProfile.jspa?name=desjard) :

exposure grouping isn’t being set properly I believe. If I recall properly, it should be the roll up of what was combined. Imagine if you stacked everything in a pass, then it should be something like “r0034201001” for PID 342, execution plan 1, pass 1. And other fields should be null….you can’t have pass be null but segment is set. That makes no sense.

mairanteodoro:
Resolves [RCAL-1438](https://jira.stsci.edu/browse/RCAL-1438)

Closes #2425

Harden L3→L3 metadata blending for multiband source catalogs so DR multi-filter products no longer inherit a single-filter `optical_element` (e.g. F184) from the first input mosaic. Finalize `coadd_info` with min/mean/max rules (including `max_exposure_time`), keep `meta.image.filename` as the association name, preserve per-filter meta in `image_metas`, and leave `wcsinfo` as first-input only.

### Regression tests
- https://github.com/spacetelescope/RegressionTests/actions/runs/35247922406 (09/17/26 - all passing)

## Tasks


romancal/multiband_catalog/_metadata.py
```diff
@@ -69,10 +82,63 @@ def blend_image_metadata(
...
+
+    max_exposure_times : list
+        Accumulated max exposure times from input L3 coadds.
+    """
+    if time_means:
+        cat_model.meta.coadd_info.time_mean = Time(time_means).mean()
+    if exposure_times:
+        cat_model.meta.coadd_info.exposure_time = float(np.mean(exposure_times))
+    if max_exposure_times:
+        cat_model.meta.coadd_info.max_exposure_time = float(np.max(max_exposure_times))
```
schlafly, line 131: In the previous design things like max(...) are accumulated in the blender; this could likewise go there?

romancal/multiband_catalog/_metadata.py
```diff
@@ -69,10 +82,63 @@ def blend_image_metadata(
...
+
+    # Explicit multi-filter rule: top-level optical_element cannot hold a
+    # list (enum), so set None when more than one distinct filter is present.
+    # Individual filters remain in image_metas.
+    filters = []
+    for image_meta in cat_model.meta.get("image_metas", []):
+        instrument = image_meta.get("instrument") or {}
+        optical_element = instrument.get("optical_element")
+        if optical_element is not None:
+            filters.append(optical_element)
```
schlafly, line 141: Remind me what the blender currently does here?  The blender could accumulate this as well, without needing separate behavior here?

romancal/multiband_catalog/_metadata.py
```diff
@@ -69,10 +82,63 @@ def blend_image_metadata(
...
+    exposure_times : list
+        Accumulated exposure times from input L3 coadds.
+
+    max_exposure_times : list
+        Accumulated max exposure times from input L3 coadds.
+    """
+    if time_means:
+        cat_model.meta.coadd_info.time_mean = Time(time_means).mean()
+    if exposure_times:
+        cat_model.meta.coadd_info.exposure_time = float(np.mean(exposure_times))
```
schlafly, line 129: The two means would be annoying to handle in the blender so it makes sense to accumulate them in the blender and finalize, as you're doing here.

schlafly (commented):
Some comments inline.

The other part of the issue was about segment / observation / visit / ... grouping.  I speculated there that we happened to have a set of L2s that all had the same segment but different passes, leading to somewhat weird metadata.  Was that the right speculation?

romancal/multiband_catalog/_metadata.py
mairanteodoro: Yep. Done.

romancal/multiband_catalog/_metadata.py
mairanteodoro: Yeah, I think you're right. Moved it.

mairanteodoro:
> The other part of the issue was about segment / observation / visit / ... grouping. I speculated there that we happened to have a set of L2s that all had the same segment but different passes, leading to somewhat weird metadata. Was that the right speculation?

I reached out to Scott Fleming and Jennifer Melot, and she pointed me to where the dataset is so that I could take a look at it.

Indeed, your suspicion has been confirmed: all the L2s have the same segment and different passes. So the L2 to L3 `MetaBlender` correctly keeps shared `segment=1` and nulls disagreeing `pass`/`visit`/`exposure`. Because `pass`/`visit` are null, `exposure_grouping` stays null.

PR #2447, opened 2026-09-17
RCAL-981: Catch KeyError when resolving skycell WCS in ResampleData

### LINKED ISSUE #1565: Mosaic pipeline crashed when target is not given
_Issue [RCAL-981](https://jira.stsci.edu/browse/RCAL-981) was created on JIRA by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis):_

The mosaic pipeline crashes when the target is not a valid skycell name or is missing.

The correct behavior is that if the skycell name is missing to resample the data onto a regular grid instead of a predefined skycell.

mairanteodoro:
Resolves [RCAL-981](https://jira.stsci.edu/browse/RCAL-981)

Closes #1565

This PR updates `ResampleData.__init__` so skycell WCS resolution catches `KeyError` in addition to `ValueError`. Unknown skycell names from `SkyCells.from_names` previously escaped the fallback path; now resample continues by building the output WCS from the input models. A unit test covers both caught error types and verifies the fallback WCS.

### Regression tests
- https://github.com/spacetelescope/RegressionTests/actions/runs/35247855628 (09/17/26 - all passing)

## Tasks


schlafly:
This PR looks sane but can you remind me of the design?  Which key are we looking for?  If the user has specified a skycell that they want and ~misspelled it, crashing with a warning seems better than making the wrong product.  If we're overloading some key that doesn't need to be a skycell and plausibly has an unrelated sensible value, then falling back to a non-skycell product is the right choice.

Can you remind me how this works?  Thanks!

mairanteodoro:
> Can you remind me how this works? Thanks!

Sure. If no `skycell_wcs_info` and no usable `target` are provided, a `ValueError` is caught, a warning message is logged, and then we proceed to build an output GWCS from the input images' footprint. This same path is used in the case where `target` is "none" or missing or no WCS info is provided.

With this PR, if `target` is set to something that's not a skycell, or if it's a typo, or if `skycell_wcs_info.name` set but unknown in skymap, a `KeyError` is caught, a warning message is logged, and then we proceed to build the output GWCS from the sky footprint, like the previous path.

If `target` was never overloaded, then I think crashing would be reasonable (e.g. a typo in the custom skycell name). However, because "target" sometimes means something else, a non-skycell value (other label the association format carries, for example, a program/target name) would still be valid, not necessarily a misspelled skycell name, in which case fallback to a footprint WCS would be preferred, no? If not, we can tighten the behavior so that we hard-fail on bad `skycell_wcs_info.name` only.

PR #2454, opened 2026-09-17
RCAL-1449: cache ProjectionRegion

### LINKED ISSUE #2452: Cache ProjectionRegion and cell polygons across find_skycell_matches calls
## Summary

`find_skycell_matches` constructs a new `ProjectionRegion` on every candidate index and therefore rebuilds that region’s ~2000-cell `SkyCells` k-d tree and spherical polygons on every call. Region geometry is immutable for a given skymap reference file. We should cache the constructed objects for performance.

Seen on current `main`:
- https://github.com/spacetelescope/romancal/blob/main/romancal/skycell/match.py#L219
- https://github.com/spacetelescope/romancal/blob/main/romancal/skycell/skymap.py#L523
- https://github.com/spacetelescope/romancal/blob/main/romancal/skycell/skymap.py#L218
- https://github.com/spacetelescope/romancal/blob/main/romancal/skycell/skymap.py#L492

## What happens

```python
for projregion_index in nearby_projregion_indices:
    projregion = sc.ProjectionRegion(projregion_index)
    if footprint.polygon.intersects_poly(projregion.polygon):
        projregion.skycells.kdtree.query_ball_point(...)
        # then SkyCells(...).polygons for the inner hits
```

`ProjectionRegion.polygon`, `ProjectionRegion.skycells`, `SkyCells.kdtree`, and `SkyCells.polygons` are `@cached_property`. That cache dies with the instance. The matcher allocates a new `ProjectionRegion` each time, so the properties are rebuilt from scratch.

A typical non-polar projection region has ~2000 skycells. Each match therefore:

1. Builds the region spherical polygon.
2. Builds a `SkyCells` for **all** cells in the region and a k-d tree of their centers.
3. Builds `SingleSphericalPolygon` objects for the inner-tree hits (often most of the region).

None of that depends on the input footprint. It depends only on the skymap table.

## Why this matters

L3 association generation calls `find_skycell_matches` **once per calibrated L2**:

https://github.com/spacetelescope/romancal/blob/main/romancal/associations/skycell_asn.py#L130

Deep-field and mosaic programs may process thousands of L2 files that fall in the same few projection regions. Catalog, cutout, and reprocessing jobs have the same pattern. `intersects_poly` still performs the final exact check, so rebuilding the same geometry for every file only adds overhead.

The per-region tree currently does not narrow the search much because its radius mixes radians and degrees. It often considers nearly all ~2,000 cells in a region. Fixing the units would improve filtering, but caching would still avoid rebuilding the tree and polygons on later calls. See: https://github.com/spacetelescope/romancal/issues/2448

## Suggested approach

Add a cache to `SkyMap` that maps projection-region indices to `ProjectionRegion` objects.

When `find_skycell_matches` needs a region, it should:

1. Return the cached object if it exists.
2. Otherwise create the region, store it in the cache, and return it.

The cache should belong to the `SkyMap`, rather than using a process-wide cache keyed by `id(skymap)`. It will then have the same lifetime as the reference data and cannot accidentally mix different skymaps.

When creating a region, make sure its `SkyCells` uses the same `SkyMap` instead of silently falling back to the global `SKYMAP`.

The existing cached properties on `ProjectionRegion` and `SkyCells`, including the k-d tree and polygons, will then be reused automatically. No additional polygon cache or ASDF schema change is needed.

The public `find_skycell_matches` API and its results should remain unchanged.

## Tests

Add an assertion that a second call for the same footprint (same `skymap`) does not construct new `ProjectionRegion` objects / does not rebuild the region k-d tree. Patching `ProjectionRegion.__init__` or `SkyCells.kdtree` and comparing call counts should be enough.

## Related

https://github.com/spacetelescope/romancal/issues/2000 (serialize a global cell k-d tree into the skymap ASDF). That is a larger change: ~8e6 cells, ~55s to build, and scipy trees do not serialize cleanly into ASDF.

Caching constructed `ProjectionRegion` objects is smaller, process-local, and needs no schema change. It helps the current two-level matcher immediately. A serialized global tree can still replace the outer/inner queries later.

This issue was discovered / reported with the assistance of AI. I've reviewed and verified the issue.

falkben:
Resolves [RCAL-1449](https://jira.stsci.edu/browse/RCAL-1449)

Closes #2452

This PR caches constructed `ProjectionRegion` objects on the `SkyMap` instance. Later matches reuse the same region polygon, per-region k-d tree, and cell polygons. `ProjectionRegion.skycells` now uses the parent skymap. Public `find_skycell_matches` results are unchanged.

## Tasks


zacharyburnett:
I think I prefer #2453, but I still need to get the tests passing and add a change log entry there

I don't think these new tests added here are necessary

falkben:
> I think I prefer #2453, but I still need to get the tests passing and add a change log entry there
> 
> I don't think these new tests added here are necessary

The tests were required to pass the coverage checks in CI.

PR #2455, opened 2026-09-18
Use photutils 3.1 for centroid errors, sky orientation, and annulus background error

larrybradley:

This PR updates the `source_catalog` step to use new features in
photutils 3.1 and pins romancal to that version. The pin currently
points at photutils `main` and will switch to `photutils>=3.1,<4` once
3.1.0 is released (hopefully next week). Note that photutils 3.1 requires newer
minimum versions than romancal currently declares, so this PR raises them to match:
* Python 3.12+
* Astropy 7.0+
* Scipy 1.15+

I'll keep this PR in a Draft state until 3.1.0 is released, unless you want it sooner.

### Changes

- The eight centroid error columns (`x/y_centroid_err`,
  `x/y_centroid_win_err`, `ra/dec_centroid_err`, and
  `ra/dec_centroid_win_err`) are now calculated by `SourceCatalog`,
  replacing the zero-valued placeholders.
- `orientation_sky` now comes from `SourceCatalog.sky_orientation`,
  which propagates the pixel covariance through the local WCS Jacobian.
  This accounts for local rotation, parity, and distortion. The
  romancal `north_angle_at` helper is removed. Values are now in the
  range (-90, 90] degrees instead of [0, 360).
- `aper_bkg_flux` and `aper_bkg_flux_err` are now calculated with a
  batched `ApertureStats` call using the new `median_err` property,
  replacing a per-source Python loop. The error uses the sample
  standard deviation, so values are ~0.1-0.3% larger.

Regression test truth files will need to be updated.

### photutils 3.1 performance improvements

Many of the photutils 3.1 performance improvements were driven by the
romancal `source_catalog` step, which benefits from them directly.

- **Aperture photometry** is typically ~2-25x faster. All sources are
  computed in a single call into compiled code.
- **`ApertureStats`** is typically ~5-15x faster, including
  sigma-clipped statistics computed in compiled code.
- **`SourceCatalog`** windowed centroids, Kron radii and photometry,
  `flux_radius`, image moments, and segment statistics are typically
  ~3-20x faster.
- **Source deblending** (`deblend_sources` and `SourceFinder`) is
  typically ~7-30x faster with identical results.
- **PSF photometry** with `GriddedPSFModel` is ~20-25% faster per
  evaluation. A new analytic Jacobian removes ~60% of the model
  evaluations during fitting.
- **`Background2D`** sigma clipping and box statistics are now computed
  in a single pass in compiled code.
- New `n_threads` keywords on `AperturePhotometry`, `ApertureStats`,
  `SourceCatalog`, `deblend_sources`, `SourceFinder`, and
  `Background2D` allow multithreading. If the operational pipeline allows
  multithreading, this would improve performance even more (if needed), but it's
  already very fast. The compiled code releases the GIL and results are identical 
  to the single-threaded computation.
  
This PR revealed a bug in the forced catalogs that I will fix in a followup PR (#2456).

PR developed in collaboration with Claude / AI.

## Tasks


schlafly:
Regression tests started here: https://github.com/spacetelescope/RegressionTests/actions/runs/35395355525
Maybe a tiny amount of churn in the source positions is expected?

schlafly:
This is great, Larry!  A few questions:
- Just checking, are you careful about the direction to north close to the pole?  We have skycells right at the N/S poles and I realized as I was doing the other PR that by putting the center right at dec = +/-90 we were asking for trouble there.
- We have a delivery in one month; if you are confident that 3.1.0 will be out by then, happy to move to photutils main.
- We would want nthreads = 1 by default, but cool to support >1 values.  
- Huge performance improvements!

@braingram , this would remove python 3.11, which we've been reluctant to do, and increase astropy to 7 and scipy to 1.15.  I think we want to get this in but I'd appreciate feedback.

schlafly:
Regtests show only minor churn, matching my expectations.  Somewhat interestingly, the PSF astrometry matches Gaia meaningfully better (1.4 mas before this PR, 1.0 mas after this PR), which is an unexpected benefit?

braingram:
> @braingram , this would remove python 3.11, which we've been reluctant to do, and increase astropy to 7 and scipy to 1.15. I think we want to get this in but I'd appreciate feedback.

I don't see an option to continue supporting 3.11 with this change so users will be forced to update. We should coordinate regtest runner changes with this PR.

braingram:
FWIW the regtests are showing several increases in memory usage. Perhaps there is a trade-off in performance? Based off of https://innerspace.stsci.edu/spaces/SCSB/pages/347636565/RCAL+pipeline+Timing+and+Memory?preview=/347636565/718385813/mem_mos_r00_1-6_flatsat_f158_Jul06.png source catalog is a small portion of the runtime and the current max memory for HLP. Is there a way to tune this to reduce the memory usage (or return it to previous levels)?

larrybradley:
@schlafly Yes, I was careful about the poles. I'll add a test here. Photutils 3.1 will be out before Oct 15 (is that the deadline?). I'll remove draft status on this PR after I investigate a few things (see below).

> Regtests show only minor churn, matching my expectations. Somewhat interestingly, the PSF astrometry matches Gaia meaningfully better (1.4 mas before this PR, 1.0 mas after this PR), which is an unexpected benefit?

That's great to hear!  The only change I can think of for the improvement is that `GriddedPSFModel` now computes analytic fit derivatives during the fitting. Let me check.

> I don't see an option to continue supporting 3.11 with this change so users will be forced to update. We should coordinate regtest runner changes with this PR.

@braingram Do you prefer the update to Python 3.12+ go in a separate PR (merged before this one)?

> FWIW the regtests are showing several increases in memory usage. Perhaps there is a trade-off in performance? Based off of https://innerspace.stsci.edu/spaces/SCSB/pages/347636565/RCAL+pipeline+Timing+and+Memory?preview=/347636565/718385813/mem_mos_r00_1-6_flatsat_f158_Jul06.png source catalog is a small portion of the runtime and the current max memory for HLP. Is there a way to tune this to reduce the memory usage (or return it to previous levels)?

@braingram I'll look into possibly reducing the memory usage.

braingram:
> @braingram Do you prefer the update to Python 3.12+ go in a separate PR (merged before this one)?

Thanks. Yes I think that would be easier. @schlafly any reason to not do it before this PR (whenever a PR dropping 3.11 is available)?

schlafly:
Fine with me to do a separate python 3.12 PR, thanks both.

FWIW Brett, the regtests don't use realistic source catalogs and so the source catalog step time consumption is not representative there.  But thanks for noticing the increased memory consumption.

larrybradley:
@schlafly I've confirmed that `GriddedPSFModel` analytic derivatives that are part of photutils 3.1 are the source of the improved astrometry. I added the derivatives only to improve PSF fitting performance, so this is a nice bonus!

But the reason is a bit of a surprise. This revealed a longstanding bug in astropy modeling/fitting that has been there from the start. The previous astrometry should have been better. The bug causes fitting degradation when models don't provide derivatives (I'm going to submit a fix to astropy). Adding the model derivatives effectively bypassed the astropy bug, giving the improved astrometry.

schlafly:
Yeah, if the derivatives weren't being computed right that would be another performance bonus on top of their being analytic!

larrybradley:
@schlafly I added the test at the poles. Also, it doesn't affect this PR, but I submitted the astropy modeling/fitting fix (https://github.com/astropy/astropy/pull/20462).

schlafly:
Thanks.  I'll try to do the python 3.12 patch before we merge this.  I was just interested in knowing that you were careful at the poles and didn't need the extra test, sorry!  Were you able to decide if the increased memory usage was necessary?

larrybradley:
I've identified a few areas where I can reduce memory usage. That's still in progress, but it shouldn't require updates to this PR. `romancal` would pick those up when `photutils` `main` is updated.

romancal/source_catalog/tests/test_wcs_utils.py
```diff
@@ -206,39 +209,102 @@ def test_segment_geometry_tracks_local_pixel_area():
...
+        shape,
+        distortion=models.Shift(-center) & models.Shift(-center),
+    )
+    assert_allclose(wcs(center, center)[1], dec_pole, atol=1e-10)
 
+    yy, xx = np.mgrid[: shape[0], : shape[1]]
+    theta = np.deg2rad(25.0)
+    data = models.Gaussian2D(500.0, center, center, 3.0, 1.5, theta)(xx, yy)
+    segment_img = detect_sources(data, 10.0, n_pixels=10)
+    assert segment_img.n_labels == 1
```
schlafly: FWIW, I'm happy just deleting this test; when I moved the code to be in romancal it felt important to test the angle to north behavior, but since photutils is now responsible I don't feel a need to test it, especially at this level of detail.  I'll let you be the judge but I don't think that romancal really needs this now.

schlafly (commented):
One more minor comment about the testing of the pole handling---IMO since romancal is now delegating the pole handling to photutils I don't think we need to be so detailed, but I'll let you decide.  I merged #2464 and this looks pretty good to me so I'm willing to merge this when you are, thanks.

romancal/source_catalog/tests/test_wcs_utils.py
larrybradley: I'll drop the test. There's a similar one in photutils.

larrybradley: Done.

larrybradley:
@schlafly This is ready for merge now.

pyproject.toml
```diff
@@ -12,13 +12,14 @@ classifiers = [
...
+    "astropy>=7.0,<9",
     "astropy-healpix>=1.1.2",
     "jsonschema>=4.8",
     "numpy>=2.0",
-    "scipy>=1.14.1",
+    "scipy>=1.15.0",
     "requests>=2.26",
-    "photutils>=3.0,<4",
+    # "photutils>=3.1,<4",
+    "photutils@git+https://github.com/astropy/photutils@main",
```
braingram: @larrybradley would you make a release of photutils? If not is there something holding up that release? We typically only pin to main for a few packages, all of which can be released by the build engineers if needed for a romancal release. Since the build engineers cannot release photutils I think this should wait until a release is available.

larrybradley: I'm not ready to release 3.1 yet. I'm working on improving memory usage and a few finishing touches. It will be out before Oct 15 (next romancal release).

schlafly: Brett, I'm ready to go ahead with the photutils github pin given the schedule for 3.1; let me know if you still think we should wait.

braingram: I think we should wait until there is a release.

PR #2456, opened 2026-09-18
Fix double unit conversion of the forced detection image

larrybradley:
`SourceCatalogStep` saved the detection image after it had been converted
to nJy in place, but with no unit attached. Forced photometry then
converted it a second time.

PR changes:
- Save a `detection_image_unit` key in the segmentation map file. Forced
  photometry uses it to skip the second conversion, without copying the
  image.
- A `SourceCatalogStep` segmentation file without the key is assumed to
  be in nJy, with a warning. A multiband file without the key is treated
  as being in the units of the input images.
- Fix the multiband detection image being scaled twice. It is now saved
  in nJy with the unit key instead of as a copy.
- `detection_image` is no longer saved when no sources are detected.
- Document both keys and fix a regression test assertion that compared
  the truth column names with themselves.

Regression test truth files may need updating. The flux-scale-dependent
values in the multiband detection catalog change.

PR developed in collaboration with Claude / AI.

## Tasks


schlafly:
Regression tests started here: https://github.com/spacetelescope/RegressionTests/actions/runs/35395426332

schlafly:
First version of the regtests failed because this PR fixed the regtest test, which revealed that an earlier PR had inadvertently put in the wrong regtest file, which was previously unchecked.  I fixed the file and pushed; my expectation is that regtests pass following this change.

schlafly:
Second regression test attempt is green: https://github.com/spacetelescope/RegressionTests/actions/runs/35395426332 .

schlafly (approved):

